PR #2, opened 2020-12-15
app/internal/window,cmd/gogio: fix golint issues and sort imports using goimports

mewmew:
Golint issues:

	app/internal/window/os_wayland.go:1448:12: gioui.org/io/system.StageEvent composite literal uses unkeyed fields
	app/internal/window/os_x11.go:183:12: gioui.org/io/system.StageEvent composite literal uses unkeyed fields

Goimports command:

	find gio/ -type f -name '*.go' | xargs -I '{}' goimports -w '{}'

---

**Note:** This PR is intended to assess the Gio Sourcehut/GitHub integration. It contains no functional code changes, only stylistic.

Preferably, Gio would be able to employ similar collaboration synchronization tools as those employed by the official Go project, which would enable automatically creating Sourcehut patches for corresponding GitHub PRs made to the Gio mirror repository.

Thusly, consider this PR not as a suggestion for code change to Gio, it may very well be ignored and closed. Consider it instead a test of the integration bridge between the Sourcehut Gio repository and the GitHub Gio mirror repository.

Cheers,
Robin

PR #3, opened 2020-12-16
gesture,widget: Text selection in Editor widgets

theclapp:
- [x] Click-and-drag to select text.
- [x] Cut/Copy selected text via Shortcut-X/C
- [x] Paste text via Shortcut-V
- [x] Paste overwrites selected text
- [x] Typing overwrites selected text
- [x] A "select all" hotkey, Shortcut-A
- [x] Selection via keyboard via Shift-Arrows
- [ ] Touch support
- [x] Double-click selects a word
- [ ] Triple-click selects a line
- [x] Shift-click selects from the caret to the clicked position, or adjusts the current selection.

### DIFF
--- a/widget/buffer.go
+++ b/widget/buffer.go
@@ -3,18 +3,13 @@
 package widget
 
 import (
-	"fmt"
 	"io"
 	"strings"
 	"unicode/utf8"
 )
 
-const bufferDebug = false
-
 // editBuffer implements a gap buffer for text editing.
 type editBuffer struct {
-	// caret is the caret position in bytes.
-	caret int
 	// pos is the byte position for Read and ReadRune.
 	pos int
 
@@ -35,12 +30,12 @@ func (e *editBuffer) Changed() bool {
 	return c
 }
 
-func (e *editBuffer) deleteRunes(runes int) {
-	e.moveGap(0)
+func (e *editBuffer) deleteRunes(caret, runes int) int {
+	e.moveGap(caret, 0)
 	for ; runes < 0 && e.gapstart > 0; runes++ {
 		_, s := utf8.DecodeLastRune(e.text[:e.gapstart])
 		e.gapstart -= s
-		e.caret -= s
+		caret -= s
 		e.changed = e.changed || s > 0
 	}
 	for ; runes > 0 && e.gapend < len(e.text); runes-- {
@@ -48,12 +43,12 @@ func (e *editBuffer) deleteRunes(runes int) {
 		e.gapend += s
 		e.changed = e.changed || s > 0
 	}
-	e.dump()
+	return caret
 }
 
 // moveGap moves the gap to the caret position. After returning,
 // the gap is guaranteed to be at least space bytes long.
-func (e *editBuffer) moveGap(space int) {
+func (e *editBuffer) moveGap(caret, space int) {
 	if e.gapLen() < space {
 		if space < minSpace {
 			space = minSpace
@@ -62,29 +57,28 @@ func (e *editBuffer) moveGap(space int) {
 		// Expand to capacity.
 		txt = txt[:cap(txt)]
 		gaplen := len(txt) - e.len()
-		if e.caret > e.gapstart {
+		if caret > e.gapstart {
 			copy(txt, e.text[:e.gapstart])
-			copy(txt[e.caret+gaplen:], e.text[e.caret:])
-			copy(txt[e.gapstart:], e.text[e.gapend:e.caret+e.gapLen()])
+			copy(txt[caret+gaplen:], e.text[caret:])
+			copy(txt[e.gapstart:], e.text[e.gapend:caret+e.gapLen()])
 		} else {
-			copy(txt, e.text[:e.caret])
+			copy(txt, e.text[:caret])
 			copy(txt[e.gapstart+gaplen:], e.text[e.gapend:])
-			copy(txt[e.caret+gaplen:], e.text[e.caret:e.gapstart])
+			copy(txt[caret+gaplen:], e.text[caret:e.gapstart])
 		}
 		e.text = txt
-		e.gapstart = e.caret
+		e.gapstart = caret
 		e.gapend = e.gapstart + gaplen
 	} else {
-		if e.caret > e.gapstart {
-			copy(e.text[e.gapstart:], e.text[e.gapend:e.caret+e.gapLen()])
+		if caret > e.gapstart {
+			copy(e.text[e.gapstart:], e.text[e.gapend:caret+e.gapLen()])
 		} else {
-			copy(e.text[e.caret+e.gapLen():], e.text[e.caret:e.gapstart])
+			copy(e.text[caret+e.gapLen():], e.text[caret:e.gapstart])
 		}
 		l := e.gapLen()
-		e.gapstart = e.caret
+		e.gapstart = caret
 		e.gapend = e.gapstart + l
 	}
-	e.dump()
 }
 
 func (e *editBuffer) len() int {
@@ -96,7 +90,25 @@ func (e *editBuffer) gapLen() int {
 }
 
 func (e *editBuffer) Reset() {
-	e.pos = 0
+	e.Seek(0, io.SeekStart)
+}
+
+// Seek implements io.Seeker
+func (e *editBuffer) Seek(offset int64, whence int) (ret int64, err error) {
+	switch whence {
+	case io.SeekStart:
+		e.pos = int(offset)
+	case io.SeekCurrent:
+		e.pos += int(offset)
+	case io.SeekEnd:
+		e.pos = e.len() - int(offset)
+	}
+	if e.pos < 0 {
+		e.pos = 0
+	} else if e.pos > e.len() {
+		e.pos = e.len()
+	}
+	return int64(e.pos), nil
 }
 
 func (e *editBuffer) Read(p []byte) (int, error) {
@@ -138,18 +150,11 @@ func (e *editBuffer) String() string {
 	return b.String()
 }
 
-func (e *editBuffer) prepend(s string) {
-	e.moveGap(len(s))
-	copy(e.text[e.caret:], s)
+func (e *editBuffer) prepend(caret int, s string) {
+	e.moveGap(caret, len(s))
+	copy(e.text[caret:], s)
 	e.gapstart += len(s)
 	e.changed = e.changed || len(s) > 0
-	e.dump()
-}
-
-func (e *editBuffer) dump() {
-	if bufferDebug {
-		fmt.Printf("len(e.text) %d e.len() %d e.gapstart %d e.gapend %d e.caret %d txt:\n'%+x'<-%d->'%+x'\n", len(e.text), e.len(), e.gapstart, e.gapend, e.caret, e.text[:e.gapstart], e.gapLen(), e.text[e.gapend:])
-	}
 }
 
 func (e *editBuffer) runeBefore(idx int) (rune, int) {
--- a/widget/editor.go
+++ b/widget/editor.go
@@ -18,6 +18,7 @@ import (
 	"gioui.org/f32"
 	"gioui.org/gesture"
 	"gioui.org/io/clipboard"
+	"gioui.org/io/event"
 	"gioui.org/io/key"
 	"gioui.org/io/pointer"
 	"gioui.org/layout"
@@ -65,10 +66,17 @@ type Editor struct {
 	caret struct {
 		on     bool
 		scroll bool
-		// pos is the current caret position.
-		pos combinedPos
+		// start is the current caret position, and also the start position of
+		// selected text. end is the end positon of selected text. If start.ofs
+		// == end.ofs, then there's no selection. Note that it's possible (and
+		// common) that the caret (start) is after the end, e.g. after
+		// Shift-DownArrow.
+		start combinedPos
+		end   combinedPos
 	}
 
+	dragging  bool
+	dragger   gesture.Drag
 	scroller  gesture.Scroll
 	scrollOff image.Point
 
@@ -106,6 +114,13 @@ type combinedPos struct {
 	// xoff is the offset to the current position when moving between lines.
 	xoff fixed.Int26_6
 }
+
+type selectionAction int
+
+const (
+	selectionExtend selectionAction = iota
+	selectionClear
+)
 
 func (m *maskReader) Reset(r io.RuneReader, mr rune) {
 	m.rr = r
@@ -153,9 +168,20 @@ type SubmitEvent struct {
 	Text string
 }
 
+// A SelectEvent is generated when the user selects some text, or changes the
+// selection (e.g. with a shift-click), including if they remove the
+// selection. The selected text is not part of the event, on the theory that
+// it could be a relatively expensive operation (for a large editor), most
+// applications won't actually care about it, and those that do can call
+// Editor.SelectedText() (which can be empty).
+type SelectEvent struct{}
+
 type line struct {
-	offset image.Point
-	clip   op.CallOp
+	offset         image.Point
+	clip           op.CallOp
+	selected       bool
+	selectionYOffs int
+	selectionSize  image.Point
 }
 
 const (
@@ -181,8 +207,13 @@ func (e *Editor) processEvents(gtx layout.Context) {
 		// Can't process events without a shaper.
 		return
 	}
+	oldStart, oldLen := min(e.caret.start.ofs, e.caret.end.ofs), e.SelectionLen()
 	e.processPointer(gtx)
 	e.processKey(gtx)
+	// Queue a SelectEvent if the selection changed, including if it went away.
+	if newStart, newLen := min(e.caret.start.ofs, e.caret.end.ofs), e.SelectionLen(); oldStart != newStart || oldLen != newLen {
+		e.events = append(e.events, SelectEvent{})
+	}
 }
 
 func (e *Editor) makeValid(positions ...*combinedPos) {
@@ -190,19 +221,7 @@ func (e *Editor) makeValid(positions ...*combinedPos) {
 		return
 	}
 	e.lines, e.dims = e.layoutText(e.shaper)
-
-	// Jump through some hoops to order the offsets given to offsetToScreenPos,
-	// but still be able to update them correctly with the results thereof.
-	positions = append(positions, &e.caret.pos)
-	sort.Slice(positions, func(i, j int) bool {
-		return positions[i].ofs < positions[j].ofs
-	})
-	var iter func(offset int) combinedPos
-	*positions[0], iter = e.offsetToScreenPos(positions[0].ofs)
-	for _, cp := range positions[1:] {
-		*cp = iter(cp.ofs)
-	}
-
+	e.makeValidCaret(positions...)
 	e.valid = true
 }
 
@@ -226,24 +245,78 @@ func (e *Editor) processPointer(gtx layout.Context) {
 		e.scrollRel(0, sdist)
 		soff = e.scrollOff.Y
 	}
-	for _, evt := range e.clicker.Events(gtx) {
-		switch {
-		case evt.Type == gesture.TypePress && evt.Source == pointer.Mouse,
-			evt.Type == gesture.TypeClick && evt.Source == pointer.Touch:
-			e.blinkStart = gtx.Now
-			e.moveCoord(image.Point{
-				X: int(math.Round(float64(evt.Position.X))),
-				Y: int(math.Round(float64(evt.Position.Y))),
-			})
-			e.requestFocus = true
-			if e.scroller.State() != gesture.StateFlinging {
-				e.caret.scroll = true
+	for _, evt := range e.clickDragEvents(gtx) {
+		switch evt := evt.(type) {
+		case gesture.ClickEvent:
+			switch {
+			case evt.Type == gesture.TypePress && evt.Source == pointer.Mouse,
+				evt.Type == gesture.TypeClick:
+				prevCaretPos := e.caret.start
+				e.blinkStart = gtx.Now
+				e.moveCoord(image.Point{
+					X: int(math.Round(float64(evt.Position.X))),
+					Y: int(math.Round(float64(evt.Position.Y))),
+				})
+				e.requestFocus = true
+				if e.scroller.State() != gesture.StateFlinging {
+					e.caret.scroll = true
+				}
+
+				if evt.Modifiers == key.ModShift {
+					// If they clicked closer to the end, then change the end to
+					// where the caret used to be (effectively swapping start & end).
+					if abs(e.caret.end.ofs-e.caret.start.ofs) < abs(e.caret.start.ofs-prevCaretPos.ofs) {
+						e.caret.end = prevCaretPos
+					}
+				} else {
+					e.ClearSelection()
+				}
+				e.dragging = true
+
+				// Process a double-click.
+				if evt.NumClicks == 2 {
+					e.moveWord(-1, selectionClear)
+					e.moveWord(1, selectionExtend)
+					e.dragging = false
+				}
+			}
+		case pointer.Event:
+			release := false
+			switch {
+			case evt.Type == pointer.Release && evt.Source == pointer.Mouse:
+				release = true
+				fallthrough
+			case evt.Type == pointer.Drag && evt.Source == pointer.Mouse:
+				if e.dragging {
+					e.blinkStart = gtx.Now
+					e.moveCoord(image.Point{
+						X: int(math.Round(float64(evt.Position.X))),
+						Y: int(math.Round(float64(evt.Position.Y))),
+					})
+					e.caret.scroll = true
+
+					if release {
+						e.dragging = false
+					}
+				}
 			}
 		}
 	}
+
 	if (sdist > 0 && soff >= smax) || (sdist < 0 && soff <= smin) {
 		e.scroller.Stop()
 	}
+}
+
+func (e *Editor) clickDragEvents(gtx layout.Context) []event.Event {
+	var combinedEvents []event.Event
+	for _, evt := range e.clicker.Events(gtx) {
+		combinedEvents = append(combinedEvents, evt)
+	}
+	for _, evt := range e.dragger.Events(gtx.Metric, gtx, gesture.Both) {
+		combinedEvents = append(combinedEvents, evt)
+	}
+	return combinedEvents
 }
 
 func (e *Editor) processKey(gtx layout.Context) {
@@ -287,8 +360,9 @@ func (e *Editor) processKey(gtx layout.Context) {
 	}
 }
 
-func (e *Editor) moveLines(distance int) {
-	e.caret.pos = e.movePosToLine(e.caret.pos, e.caret.pos.x+e.caret.pos.xoff, e.caret.pos.lineCol.Y+distance)
+func (e *Editor) moveLines(distance int, selAct selectionAction) {
+	e.caret.start = e.movePosToLine(e.caret.start, e.caret.start.x+e.caret.start.xoff, e.caret.start.lineCol.Y+distance)
+	e.updateSelection(selAct)
 }
 
 func (e *Editor) command(gtx layout.Context, k key.Event) bool {
@@ -297,6 +371,10 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
 		modSkip = key.ModAlt
 	}
 	moveByWord := k.Modifiers.Contain(modSkip)
+	selAct := selectionClear
+	if k.Modifiers.Contain(key.ModShift) {
+		selAct = selectionExtend
+	}
 	switch k.Name {
 	case key.NameReturn, key.NameEnter:
 		e.append("\n")
@@ -313,29 +391,29 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
 			e.Delete(1)
 		}
 	case key.NameUpArrow:
-		e.moveLines(-1)
+		e.moveLines(-1, selAct)
 	case key.NameDownArrow:
-		e.moveLines(+1)
+		e.moveLines(+1, selAct)
 	case key.NameLeftArrow:
 		if moveByWord {
-			e.moveWord(-1)
+			e.moveWord(-1, selAct)
 		} else {
-			e.MoveCaret(-1)
+			e.MoveCaret(-1, -1*int(selAct))
 		}
 	case key.NameRightArrow:
 		if moveByWord {
-			e.moveWord(1)
+			e.moveWord(1, selAct)
 		} else {
-			e.MoveCaret(1)
+			e.MoveCaret(1, int(selAct))
 		}
 	case key.NamePageUp:
-		e.movePages(-1)
+		e.movePages(-1, selAct)
 	case key.NamePageDown:
-		e.movePages(+1)
+		e.movePages(+1, selAct)
 	case key.NameHome:
-		e.moveStart()
+		e.moveStart(selAct)
 	case key.NameEnd:
-		e.moveEnd()
+		e.moveEnd(selAct)
 	// Initiate a paste operation, by requesting the clipboard contents; other
 	// half is in Editor.processKey() under clipboard.Event.
 	case "V":
@@ -343,12 +421,23 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
 			return false
 		}
 		clipboard.ReadOp{Tag: &e.eventKey}.Add(gtx.Ops)
-	// Copy all text.
-	case "C":
+	// Copy or Cut selection -- ignored if nothing selected.
+	case "C", "X":
 		if k.Modifiers != key.ModShortcut {
 			return false
 		}
-		clipboard.WriteOp{Text: e.Text()}.Add(gtx.Ops)
+		if text := e.SelectedText(); text != "" {
+			clipboard.WriteOp{Text: text}.Add(gtx.Ops)
+			if k.Name == "X" {
+				e.Delete(1)
+			}
+		}
+	// Select all
+	case "A":
+		if k.Modifiers != key.ModShortcut {
+			return false
+		}
+		e.caret.end, e.caret.start = e.offsetToScreenPos2(0, e.Len())
 	default:
 		return false
 	}
@@ -418,7 +507,10 @@ func (e *Editor) layout(gtx layout.Context) layout.Dimensions {
 	}
 	clip := textPadding(e.lines)
 	clip.Max = clip.Max.Add(e.viewSize)
-	it := lineIterator{
+	startSel, endSel := sortPoints(e.caret.start.lineCol, e.caret.end.lineCol)
+	it := segmentIterator{
+		startSel:  startSel,
+		endSel:    endSel,
 		Lines:     e.lines,
 		Clip:      clip,
 		Alignment: e.Alignment,
@@ -427,12 +519,12 @@ func (e *Editor) layout(gtx layout.Context) layout.Dimensions {
 	}
 	e.shapes = e.shapes[:0]
 	for {
-		layout, off, ok := it.Next()
+		layout, off, selected, yOffs, size, ok := it.Next()
 		if !ok {
 			break
 		}
 		path := e.shaper.Shape(e.font, e.textSize, layout)
-		e.shapes = append(e.shapes, line{off, path})
+		e.shapes = append(e.shapes, line{off, path, selected, yOffs, size})
 	}
 
 	key.InputOp{Tag: &e.eventKey}.Add(gtx.Ops)
@@ -451,6 +543,7 @@ func (e *Editor) layout(gtx layout.Context) layout.Dimensions {
 	pointer.CursorNameOp{Name: pointer.CursorText}.Add(gtx.Ops)
 	e.scroller.Add(gtx.Ops)
 	e.clicker.Add(gtx.Ops)
+	e.dragger.Add(gtx.Ops)
 	e.caret.on = false
 	if e.focused {
 		now := gtx.Now
@@ -468,14 +561,33 @@ func (e *Editor) layout(gtx layout.Context) layout.Dimensions {
 	return layout.Dimensions{Size: e.viewSize, Baseline: e.dims.Baseline}
 }
 
+// PaintSelection paints the contrasting background for selected text.
+func (e *Editor) PaintSelection(gtx layout.Context) {
+	cl := textPadding(e.lines)
+	cl.Max = cl.Max.Add(e.viewSize)
+	clip.Rect(cl).Add(gtx.Ops)
+	for _, shape := range e.shapes {
+		if !shape.selected {
+			continue
+		}
+		stack := op.Save(gtx.Ops)
+		offset := shape.offset
+		offset.Y += shape.selectionYOffs
+		op.Offset(layout.FPt(offset)).Add(gtx.Ops)
+		clip.Rect(image.Rectangle{Max: shape.selectionSize}).Add(gtx.Ops)
+		paint.PaintOp{}.Add(gtx.Ops)
+		stack.Load()
+	}
+}
+
 func (e *Editor) PaintText(gtx layout.Context) {
 	cl := textPadding(e.lines)
 	cl.Max = cl.Max.Add(e.viewSize)
+	clip.Rect(cl).Add(gtx.Ops)
 	for _, shape := range e.shapes {
 		stack := op.Save(gtx.Ops)
 		op.Offset(layout.FPt(shape.offset)).Add(gtx.Ops)
 		shape.clip.Add(gtx.Ops)
-		clip.Rect(cl.Sub(shape.offset)).Add(gtx.Ops)
 		paint.PaintOp{}.Add(gtx.Ops)
 		stack.Load()
 	}
@@ -487,12 +599,12 @@ func (e *Editor) PaintCaret(gtx layout.Context) {
 	}
 	e.makeValid()
 	carWidth := fixed.I(gtx.Px(unit.Dp(1)))
-	carX := e.caret.pos.x
-	carY := e.caret.pos.y
+	carX := e.caret.start.x
+	carY := e.caret.start.y
 
 	defer op.Save(gtx.Ops).Load()
 	carX -= carWidth / 2
-	carAsc, carDesc := -e.lines[e.caret.pos.lineCol.Y].Bounds.Min.Y, e.lines[e.caret.pos.lineCol.Y].Bounds.Max.Y
+	carAsc, carDesc := -e.lines[e.caret.start.lineCol.Y].Bounds.Min.Y, e.lines[e.caret.start.lineCol.Y].Bounds.Max.Y
 	carRect := image.Rectangle{
 		Min: image.Point{X: carX.Ceil(), Y: carY - carAsc.Ceil()},
 		Max: image.Point{X: carX.Ceil() + carWidth.Ceil(), Y: carY + carDesc.Ceil()},
@@ -530,10 +642,11 @@ func (e *Editor) Text() string {
 	return e.rr.String()
 }
 
-// SetText replaces the contents of the editor.
+// SetText replaces the contents of the editor, clearing any selection first.
 func (e *Editor) SetText(s string) {
 	e.rr = editBuffer{}
-	e.caret.pos = combinedPos{}
+	e.caret.start = combinedPos{}
+	e.caret.end = combinedPos{}
 	e.prepend(s)
 }
 
@@ -590,8 +703,8 @@ func (e *Editor) moveCoord(pos image.Point) {
 		carLine++
 	}
 	x := fixed.I(pos.X + e.scrollOff.X)
-	e.caret.pos = e.movePosToLine(e.caret.pos, x, carLine)
-	e.caret.pos.xoff = 0
+	e.caret.start = e.movePosToLine(e.caret.start, x, carLine)
+	e.caret.start.xoff = 0
 }
 
 func (e *Editor) layoutText(s text.Shaper) ([]text.Line, layout.Dimensions) {
@@ -625,14 +738,21 @@ func (e *Editor) layoutText(s text.Shaper) ([]text.Line, layout.Dimensions) {
 // CaretPos returns the line & column numbers of the caret.
 func (e *Editor) CaretPos() (line, col int) {
 	e.makeValid()
-	return e.caret.pos.lineCol.Y, e.caret.pos.lineCol.X
+	return e.caret.start.lineCol.Y, e.caret.start.lineCol.X
 }
 
 // CaretCoords returns the coordinates of the caret, relative to the
 // editor itself.
 func (e *Editor) CaretCoords() f32.Point {
 	e.makeValid()
-	return f32.Pt(float32(e.caret.pos.x)/64, float32(e.caret.pos.y))
+	return f32.Pt(float32(e.caret.start.x)/64, float32(e.caret.start.y))
+}
+
+// offsetToScreenPos2 is a utility function to shortcut the common case of
+// wanting the positions of exactly two offsets.
+func (e *Editor) offsetToScreenPos2(o1, o2 int) (combinedPos, combinedPos) {
+	cp1, iter := e.offsetToScreenPos(o1)
+	return cp1, iter(o2)
 }
 
 // offsetToScreenPos takes an offset into the editor text (e.g.
@@ -692,42 +812,56 @@ func (e *Editor) invalidate() {
 
 // Delete runes from the caret position. The sign of runes specifies the
 // direction to delete: positive is forward, negative is backward.
+//
+// If there is a selection, it is deleted and counts as a single rune.
 func (e *Editor) Delete(runes int) {
 	if runes == 0 {
 		return
 	}
-	e.caret.pos.ofs = e.rr.deleteRunes(e.caret.pos.ofs, runes)
-	e.caret.pos.xoff = 0
+
+	if l := e.caret.end.ofs - e.caret.start.ofs; l != 0 {
+		e.caret.start.ofs = e.rr.deleteRunes(e.caret.start.ofs, l)
+		runes -= sign(runes)
+	}
+
+	e.caret.start.ofs = e.rr.deleteRunes(e.caret.start.ofs, runes)
+	e.caret.start.xoff = 0
+	e.ClearSelection()
 	e.invalidate()
 }
 
-// Insert inserts text at the caret, moving the caret forward.
+// Insert inserts text at the caret, moving the caret forward. If there is a
+// selection, Insert overwrites it.
 func (e *Editor) Insert(s string) {
 	e.append(s)
 	e.caret.scroll = true
 }
 
-// append inserts s at the cursor, leaving the caret is at the end of s.
+// append inserts s at the cursor, leaving the caret is at the end of s. If
+// there is a selection, append overwrites it.
 // xxx|yyy + append zzz => xxxzzz|yyy
 func (e *Editor) append(s string) {
 	e.prepend(s)
-	e.caret.pos.ofs += len(s)
+	e.caret.start.ofs += len(s)
+	e.caret.end.ofs = e.caret.start.ofs
 }
 
-// prepend inserts s after the cursor; the caret does not change.
+// prepend inserts s after the cursor; the caret does not change. If there is
+// a selection, prepend overwrites it.
 // xxx|yyy + prepend zzz => xxx|zzzyyy
 func (e *Editor) prepend(s string) {
 	if e.SingleLine {
 		s = strings.ReplaceAll(s, "\n", " ")
 	}
-	e.rr.prepend(e.caret.pos.ofs, s)
-	e.caret.pos.xoff = 0
+	e.caret.start.ofs = e.rr.deleteRunes(e.caret.start.ofs, e.caret.end.ofs-e.caret.start.ofs) // Delete any selection first.
+	e.rr.prepend(e.caret.start.ofs, s)
+	e.caret.start.xoff = 0
 	e.invalidate()
 }
 
-func (e *Editor) movePages(pages int) {
+func (e *Editor) movePages(pages int, selAct selectionAction) {
 	e.makeValid()
-	y := e.caret.pos.y + pages*e.viewSize.Y
+	y := e.caret.start.y + pages*e.viewSize.Y
 	var (
 		prevDesc fixed.Int26_6
 		carLine2 int
@@ -746,7 +880,8 @@ func (e *Editor) movePages(pages int) {
 		y2 += h
 		carLine2++
 	}
-	e.caret.pos = e.movePosToLine(e.caret.pos, e.caret.pos.x+e.caret.pos.xoff, carLine2)
+	e.caret.start = e.movePosToLine(e.caret.start, e.caret.start.x+e.caret.start.xoff, carLine2)
+	e.updateSelection(selAct)
 }
 
 func (e *Editor) movePosToLine(pos combinedPos, x fixed.Int26_6, line int) combinedPos {
@@ -807,13 +942,22 @@ func (e *Editor) movePosToLine(pos combinedPos, x fixed.Int26_6, line int) combi
 	return pos
 }
 
-// MoveCaret moves the caret relative to its current position. Positive
-// distance moves forward, negative distance moves backward. Distance is in
-// runes.
-func (e *Editor) MoveCaret(distance int) {
+// MoveCaret moves the caret (aka selection start) and the selection end
+// relative to their current positions. Positive distances moves forward,
+// negative distances moves backward. Distances are in runes.
+func (e *Editor) MoveCaret(startDelta, endDelta int) {
 	e.makeValid()
-	e.caret.pos = e.movePos(e.caret.pos, distance)
-	e.caret.pos.xoff = 0
+	keepSame := e.caret.start.ofs == e.caret.end.ofs && startDelta == endDelta
+	e.caret.start = e.movePos(e.caret.start, startDelta)
+	e.caret.start.xoff = 0
+	// If they were in the same place, and we're moving them the same distance,
+	// just assign the new position, instead of recalculating it.
+	if keepSame {
+		e.caret.end = e.caret.start
+	} else {
+		e.caret.end = e.movePos(e.caret.end, endDelta)
+		e.caret.end.xoff = 0
+	}
 }
 
 func (e *Editor) movePos(pos combinedPos, distance int) combinedPos {
@@ -849,8 +993,9 @@ func (e *Editor) movePos(pos combinedPos, distance int) combinedPos {
 	return pos
 }
 
-func (e *Editor) moveStart() {
-	e.caret.pos = e.movePosToStart(e.caret.pos)
+func (e *Editor) moveStart(selAct selectionAction) {
+	e.caret.start = e.movePosToStart(e.caret.start)
+	e.updateSelection(selAct)
 }
 
 func (e *Editor) movePosToStart(pos combinedPos) combinedPos {
@@ -866,8 +1011,9 @@ func (e *Editor) movePosToStart(pos combinedPos) combinedPos {
 	return pos
 }
 
-func (e *Editor) moveEnd() {
-	e.caret.pos = e.movePosToEnd(e.caret.pos)
+func (e *Editor) moveEnd(selAct selectionAction) {
+	e.caret.start = e.movePosToEnd(e.caret.start)
+	e.updateSelection(selAct)
 }
 
 func (e *Editor) movePosToEnd(pos combinedPos) combinedPos {
@@ -894,7 +1040,7 @@ func (e *Editor) movePosToEnd(pos combinedPos) combinedPos {
 // moveWord moves the caret to the next word in the specified direction.
 // Positive is forward, negative is backward.
 // Absolute values greater than one will skip that many words.
-func (e *Editor) moveWord(distance int) {
+func (e *Editor) moveWord(distance int, selAct selectionAction) {
 	e.makeValid()
 	// split the distance information into constituent parts to be
 	// used independently.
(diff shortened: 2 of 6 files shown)


theclapp:
Force-pushed with signed commits.  Appreciation to @whereswaldon in https://github.com/gioui/gio/pull/2#issuecomment-745694021 for laying out how to do that.

widget/editor.go
```diff
@@ -225,6 +237,82 @@ func (e *Editor) processPointer(gtx layout.Context) {
...
+				b.WriteString(e.lines[e.startDrag.Y].Layout.Text[e.startDrag.X:])
+				for i := e.startDrag.Y + 1; i < e.endDrag.Y; i++ {
+					b.WriteString(e.lines[i].Layout.Text)
+				}
+				b.WriteString(e.lines[e.endDrag.Y].Layout.Text[:e.endDrag.X])
+				selection = b.String()
+			}
+
+			if selection != "" {
+				fmt.Printf("Selection: '%s'\n", selection)
```
eliasnaur: Remove debugging.

eliasnaur:
Haven't looked closer yet, but I'd *love* a test or two to demonstrate and lock in the new behaviour :)

widget/editor.go
theclapp: Whoops, thought I had.  Will do.

theclapp: The Printf is from the first commit in the PR, and is removed in the second commit.

theclapp:
@eliasnaur I added a test case and did some other refactoring.  PTAL.  Thanks.

theclapp:
Rebased from main & force-pushed.

eliasnaur:
Sorry for the delay. I intend to review this soon, but given its complexity I'd like to release the new renderer first.

theclapp:
There's obviously more to do for this functionality:
* Copy selected text
* Paste overwrites selected text
* Typing overwrites selected text
* A "select all" hotkey
* Selection via keyboard
* Touch support

If I have time to work on any of that, would you prefer that I just added it to this branch (probably making the change larger and presumably harder to review), or work on that in a different branch (branched from this one)?  Happy to do either, but would prefer to optimize in favor of getting it merged sooner.

eliasnaur:
Please add new work as separate commits on top of the others. The combined change will be larger, but that's ok as long as the commits are logically separate units.

theclapp:
Rebased to current main (with the compute renderer) and force-pushed.

Also, earlier, moved the feature list from an earlier comment to the PR description as a checklist.

widget/editor.go
```diff
@@ -318,22 +445,64 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
...
+		s.End.ofs = e.Len()
+		s.makeValid(e)
 	default:
 		return false
 	}
+
+	// Process "shifted" movement, e.g. shift-UpArrow.
+	if mt := movementType(k); mt > 0 {
+		if k.Modifiers.Contain(key.ModShift) {
+			if e.IsSelected() {
```
eliasnaur, line 479: Again, I wouldn't expect an IsSelected check here because the behaviour is the same regardless of the selection: shift+arrow moves the selection end, period.

widget/editor.go
```diff
@@ -318,22 +445,64 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
...
+			e.ClearSelection()
+		}
+	}
 	return true
 }
 
+func movementType(k key.Event) int {
+	switch k.Name {
+	case key.NameUpArrow, key.NameDownArrow, key.NameLeftArrow, key.NameRightArrow:
+		return 1
```
eliasnaur, line 498: Name the constants 0, 1, 2.

widget/editor.go
```diff
@@ -376,8 +545,8 @@ func (e *Editor) Layout(gtx layout.Context, sh text.Shaper, font text.Font, size
 	if viewSize := gtx.Constraints.Constrain(e.dims.Size); viewSize != e.viewSize {
 		e.viewSize = viewSize
 		e.invalidate()
+		e.makeValid()
 	}
-	e.makeValid()
```
eliasnaur, line 380: Is this a performance optimization? If so, please split it out into a separate change; makeValid calls have historically been very tricky to get right, and a separate change for this move will make future git bisections easier.

widget/editor.go
```diff
@@ -936,6 +1153,115 @@ func (e *Editor) NumLines() int {
...
+	l := e.Selection.Len()
+	if l == 0 {
+		return
+	}
+	e.rr.caret = min(e.Selection.Start.ofs, e.Selection.End.ofs)
+	e.ClearSelection()
+	e.Delete(l)
+}
+
+func (e *Editor) ClearSelection() {
```
eliasnaur, line 1170: I suggest unexporting this method; it seems to be a side-effect of higher-level operations (move, delete etc.)

widget/label.go
```diff
@@ -25,67 +27,135 @@ type Label struct {
 	MaxLines int
 }
 
-type lineIterator struct {
+// ScreenPos describes a character position (in text line and column numbers,
+// not pixels): Y = line number, X = rune column.
+type ScreenPos image.Point
```
eliasnaur, line 32: Unexport ScreenPos; it doesn't seem used in public API.

eliasnaur (changes requested):
Excellent work, I can't wait to have selections in Gio!

This is the first round of review comments. The comments are mainly about making Editor always have a selection (or, equivalently, adding width to the caret), and making the zero value (zero characters selected) useful in the existing operations (move, delete etc.).

widget/editor.go
theclapp: > it seems to be a side-effect of higher-level operations 

It is within the Editor code, but I also wanted there to be an easy, obvious way to clear the selection programmatically.  I guess since `Editor.Selection` and `type Selection` are both exported, a user could just say `e.Selection = Selection{}` directly.

Actually though, probably _neither_ should be exported, unless I document that it shouldn't be written to.  Setting up a Selection correctly is tricky, and probably should only be done by setting `e.Selection.Start.ofs` and `e.Selection.End.ofs` and calling `Selection.makeValid` (just as I do now for the "select all" (`Shortcut-A`) code).

How about I unexport both, and add these:
* `func (e *Editor) Selection() (startOfs, endOfs int)` to get the selection offsets
* `func (e *Editor) SetSelection(startOfs, endOfs int)` to set the selection
* `func (e *Editor) SelectedText() string` to get the selected text.  Or should this just be another return value of `Selction()`, above?

widget/editor.go
```diff
@@ -936,6 +1153,115 @@ func (e *Editor) NumLines() int {
...
+	e.Selection = Selection{}
+}
+
+func (e *Editor) IsSelected() bool {
```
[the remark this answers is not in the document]
theclapp: So a positive width means the cursor is at the beginning of the selection, and a negative width means the cursor is at the end of the selection?

widget/editor.go
```diff
@@ -223,8 +268,78 @@ func (e *Editor) processPointer(gtx layout.Context) {
 			if e.scroller.State() != gesture.StateFlinging {
 				e.caret.scroll = true
 			}
+
+			if evt.Modifiers == key.ModShift {
+				if e.IsSelected() {
```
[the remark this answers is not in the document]
theclapp: If you try it in other editors, you'll see that this is not so.  (At least the editor that I tried, to wit, the Github comment editor, ha; and also I just tried it in the Google Mail editor.)  Shift-Click near one side of the selection moves that side, and nearer the other moves that side.

Consider

```
aaa s|ss ddd
```

with the caret where the `|` is.  If you Shift-Click on the last d of ddd, that should select `ss ddd`, and move the caret to where you clicked.  If you then Shift-Click in front of the first a of aaa, that should select `aaa sss ddd` (and also move the caret to where you clicked).

If I do as you suggest, then the first Shift-Click does select `ss ddd`, but the second selects `aaa s`, which is wrong.  Or at least not what *I* want.

widget/editor.go
```diff
@@ -318,22 +445,64 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
...
 	case "V":
 		if k.Modifiers != key.ModShortcut {
 			return false
 		}
 		clipboard.ReadOp{Tag: &e.eventKey}.Add(gtx.Ops)
-	case "C":
+	// Copy or Cut selection -- ignored if nothing selected.
+	case "C", "X":
+		if k.Modifiers != key.ModShortcut || !e.IsSelected() {
+			return false
```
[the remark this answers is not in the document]
theclapp: > I'd expect the usual scroll-to-caret behaviour regardless of the selection.

Neither the Gmail editor nor the Github comment editor do that.  That is, if you have no text selected and scroll away from the caret and press Cmd-X, the editor is not scrolled back to the caret (and the caret is not moved).

The two differ with respect to whether a Cut or Copy clear the clipboard if you have nothing selected.  Gmail clears it, Github doesn't.  I prefer the latter behavior (which, um, is why I wrote it to work that way :).

> I suggest replacing IsSelected here with a check for SelectedText != "" below.

I'm not sure what you mean.  I think you mean: scroll back to the caret, and don't update the clipboard if there's nothing selected?

widget/editor.go
theclapp: No.  See my response to your comment on line 273.

widget/editor.go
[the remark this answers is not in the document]
theclapp: A problem with this idea is that PageUp/Down move the caret but not the selection.

On the other hand, neither Gmail nor Github do it that way.  (PageUp/Down scroll the screen, but if you then use an arrow key, you scroll back, whether you have a selection or not.)

(Vim does it yet a third way, where starting "visual mode" and then pressing PageUp/Down moves both the cursor and the selection end point; for the Gio editor, it'd be equivalent to treating PageUp/Down as their shifted equivalents if there was a selection.)

So I think I could make that change, at the expense of having PageUp/Down scroll the editor but not move the caret.  Or having PageUp/Down remove the selection.  The latter would be easier to implement, I think, but also I think I prefer the former behavior.

Thoughts?

widget/editor.go
```diff
@@ -223,8 +268,78 @@ func (e *Editor) processPointer(gtx layout.Context) {
...
+				s.dragSide = &s.End
+			}
+			e.dragging = true
+
+		// Process a double-click. Note that it's not possible to get a
+		// double-click without also having already gotten a Press, and so we
+		// need not duplicate any of the logic in the Press code, like moving
+		// the cursor or checking the scroller state.
+		case evt.Type == gesture.TypeClick && evt.Source == pointer.Mouse,
+			evt.Type == gesture.TypeClick && evt.Source == pointer.Touch:
```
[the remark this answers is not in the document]
theclapp: Done.

theclapp:
Pushed some new code.  Marked several comments as "resolved".  The rest are still outstanding.

theclapp:
You mentioned "this is the first round of review comments", so maybe you meant to address this later, but just in case: The selection color is hardcoded to `color.NRGBA{B: 0xff, A: 0x40}` which is suboptimal, to say the least.  But I'm not sure where the color should be specified.  Any thoughts?

eliasnaur:
> You mentioned "this is the first round of review comments", so maybe you meant to address this later, but just in case: The selection color is hardcoded to `color.NRGBA{B: 0xff, A: 0x40}` which is suboptimal, to say the least. But I'm not sure where the color should be specified. Any thoughts?

Specifying the color is best left for the theme, which in your case probably means export paintSelection as PaintSelection, calling it from material.Editor. Just like PaintText, PaintSelection would just use op.PaintOp and not specify a color.

widget/editor.go
eliasnaur: The three replacement methods SGTM.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: I tried your example in the GH comment editor on macOS. It seems to me your description of the intended behaviour doesn't match:

https://user-images.githubusercontent.com/524812/103751349-70f50b80-5008-11eb-93c7-cd2013850775.mov

Regardless of the correct behaviour, I stand by my point to move the selection awareness to the lower-level Editor operations (move, delete etc.). In other words, I hope that the higher-level processPointer can remain unaware of whether there are a non-zero selection.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: > So a positive width means the cursor is at the beginning of the selection, and a negative width means the cursor is at the end of the selection?

Yes. Just to be clear, I was merely bringing an alternative representation to your attention in case you found it better. I'm fine with (selectionStart, selectionEnd) as well. The main point was that I'd like to assume a selection, and move the zero-width selection special cases as low-level as possible, in order to decrease the risk of forgetting edge cases.

> A problem with this idea is that PageUp/Down move the caret but not the selection.
> 

I'm not sure I see the problem, can you elaborate? I don't mind there be two methods for pageup/down, one for regular pageup/down, and another for shift+pageup/down. In other words, it's fine (and expected) to check for ModShift in high-level code; my hope is to remove checks on SelectionLength/IsSelected in high-level logic.

> On the other hand, neither Gmail nor Github do it that way. (PageUp/Down scroll the screen, but if you then use an arrow key, you scroll back, whether you have a selection or not.)
> 

I think that's a macOS specific behaviour. On my macOS Firefox, the GH comment editor scrolls on pageup/down but don't move the caret. On my Linux Firefox, pageup/down scrolls and moves the caret. I believe, but didn't check, that Windows behave as Linux.

> So I think I could make that change, at the expense of having PageUp/Down scroll the editor but not move the caret. Or having PageUp/Down remove the selection. The latter would be easier to implement, I think, but also I think I prefer the former behavior.
> 

Judging by the kitchen example it seems Gio implements the Linux behaviour, so I'd expect your latter selection behaviour for consistency. Note that I don't mind a future change that changes behaviour according to runtime.GOOS.

widget/editor.go
eliasnaur: Maybe I'm being dense, but judging from the GH comment editor in macOS, shift+arrow really only move the selection end regardless of where the end was to begin with. That is, assuming selectionEnd = selectionStart = caretPos for the normal case of nothing selected.

eliasnaur (changes requested):

widget/editor.go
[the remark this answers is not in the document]
theclapp: >  The main point was that I'd like to always assume a selection, and move the zero-width selection special cases as low-level as possible

Sure, I'm fine with that.  Thank you for clarifying, though.

> > A problem with this idea is that PageUp/Down move the caret but not the selection.
>
> I'm not sure I see the problem, can you elaborate?

If selection is necessarily relative to the caret, because selection is defined by the caret + a signed width, then the caret must be either at the beginning or end of the selection.

But in the kitchen app, with this patch currently as-is, PageUp & PageDown move the caret completely outside the selection, without changing the selection.

So if selection is defined as caret + width, then one way or the other the behavior of PageUp/Down needs to change, either moving the caret and thus extending the selection, or by moving the caret and removing the selection, or by scrolling the text but not actually moving the caret.

So having written all that, and rereading your response, I think you're saying that PageUp/Down (that is, `e.movePages(+/-1)`) should move the caret and dismiss the selection, and that I should add another method (in addition to `movePages`) along the lines of `movePagesWithSelection`.  (Or possibly a `select` flag to `movePages`.)

> Judging by the kitchen example it seems Gio implements the Linux behaviour, so I'd expect your latter selection behaviour for consistency.

Okay.

I'm thinking `func (e *Editor) movePages(pages int, select bool)`, which will always move the caret, and optionally create/move/extend the selection.  And also change the calling code to be something like

```
	case key.NamePageUp:
		e.movePages(-1, k.Modifiers == key.ModShift)
	case key.NamePageDown:
		e.movePages(+1, k.Modifiers == key.ModShift)
```

widget/editor.go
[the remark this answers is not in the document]
theclapp: Hahaha, it's good that you posted that video, because otherwise I think we'd've assumed we were talking past each other.  But nope, you and I actually see different behavior.

aaa sss ddd

https://user-images.githubusercontent.com/2324697/103779959-633c8780-5002-11eb-9752-604c8cbc7099.mov

> I stand by my point to move the selection awareness to the lower-level Editor operations

Sure, I'm fine with that.

I also stand by my point that my desired behavior is to move the selection endpoints depending on what you clicked closest to.  I like that behavior and I think it's most convenient for users.

That said, just to be clear, I'm willing to do it however you want me to do it.  :)  The current patch implements the "extend" behavior; give it a try and see what you think.  (Or, obviously, if you have already and know that you hate it, just tell me what you want, that's fine too.)

widget/editor.go
theclapp: Ah.  For the normal case of nothing selected, that is true, I set the selection start to where the cursor was and the selection end to where the cursor is.  For the case where you do have something selected, you should move the last side that you moved (`*e.dragSide`), which (given the rest of the code as it is now) could be the front of the selection.  Consider if you have nothing selected, type shift-down-arrow, and then shift-click before the start of the selection, and then say shift-up-arrow (this video shows more than just that):

https://user-images.githubusercontent.com/2324697/103782559-9cc2c200-5005-11eb-983d-9f8fe978bf63.mov

theclapp:
> > You mentioned "this is the first round of review comments" [...]
> 
> Specifying the color is best left for the theme

I agree.

> which in your case probably means export paintSelection as PaintSelection, calling it from material.Editor. Just like PaintText, PaintSelection would just use op.PaintOp and not specify a color.

I'm not sure I know how to do that.  (I'm not saying it's impossible, just that I don't know how.)  `paintSelection` calls `paint.FillShape`, which takes a color explicitly.  And where would the color come from in the first place?  Can I just add `SelectionColor` to `EditorStyle`, and default it to `Theme.Palette.ContrastBg` in widget/material.Editor(), and pass it to `PaintText`, where it's used in the `paintSelection` color?  Also, making `paintSelection` more stand-alone would appear to require iterating over `e.shapes` again (unless I'm missing something), which seems wasteful.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: The difference seems to be our browsers: Firefox has the move-endpoints behavior, while Chrome has your extend behavior. I suspect move-endpoints is consistent with Linux/Windows while extend is consistent with macOS. I tried macOS' TextEdit and it also extends.

Given the inconsistencies, and your preference, please do keep the extend behaviour. We can always add a runtime.GOOS check for the move-endpoints behaviour later.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: > > I'd expect the usual scroll-to-caret behaviour regardless of the selection.
> 
> Neither the Gmail editor nor the Github comment editor do that. That is, if you have no text selected and scroll away from the caret and press Cmd-X, the editor is not scrolled back to the caret (and the caret is not moved).
> 
> The two differ with respect to whether a Cut or Copy clear the clipboard if you have nothing selected. Gmail clears it, Github doesn't. I prefer the latter behavior (which, um, is why I wrote it to work that way :).
> 

Ok. Implement the behaviour you like (as long as it matches some OS, presumably macOS) and we'll defer matching the behaviours of other OS for another time.

> > I suggest replacing IsSelected here with a check for SelectedText != "" below.
> 
> I'm not sure what you mean. I think you mean: scroll back to the caret, and don't update the clipboard if there's nothing selected?

I just meant that IsSelected should be deleted, and that you can check for SelectedText != "" (or SelectedLength == 0) instead.

widget/editor.go
eliasnaur: Indeed, this seems like another Chrome/Firefox difference. Please keep your preferred behaviour.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: > > The main point was that I'd like to always assume a selection, and move the zero-width selection special cases as low-level as possible
> 
> Sure, I'm fine with that. Thank you for clarifying, though.
> 
> > > A problem with this idea is that PageUp/Down move the caret but not the selection.
> > 
> > 
> > I'm not sure I see the problem, can you elaborate?
> 
> If selection is necessarily relative to the caret, because selection is defined by the caret + a signed width, then the caret must be either at the beginning or end of the selection.
> 
> But in the kitchen app, with this patch currently as-is, PageUp & PageDown move the caret completely outside the selection, without changing the selection.
> 
> So if selection is defined as caret + width, then one way or the other the behavior of PageUp/Down needs to change, either moving the caret and thus extending the selection, or by moving the caret and removing the selection, or by scrolling the text but not actually moving the caret.
> 

Indeed, that's one reason I'd like selection-awareness as low-level as possible: so that we're forced to handle all cases, including how to deal with the selection with pageup/down. And by defining caretPos to always be equal selectionStart, we've eliminated one source of inconsistency. (Again, it's up to you whether you like to represent the selection with and absolute selectionEnd or a relative, signed, selectionWidth).

> So having written all that, and rereading your response, I think you're saying that PageUp/Down (that is, `e.movePages(+/-1)`) should move the caret and dismiss the selection, and that I should add another method (in addition to `movePages`) along the lines of `movePagesWithSelection`. (Or possibly a `select` flag to `movePages`.)
> 

Moving the caret and dismissing the selection (or setting selectionEnd to caretPos) SGTM. However,
since you're implementing the macOS behaviour for selection anyway, I'm also ok with you changing the behaviour of pageup/pagedown to just scroll the Editor without moving the caret. If you do, please make the pageup/down behaviour change in a separate commit and place it first in your commit stack (for example by using `git rebase -i origin/main`).

If you choose to keep the current pageup/down behaviour of moving the caret, then adding a flag to movePages sounds best.

> > Judging by the kitchen example it seems Gio implements the Linux behaviour, so I'd expect your latter selection behaviour for consistency.
> 
> Okay.
> 

See above. You're free to implement the behaviour you like best.

> I'm thinking `func (e *Editor) movePages(pages int, select bool)`, which will always move the caret, and optionally create/move/extend the selection. And also change the calling code to be something like
> 
> ```
> 	case key.NamePageUp:
> 		e.movePages(-1, k.Modifiers == key.ModShift)
> 	case key.NamePageDown:
> 		e.movePages(+1, k.Modifiers == key.ModShift)
> ```

Right, SGTM (again assuming you keep the current pageup/down behaviour).

eliasnaur (changes requested):
> > which in your case probably means export paintSelection as PaintSelection, calling it from material.Editor. Just like PaintText, PaintSelection would just use op.PaintOp and not specify a color.
> 
> I'm not sure I know how to do that. (I'm not saying it's impossible, just that I don't know how.) `paintSelection` calls `paint.FillShape`, which takes a color explicitly.

See Editor.PaintText that uses raw clip and PaintOps for painting the text with whatever material is current.

> And where would the color come from in the first place? Can I just add `SelectionColor` to `EditorStyle`, and default it to `Theme.Palette.ContrastBg` in widget/material.Editor()

SGTM

> and pass it to `PaintText`, where it's used in the `paintSelection` color? Also, making `paintSelection` more stand-alone would appear to require iterating over `e.shapes` again (unless I'm missing something), which seems wasteful.

I don't mind the waste for now; it's only iterating the visible lines of text once more. We can get clever in a follow-up, for example by storing into e.selectedShapes just the e.shapes that need selection painted (usually none), or by constructing the selection area as one large clip operation.

widget/editor.go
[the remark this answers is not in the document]
theclapp: A lot of the discussion above was predicated on using a caret+signed length representation for the selection, which I'm not doing, so that part is moot.  I have pushed the selection-aware bits down as far as they can go, and generalized the API to address selection in a universal, consistent manner (I hope, haha).  Please take a look.

theclapp:
Pushed some new code, marked some more conversations as resolved.

Still more work to be done, but getting there.

Pushing the selection-aware code down as far as it could go does in fact make the code cleaner and more streamlined.  Thanks for the suggestion.

theclapp:
I think I've addressed all comments / objections / suggestions so far.  PTAL.  Thank you for your feedback.

widget/editor.go
```diff
@@ -259,75 +368,99 @@ func (e *Editor) processKey(gtx layout.Context) {
...
 			e.scroller.Stop()
 			e.append(ke.Text)
+		// Complete a paste event, initiated by Shortcut-V in Editor.command().
 		case clipboard.Event:
 			e.caret.scroll = true
 			e.scroller.Stop()
 			e.append(ke.Text)
 		}
 		if e.rr.Changed() {
+			e.ClearSelection()
```
eliasnaur, line 378: If e.rr.Changed, why didn't the changer also update (clear) the selection?

widget/editor.go
```diff
@@ -99,6 +112,39 @@ type maskReader struct {
...
+
+	// Pixel coordinates
+	x fixed.Int26_6
+	y int
+}
+
+type SelectionAction int
+
+const (
+	SelectionDoNotChange SelectionAction = iota
```
eliasnaur, line 133: I understand SelectionDoNotChange may be needed internally, but when would you use it externally? It seems to me that the selection should always be updated somehow when moving the caret.

Perhaps the caret should be defined as a "fat caret" (start, end) and Editor should have two move methods, MoveCaret and MoveCaretEnd?

widget/editor.go
```diff
@@ -447,14 +584,32 @@ func (e *Editor) layout(gtx layout.Context) layout.Dimensions {
 	return layout.Dimensions{Size: e.viewSize, Baseline: e.dims.Baseline}
 }
 
+// PaintSelection paints the contrasting background for selected text.
+func (e *Editor) PaintSelection(gtx layout.Context) {
+	cl := textPadding(e.lines)
+	cl.Max = cl.Max.Add(e.viewSize)
+	clip.Rect(cl).Add(gtx.Ops)
+	for _, shape := range e.shapes {
+		if shape.selected {
```
eliasnaur, line 593: Nit: the usual Go style is early exit and less indentation:

```
if !shape.selected {
    continue
}
```

widget/editor.go
```diff
@@ -648,13 +845,28 @@ func (e *Editor) invalidate() {
...
+// If there is a selection, it is deleted, and counts as a single rune. As a
+// special case, if runes == 0, deletes the selection (if any) and returns.
 func (e *Editor) Delete(runes int) {
+	if l := e.selection.len(); l > 0 {
+		e.rr.caret = min(e.selection.start.ofs, e.selection.end.ofs)
+		e.rr.deleteRunes(l)
+		e.ClearSelection()
+		runes -= sign(runes)
+		if runes == 0 {
+			e.invalidate()
```
eliasnaur, line 858: I suggest `defer e.invalidate()`; I don't see an exit where it isn't called.

widget/editor.go
```diff
@@ -670,14 +882,15 @@ func (e *Editor) prepend(s string) {
 	if e.SingleLine {
 		s = strings.ReplaceAll(s, "\n", " ")
 	}
+	e.Delete(0) // Delete any selection first. Might move the caret.
```
eliasnaur, line 885: Why can e.Delete(0) "move the caret"?

widget/editor.go
```diff
@@ -936,6 +1171,153 @@ func (e *Editor) NumLines() int {
...
+func (e *Editor) SetSelection(start, end int) {
+	e.selection.start.ofs, e.selection.end.ofs = start, end
+	e.selection.makeValid(e)
+}
+
+// SelectedText returns the currently selected text (if any) from the editor.
+func (e *Editor) SelectedText() string {
+	start, end := sortInts(e.selection.start.ofs, e.selection.end.ofs)
+
+	// Do I really need these?
```
eliasnaur, line 1195: Well do ya'? :)

widget/editor.go
```diff
@@ -936,6 +1171,153 @@ func (e *Editor) NumLines() int {
 	return len(e.lines)
 }
 
+func (e *Editor) IsSelected() bool {
```
eliasnaur, line 1174: Do you need IsSelected externally? If not, I'd lean towards deleting it entirely, or perhaps replace it with SelectionLen(gth).

eliasnaur (changes requested):

widget/editor.go
```diff
@@ -99,6 +112,39 @@ type maskReader struct {
...
+
+type SelectionAction int
+
+const (
+	SelectionDoNotChange SelectionAction = iota
+	SelectionClear
+	SelectionExtend
+)
+
+type EditorOption struct {
```
eliasnaur, line 138: Is EditorOption overkill? What other options do you anticipate?

widget/editor.go
```diff
@@ -936,6 +1171,153 @@ func (e *Editor) NumLines() int {
...
+
+// Selection returns the start and end of the selection, as offsets into the
+// editor text. start can be > end.
+func (e *Editor) Selection() (start, end int) {
+	return e.selection.start.ofs, e.selection.end.ofs
+}
+
+// SetSelection sets the selection start and end, which are offsets into the
+// editor text.
+func (e *Editor) SetSelection(start, end int) {
```
eliasnaur, line 1186: SetSelection is maybe better named MoveAbs?

widget/editor.go
theclapp: 😆 I suspect I don't any more, now that those fields are not exported any more.

widget/editor.go
theclapp: I don't anticipate any other options.  It does seem a little overkill for that one option, I do agree, and moreover thought that already.

On the other hand, I was leery of changing the public API for Editor.Move, and I really didn't want to make the selection action a required argument, and a functional argument seemed like the best compromise.  Also I thought if we ever decided to export any of the other moveXXX functions (moveLines, etc), which seems possible, then the scaffolding would already be there to let them accept functional arguments too.

I think performance-wise the impact is minimal.

widget/editor.go
theclapp: How so?  I assume that's short for "move absolute".  Move what?  I guess it would move the selection, sure, and it's "absolute" in that it's not relative to the window or the line layout or what-have-you.  But no, I think for what it does, SetSelection is the appropriate name.

Maybe this is predicated on making the selection be caret + offset?  (Which your earlier question "Is caret.pos ever != selection.start?" sort of alludes to.) 
 In which case setting the selection would have to (be prepared to) move the caret.  Even then I think I'd prefer SetSelection, especially given the other function Selection to get the selection offsets.

widget/editor.go
```diff
@@ -89,6 +94,14 @@ type Editor struct {
 	prevEvents int
 }
 
+// selection describes the start and stop positions of selected text. Note
+// that it's possible (and common) that End < Start, e.g. if the user selects
+// from later-to-earlier in the text, e.g. with Shift-LeftArrow.
+type selection struct {
+	start, end combinedPos
+	dragSide   *combinedPos
```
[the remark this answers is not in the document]
theclapp: > Is selection.dragSide != nil always equal to dragging?

No, dragging is used only for mouse events.  I also use dragSide for Shift-arrow movement and don't set dragging in that case.  Moreover dragSide should never be nil, as near as I can tell.  I never check for dragSide != nil.

widget/editor.go
```diff
@@ -69,15 +69,20 @@ type Editor struct {
...
+		// pos.(x, y) are the caret coordinates.
+		// pos.ofs is the caret offset into the text, from rr.caret
+		//
+		// Use Editor.caretPos to copy this field, e.g. in selection-related
+		// code. caret.pos.ofs is not always kept up-to-date with rr.caret;
+		// caretPos() updates it.
+		pos combinedPos
 	}
 
+	selection selection
```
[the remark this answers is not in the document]
theclapp: Yes.  Sometimes it == selection.end.  Also, after Shortcut-A to "select all", it doesn't equal start _or_ end.  (Possibly that should be changed.)  It used to be the case that PgUp & PgDown could move the caret w/out changing the selection, but that's no longer true: now PgUp/Down clear the selection and Shift-PgUp/Down extends it.

In general I've chosen to have the selection start and end mean what they say: start is where you started, and end is where you ended.  So if you press Shift-UpArrow on line 10, the start will be on line 10, and the end will be on line 9, and the caret.pos will also be on line 9.  There is no invariant enforced that start <= end.  Shift-clicking can swap start & end if they were "backwards", but I think that's the only thing that can.

widget/editor.go
theclapp: Good point.  Will remove.

widget/editor.go
theclapp: Good catch.  That comment descends from when PgUp/Down could move the caret without dismissing the selection, which they can't any more, so I'll remove it.

widget/editor.go
eliasnaur: Ok, let's leave the exported names for now.

> In which case setting the selection would have to (be prepared to) move the caret. Even then I think I'd prefer SetSelection, especially given the other function Selection to get the selection offsets.

Yes, I very much expect the caret position to change when calling SetSelection (and vice versa). I don't want the user to have to worry about the caret and selection coming out of sync. That's the reason behind my "MoveAbs" suggestion, and the reason for my wanting to merge the selection positions with the caret position.

widget/editor.go
eliasnaur: Still, EditorOption seems overkill (in API, not in performance loss) just to avoid an argument. A specialized method seems a better choice.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: I see, the pointer confused me. Perhaps a bool or enum is a better representation? With a pointer it seems like you could point to any combinedPos, not just start or end.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: To be clear, I'm focused on minimizing [illegal states](https://fsharpforfunandprofit.com/posts/designing-with-types-making-illegal-states-unrepresentable/), not as much on the eventual representation. I aim to eliminate one of the three positions (pos, selection.start, selection.end), and I can't think of a situation where pos is not equal to either selection.start or selection.end. In fact, I'm having trouble imagining pos is not *always* selection.end (or selection.start depending on your definition).

> Yes. Sometimes it == selection.end.

Given your description below, when is pos != selection.start then? 

>Also, after Shortcut-A to "select all", it doesn't equal start or end. (Possibly that should be changed.) It used to be the case that PgUp & PgDown could move the caret w/out changing the selection, but that's no longer true: now PgUp/Down clear the selection and Shift-PgUp/Down extends it.

It sounds like Shortcut-A should be changed to maintain the caret at start or end as you suspect.

> In general I've chosen to have the selection start and end mean what they say: start is where you started, and end is where you ended. So if you press Shift-UpArrow on line 10, the start will be on line 10, and the end will be on line 9, and the caret.pos will also be on line 9. There is no invariant enforced that start <= end. Shift-clicking can swap start & end if they were "backwards", but I think that's the only thing that can.

That sounds like a great representation for start and end. I'm trying to figure out whether pos can always be one of them, and perhaps always equal end?

eliasnaur (changes requested):

widget/editor.go
theclapp: Looking at this some more, I see that SetSelection, as it stands, breaks the usual invariant that the caret is at one side or the other of the selection, and that it should probably set the editor caret, too.  But to which side?  I guess that's your point: It should be renamed to MoveAbs, accept an offset and a signed length, move the caret to the offset, and set the selection start & end based on the length.

This might be a good place for another editor option to Move.  By default, Move should move relative the current caret, and update the selection along with it, just like it does now.  But I could add a MoveAbsolute flag, where "distance" is then interpreted as an absolute offset, and also a Selection length, which would update the selection.

That would also let you dispense with SetSelection, which would now be something more like `Move(0, WithSelection(<length>))`.

theclapp: (Looks like we were commenting at the same time.)

theclapp:
I'm going to work on refactoring the selection code so that the caret is always at selection end.  I think that will, for example, make dragSide go away, and in general make the rest of the selection code cleaner, and (as you've said) reduce the chance of illegal states.

theclapp:
More updates & a force-push.  Rebased from main.  I didn't get an email, I assume because of the forced push.  PTAL.

widget/editor.go
```diff
@@ -69,15 +69,25 @@ type Editor struct {
...
-		line int
-		// col is the caret column measured in runes.
-		col int
-		// (x, y) are the caret coordinates.
-		x fixed.Int26_6
-		y int
+		// pos.lineCol.Y is the caret line position as an index into lines.
+		// pos.lineCol.X is the caret column measured in runes.
+		// pos.(x, y) are the caret coordinates.
+		// pos.ofs is the caret offset into the text, from rr.caret
```
eliasnaur, line 75: These comments, while useful, seem redundant or misplaced. Why not have them on combinedPos?

widget/editor.go
```diff
@@ -69,15 +69,25 @@ type Editor struct {
...
+		//
+		// Use Editor.copyCaret to copy this field, e.g. in selection-related
+		// code. caret.pos.ofs is not kept up-to-date with rr.caret; copyCaret()
```
eliasnaur, line 78: Similar to the previous question about redundant combinedPos values, can rr.caret always be derived from pos or selStart? If so, can it be deleted?

widget/editor.go
```diff
@@ -145,9 +177,20 @@ type SubmitEvent struct {
...
+type SelectEvent struct{}
+
 type line struct {
-	offset image.Point
-	clip   op.CallOp
+	offset         image.Point
+	clip           op.CallOp
+	selected       bool
+	selectionYOffs int
+	selectionDims  layout.Dimensions
```
eliasnaur, line 193: You don't seem to use selectionDims.Baseline anywhere. Reduce to just an image.Point.

widget/editor.go
```diff
@@ -223,8 +279,48 @@ func (e *Editor) processPointer(gtx layout.Context) {
...
+				e.moveWord(-1, SelectionClear)
+				e.moveWord(1, SelectionExtend)
+				e.dragging = false
+			}
+		}
+	}
+
+	// e.dragger gets the same Release events as e.clicker, so if e.clicker
+	// processed a double-click event, we need to ignore the Release events
+	// here.
```
eliasnaur, line 305: This comment suggests a design problem in gestures, or even in pointer event resolution. Can you think of a way to change Dragger, or something else to support composing gestures better? I find the logic hard to follow.

widget/editor.go
```diff
@@ -270,64 +367,87 @@ func (e *Editor) processKey(gtx layout.Context) {
 	}
 }
 
-func (e *Editor) moveLines(distance int) {
-	e.moveToLine(e.caret.x+e.caret.xoff, e.caret.line+distance)
+func (e *Editor) moveLines(distance int, selectionAction SelectionAction) {
```
eliasnaur, line 370: Nit: selectionAction is long; I suggest act or selAct.

widget/editor.go
```diff
@@ -447,14 +571,33 @@ func (e *Editor) layout(gtx layout.Context) layout.Dimensions {
...
+	clip.Rect(cl).Add(gtx.Ops)
+	for _, shape := range e.shapes {
+		if !shape.selected {
+			continue
+		}
+		stack := op.Save(gtx.Ops)
+		offset := shape.offset
+		offset.Y += shape.selectionYOffs
+		op.Offset(layout.FPt(offset)).Add(gtx.Ops)
+		clip.Rect(image.Rect(0, 0, shape.selectionDims.Size.X, shape.selectionDims.Size.Y)).Add(gtx.Ops)
```
eliasnaur, line 587: Nit: image.Rectangle{Max: shape.selectionDims[.Size]} is a bit shorter.

widget/editor.go
```diff
@@ -850,21 +1055,34 @@ func (e *Editor) moveWord(distance int) {
...
 }
 
-// deleteWord the next word(s) in the specified direction.
+// deleteWord deletes the next word(s) in the specified direction.
 // Unlike moveWord, deleteWord treats whitespace as a word itself.
 // Positive is forward, negative is backward.
 // Absolute values greater than one will delete that many words.
+//
+// If there is a selection, it is deleted and counts as a single word. As a
+// special case, if distance == 0, deletes the selection (if any) and returns.
```
eliasnaur, line 1074: I'd expect distance == 0 to always be a no-op. What's the reason for preferring the selection-clearing behaviour?

eliasnaur (commented):
Much simpler, thank you! I think we're getting close to merging this.

The main issue in this review round is the mismatch between e.rr.caret and the combinedPos values in Editor. I'd like to do something about the ugly interaction between Clicker and Dragger, but it's not necessarily blocking this work.

widget/editor.go
theclapp: This particular case in deleteWord was only so that it would have the same behavior as Delete.

In the case of Delete, it was because I wanted to be able to call Delete unconditionally to only delete the selection and not delete anything if there was no selection, in Editor.prepend, and also because as an API user I suspect I'd want that functionality.

In Editor.prepend, I could've make the call conditional on SelectionLen() > 0, but it bugged me that then I'd be doing that comparison twice, once in Editor.prepend, and then again in Delete.

I could make DeleteSelection() that returns true if something was actually deleted and use that in Editor.prepend and also in deleteWord, and then Delete and deleteWord could both be complete no-ops for 0.

eliasnaur: > This particular case in deleteWord was only so that it would have the same behavior as Delete.
> 
> In the case of Delete, it was because I wanted to be able to call Delete unconditionally to only delete the selection and not delete anything if there was no selection, in Editor.prepend, and also because as an API user I suspect I'd want that functionality.
> 

If you don't have a clear external use yet, I'm inclined towards the "0 means no-op" behaviour. What I imagine someone doing is something like

var deleteCounter int
var e Editor
for ... {
   ...
   deleteCounter++
   ...
}
e.Delete(deleteCounter)

This code works as expected for the usual case (no selection), but subtly fails when something is selected.

> In Editor.prepend, I could've make the call conditional on SelectionLen() > 0, but it bugged me that then I'd be doing that comparison twice, once in Editor.prepend, and then again in Delete.
> 

That sounds more like prepend/append should be merged, inlined or something else. FWIW, append/prepend are not my favorites either.

> I could make DeleteSelection() that returns true if something was actually deleted and use that in Editor.prepend and also in deleteWord, and then Delete and deleteWord could both be complete no-ops for 0.

DeleteSelection may be needed, but how often? I'd say a e.DeleteRange(e.Selection()) is a better way to expose your Delete(0). Returning a bool seems ugly, though. If you're interested in the result, why not check SelectionLen first?

widget/editor.go
theclapp: So far, I don't see a good way to do either of those things.

editorBuffer uses caret in 16 places.  Editor uses rr.caret in 40 places.  It seems like editorBuffer needs to keep caret, and, at most, Editor.caret.pos.ofs should somehow be made an alias to it.  I could make it a pointer to rr.caret, but that makes other uses of combinedPos annoying to deal with.  And actually makes copying caret.pos just as problematic, since I'd need to allocate a new pointer, so I'd certainly write a function to do it.  So that doesn't really help.

I could pull Editor.caret.pos back out into its component pieces again, so pos.ofs didn't even exist.  That would make using offsetToScreenPos a little more awkward, since I'd need code to deal with Editor.caret and Editor.selStart differently.  On the other hand I only call that twice, and I could write a helper function to encapsulate that operation.

I could make sure I update caret.pos.ofs every time rr.caret changes, and I think I tried that, but it seemed likely that later contributors wouldn't realize they needed to do the same and pos.ofs and rr.caret could easily get out of sync.

Thoughts / suggestions?

widget/editor.go
theclapp: > If you don't have a clear external use yet, I'm inclined towards the "0 means no-op" behaviour

Okay, I'll change it.

widget/editor.go
theclapp: Yeah, the logic isn't the clearest.

I'm going to try something like

```go
for _, evt := range append(e.clicker.Events(gtx), e.dragger.Events(gtx.Metric, gtx, gesture.Both))
```

and see if that works and is clearer.

theclapp: Well, something similar works:

```go
	for _, evt := range e.clickDragEvents(gtx) {
		switch evt := evt.(type) {
		case gesture.ClickEvent:
                // ...
		case pointer.Event:
                // ...
		}
	}
       // ...

func (e *Editor) clickDragEvents(gtx layout.Context) []event.Event {
	var combinedEvents []event.Event
	for _, evt := range e.clicker.Events(gtx) {
		combinedEvents = append(combinedEvents, evt)
	}
	for _, evt := range e.dragger.Events(gtx.Metric, gtx, gesture.Both) {
		combinedEvents = append(combinedEvents, evt)
	}
	return combinedEvents
}

File: widget/editor.go (hunk 28 above)
Comment (theclapp) [reply widget/editor.go]: Refactored this extensively.  PTAL.

File: widget/editor.go (hunk 23 above)
Comment (eliasnaur) [reply widget/editor.go:74]: Because Editor needs the richest caret information, I believe the best way forward is to eliminate editBuffer.caret. It's correct that there are 16 references to caret in editBuffer, but a closer look reveals that only 2 methods need caret: moveGap and prepend (deleteRunes only update caret). Given that, adding a position argument to moveGap and prepend doesn't seem unreasonable.

**Review by eliasnaur — CHANGES_REQUESTED** [review]

File: widget/editor.go [hunk 29]
```diff
@@ -326,6 +323,17 @@ func (e *Editor) processPointer(gtx layout.Context) {
...
+	var combinedEvents []event.Event
+	for _, evt := range e.clicker.Events(gtx) {
+		combinedEvents = append(combinedEvents, evt)
+	}
+	for _, evt := range e.dragger.Events(gtx.Metric, gtx, gesture.Both) {
+		combinedEvents = append(combinedEvents, evt)
+	}
+	return combinedEvents
+}
+
```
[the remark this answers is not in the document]
theclapp: 🎉

widget/editor.go
```diff
@@ -758,67 +764,61 @@ func (e *Editor) CaretCoords() f32.Point {
...
-		for _, adv := range l.Layout.Advances {
-			for idx == offsets[0] {
-				positions, offsets = appendPosition(positions, offsets, e, line, col, x, y)
-				if len(offsets) == 0 {
-					return positions
+	l := e.lines[line]
+	y := l.Ascent.Ceil()
+	prevDesc := l.Descent
+
+	iter := func(offset int) combinedPos {
```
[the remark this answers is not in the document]
theclapp: No, it's only called in makeValid (after the check of e.valid, natch) and for Shortcut-A ("select all"), so steady-state it's not called.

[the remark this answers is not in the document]
theclapp: As you surmised / implied, though, the closure does leak to the heap when it (offsetToScreenPos) _is_ called.  Given that makeValid also calls layoutText, I'm gonna guess that adding another closure will be a drop in the bucket.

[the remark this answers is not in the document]
eliasnaur: Indeed, I don't care (much) about generating garbage when updating state which is rare compared to redraws and localized to the changing component(s).

theclapp:
PTAL.

eliasnaur (changes requested):
Thanks for removing the buffer caret. I think the internal changes looks good, and this review round is about the exported API. Thank you for your patience.

We're getting close to merge, so I'm going to bring up the eventual commit order. As you know, I prefer each commit to be a well-defined and self-contained change, especially for changes to complex code such as Editor (your PR includes >600 lines of change to editor.go alone).

For the final PR, I suggest moving all the refactoring into separate commits, thus isolating the functional changes for smaller commits. The result is better history and easier review and later bisecting. So:

- One or more commits for the removal of editBuffer.caret, introduction of Editor.pos.combinedPos, API changes etc. They should contain only mechanical changes, and no functional changes.
- One or more commits for functional changes or fixes not directly related to selection (if any).
- One or more commits adding selection support and API.

widget/editor.go
```diff
@@ -99,6 +102,27 @@ type maskReader struct {
...
+	x fixed.Int26_6
+	y int
+}
+
+type SelectionAction int
+
+const (
+	SelectionClear SelectionAction = iota
+	SelectionExtend
+)
```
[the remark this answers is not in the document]
theclapp:
```go
> // Move is the relative variant of Move. It moves the caret end bytes from its current position, and
> // the selection start bytes from its current position.
> func (e *Editor) Move(start, end int)
```

I assume you mean `Move is the relative variant of MoveAbs`?

TBH I haven't looked closely at CaretCoord or CaretPos, since they seemed mostly unrelated to what I was doing.

It does seem like if we have Caret() which gets the caret and selection positions, most other APIs would have SetCaret (i.e. Thing() to get "thing(s)" and SetThing(...) to set it/them).  So what do you think of that?  Caret() to get and SetCaret() (replacing Move) for relative movement, and SetCaretAbs() (replacing MoveAbs) for absolute movement?

I have some concerns about Move()/SetCaret(), but I think we need to get some feedback on how people want to use it, so I'm willing to do as you suggest right now.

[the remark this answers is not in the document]
eliasnaur: > ```go
> > // Move is the relative variant of Move. It moves the caret end bytes from its current position, and
> > // the selection start bytes from its current position.
> > func (e *Editor) Move(start, end int)
> ```
> 
> I assume you mean `Move is the relative variant of MoveAbs`?
> 

Yes. Sorry about that.

> TBH I haven't looked closely at CaretCoord or CaretPos, since they seemed mostly unrelated to what I was doing.
> 

Me neither, but I noticed that you added more API and had to consider whether it fits.

> It does seem like if we have Caret() which gets the caret and selection positions, most other APIs would have SetCaret (i.e. Thing() to get "thing(s)" and SetThing(...) to set it/them). So what do you think of that? Caret() to get and SetCaret() (replacing Move) for relative movement, and SetCaretAbs() (replacing MoveAbs) for absolute movement?
> 

I wouldn't go so far in normalizing method names. For example SetCaret sounds like your SetCaretAbs to me and is a nicer name. What about "SetCaret" (absolute) and "MoveCaret" (relative)?

> I have some concerns about Move()/SetCaret(), but I think we need to get some feedback on how people want to use it, so I'm willing to do as you suggest right now.

Selection is subtle enough, which is why I suggested keeping Selection unexported and leaving the rest of the API alone, postponing the xported API to future changes. However, I can understand why you want the API in this change: the exported API can often guide simplifications in the internal implementation.

widget/buffer.go
```diff
@@ -35,25 +33,26 @@ func (e *editBuffer) Changed() bool {
 	return c
 }
 
-func (e *editBuffer) deleteRunes(runes int) {
-	e.moveGap(0)
+func (e *editBuffer) deleteRunes(caret, runes int) int {
```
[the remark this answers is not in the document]
theclapp: I don't think that's true?  First, recall that `runes` is signed, and that `caret` only changes if `runes` < 0.  Second, `deleteRunes` deletes runes (of course), but `caret` is a byte offset.  So if we don't return the new `caret`, then in the caller we have to recalculate the new offset based on (the new value of) `e.rr.text[e.rr.gapstart]` and the following `abs(runes)` runes.

I don't think we can use `e.rr.gapend - e.rr.gapstart` because I don't see any guarantee that the gap is exactly 0 after calling moveGap, only that it starts at `caret`.

widget/editor.go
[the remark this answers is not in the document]
theclapp: > What about "SetCaret" (absolute) and "MoveCaret" (relative)?

SGTM.

> ... exported api discussion ...

We can't keep the entirety of the selection part of the api private because we use Move internally.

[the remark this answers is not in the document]
theclapp: The (new) `MoveCaret(startDelta, endDelta int)` (which replaces `Move(distance int, selAct SelectionAction)`) has me, I don't know, looking askance at it.  In particular, I don't like that the caret is the end of the selection and not the start.  Would it be awful if I swapped them?  That is, made the caret always the start of the selection and change `selStart` to `selEnd`?  WDYT?

theclapp:
> I'm going to bring up the eventual commit order ...

So, for example, a set of N commits that does refactoring only, _setting the stage_ for the selection code, possibly changing the API (e.g. changing Move to MoveCaret, and adding SetCaret), but adding no actual selection code or api, and then another set of N commits that does add the selection code, but makes no (or as few as possible) other changes.

So the first set would, for example, change `func (e *Editor) Move(distance int)` to `func (e *Editor) MoveCaret(caretDelta int)`, and then the second set would change it to `func (e *Editor) MoveCaret(startDelta, endDelta int)`.

Yes?

So ... would doing all that in place make the comments in this PR hard to read afterwards, or does GH know how to take care of all that?  And if the former, maybe I should open a new PR with a new branch with the reordered / combined / split commits?  Any advice?  I haven't done anything like that before.

theclapp:
Just for posterity: I suspect others might benefit from this as much as I did, which has sections on squashing, splitting, and reordering commits: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

theclapp:
One of the things I'm wondering, from you (@eliasnaur) or anyone else that happens to be following along, is: would the reorganization of the commits you're requesting be easier with an interactive rebase, with (I suspect) rather a lot of commit splitting and conflict resolution, or would it be easier if I just diffed the current branch against main, and created brand new commits in a new branch, with (I suspect) way fewer problems with commit splitting and conflict resolution.

I mean, to be sure, the whole diff is 1842 lines, but at least I'd only have to go over them approximately once each, instead of quite possibly several times each.  Like I said, looking for advice.

/cc @mvdan @whereswaldon if you have any advice.  If not, no worries & thanks anyway.  😄

mvdan:
Personally, I just rebase and force-push to PRs :) Whatever you do, "see older versions of this PR" is practically impossible on GitHub, so I go for the simpler approach of fewer commits.

widget/buffer.go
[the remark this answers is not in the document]
eliasnaur: You're absolutely right. Sorry for not thinking this through.

widget/buffer.go
```diff
@@ -138,17 +155,17 @@ func (e *editBuffer) String() string {
...
+	e.moveGap(caret, len(s))
+	copy(e.text[caret:], s)
 	e.gapstart += len(s)
 	e.changed = e.changed || len(s) > 0
-	e.dump()
+	e.dump(caret)
 }
 
-func (e *editBuffer) dump() {
+func (e *editBuffer) dump(caret int) {
```
eliasnaur, line 166: Nit: there's no need for the caret parameter.

widget/buffer.go
```diff
@@ -138,17 +155,17 @@ func (e *editBuffer) String() string {
...
-func (e *editBuffer) prepend(s string) {
-	e.moveGap(len(s))
-	copy(e.text[e.caret:], s)
+func (e *editBuffer) prepend(caret int, s string) {
+	e.moveGap(caret, len(s))
+	copy(e.text[caret:], s)
 	e.gapstart += len(s)
 	e.changed = e.changed || len(s) > 0
-	e.dump()
+	e.dump(caret)
```
eliasnaur, line 163: Nit: drop these calls to dump, and perhaps even bufferDebug or dump itself. I've not used it since implementing editBuffer.

widget/editor.go
```diff
@@ -648,36 +826,56 @@ func (e *Editor) invalidate() {
...
-	e.rr.prepend(s)
-	e.caret.xoff = 0
+	e.caret.pos.ofs = e.rr.deleteRunes(e.caret.pos.ofs, e.selStart.ofs-e.caret.pos.ofs) // Delete any selection first.
+	e.rr.prepend(e.caret.pos.ofs, s)
+	e.caret.pos.xoff = 0
 	e.invalidate()
 }
 
-func (e *Editor) movePages(pages int) {
+func (e *Editor) movePages(pages int, selAct selectionAction) {
```
eliasnaur, line 876: It's weird to have selectionAction in unexported API, and (start, end) offsets in exported API (MoveCaret). In fact, I find it weird to have different methods for similar actions (moving the caret). I suggest to replace them all with a single method:

```
type MoveUnit

const (
    MoveRune MoveUnit = iota
    MoveWord
    MovePage
    // Maybe?
    MoveLineStart
    MoveLineEnd
    // If you'd like to replace SetCaret as well.
    MoveAbsolute
)

// MoveCaret move the start and end of the caret start and end
// units, respectively.
func (e *Editor) MoveCaret(unit MoveUnit, start, end int)
```

With something like MoveAbsolute you could even get rid of SetCaret.

MoveUnit is not the greatest name, I know, and it's ok to postpone this API discussion for later.

eliasnaur (commented):
> One of the things I'm wondering, from you (@eliasnaur) or anyone else that happens to be following along, is: would the reorganization of the commits you're requesting be easier with an interactive rebase, with (I suspect) rather a lot of commit splitting and conflict resolution, or would it be easier if I just diffed the current branch against main, and created brand new commits in a new branch, with (I suspect) way fewer problems with commit splitting and conflict resolution.
> 
> I mean, to be sure, the whole diff is 1842 lines, but at least I'd only have to go over them approximately once each, instead of quite possibly several times each. Like I said, looking for advice.
> 

FWIW, I rarely start over from a diff to main, and use `git rebase -i main` almost exclusively to re-order, squash/fixup, edit commits and commit messages.

widget/editor.go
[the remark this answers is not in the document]
eliasnaur: Yes, please! I also found the caret being the end of the selection.

Perhaps we could go all the way and consider a caret "fat", having a start and end coordinate? That would fit better with the MoveCaret method, and further enforces the approach of always having a selection even though it is usually zero runes.

whereswaldon:
I guess my comments are nothing new here, but I also just use `git rebase -i` for this sort of thing. I have attempted the kind of clean slate history reconstruction that you describe (diffing main), but it's very difficult to retroactively isolate logical changesets with that approach.

eliasnaur:
> > I'm going to bring up the eventual commit order ...
> 
> So, for example, a set of N commits that does refactoring only, _setting the stage_ for the selection code, possibly changing the API (e.g. changing Move to MoveCaret, and adding SetCaret), but adding no actual selection code or api, and then another set of N commits that does add the selection code, but makes no (or as few as possible) other changes.
> 

Exactly (I'm shamelessly stealing the "setting the stage" metaphor). The goal is to minimize and isolate the exciting changes, to reduce review fatigue and for easier debugging in case a future bisect blames one of your commits.

> So the first set would, for example, change `func (e *Editor) Move(distance int)` to `func (e *Editor) MoveCaret(caretDelta int)`, and then the second set would change it to `func (e *Editor) MoveCaret(startDelta, endDelta int)`.
> 
> Yes?
> 

Yes.

> So ... would doing all that in place make the comments in this PR hard to read afterwards, or does GH know how to take care of all that? And if the former, maybe I should open a new PR with a new branch with the reordered / combined / split commits? Any advice? I haven't done anything like that before.

Like Daniel, I don't have much hope for retaining the comment history. Even if we did retain history, there's little linking commits to the GitHub PRs. If you feel generous, please do summarize relevant history in the commit messages that end up merged.

widget/editor.go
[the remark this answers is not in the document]
theclapp: > Perhaps we could go all the way and consider a caret "fat", having a start and end coordinate?

I feel like, operationally speaking, that's what we have now.  I guess we could make it more obvious / formal / explicit and change caret.pos to caret.start, and change/move selStart to caret.end.  That makes sense.

Also, in case it's not obvious: the big reason I've avoided only using an offset, and actually storing the line & column coordinates, and the x,y pixel coordinates, is for efficiency.  I figure that if I only stored offsets, then for every single layout we'd have to recalculate their coordinates within the widget, and that seemed like it could get prohibitive, especially with a lot of text, during a mouse drag or a resize, where we're doing layouts hand-over-fist.

widget/editor.go
theclapp: > it's ok to postpone this API discussion for later

Cool, then I will.  😄

theclapp:
Rebased from main before big commit reorganization.

widget/editor.go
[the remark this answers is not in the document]
theclapp: > > In particular, I don't like that the caret is the end of the selection and not the start. Would it be awful if I swapped them?
>
> Yes, please!

Done.

theclapp:
I've reorganized the commits into just two, one with just refactoring, and one with just selection code, as requested.  PTAL.

widget/editor.go
```diff
@@ -936,6 +1176,99 @@ func (e *Editor) NumLines() int {
...
+}
+
+// SetCaret moves the caret to start, and sets the selection end to end. start
+// and end are in bytes, and represent offsets into the editor text.
+func (e *Editor) SetCaret(start, end int) {
+	// Constrain start and end to [0, e.Len()].
+	l := e.Len()
+	start = max(min(start, l), 0)
+	end = max(min(end, l), 0)
+	e.caret.start.ofs, e.caret.end.ofs = start, end
```
eliasnaur, line 1198: What updates the rest of e.caret.start/end? MoveCaret updates them through Editor.movePos, but I don't see anything like that here.

eliasnaur (changes requested):
Thank you for rebasing the commits ,much appreciated.

I have a single concern about SetCaret left, otherwise LGTM.

widget/editor.go
theclapp: Fixed.  PTAL.

lrewega:
Trying out the latest commit 6502c73 I noticed that using arrow keys with highlighted text does not clear the highlight, as I would expect, but instead _slides_ the highlight, which is a behaviour I've never seen before!

eliasnaur:
> Trying out the latest commit [6502c73](https://github.com/gioui/gio/commit/6502c738650b0701fc3602326986efcc39b4d996) I noticed that using arrow keys with highlighted text does not clear the highlight, as I would expect, but instead _slides_ the highlight, which is a behaviour I've never seen before!

I'm seeing this too. It's an interesting effect, for sure :)

theclapp:
> using arrow keys with highlighted text does not clear the highlight, as I would expect, but instead slides the highlight

Yep.  Good catch!  This is only for the left & right arrows; up & down clear the selection as expected.

It was like this before the commit reorganization, too, apparently.

This is a direct result of the `MoveCaret(startDelta, endDelta int)` api.  There's no value of endDelta you can send to extend the selection if you've pressed Shift _and_ to clear it if you haven't.  A 0 will shrink/extend the selection, a +/-1 will slide/shrink/enlarge it, depending on the sign of startDelta.  (Not noticing this bug was a direct result of me always either clicking the mouse or using the up/down arrows during testing, I guess.  And insufficient test cases.)

@eliasnaur, what are your thoughts?  As we've discussed, we want the api to be selection-agnostic as far down as possible.  The options as I see them:
1. Change MoveCaret to accept another flag that says clear or extend the selection, like the internal moveWord/moveLines/etc, and ignore endDelta if it's selectionClear.  `MoveCaret(startDelta, endDelta int, selAct SelectionAction)`.
2. Call ClearSelection before MoveCaret if Shift is pressed
3. Have two functions: MoveCaret(startDelta) which calls ClearSelection internally and then moves the caret, and MoveCaretShifted(startDelta, endDelta) (or possibly ChangeSelection or something else?) which does what MoveCaret does now
4. Pass an out-of-band value in endDelta (e.g. math.MaxInt32) that's special-cased to clear the selection.

```go
const NoSelection = math.MaxInt32

func (e *Editor) MoveCaret(startDelta, endDelta int) {
  if endDelta == NoSelection { e.ClearSelection() }
  // ...
```

2-4 still require selection-specific code in the left/right arrow code, though, and also user code would have to do the same thing in a similar context.

I think I prefer option 2, with the option of changing to option 1 when/if we ever export the other movement functions (moveWord/moveLine/etc).

eliasnaur:
> > using arrow keys with highlighted text does not clear the highlight, as I would expect, but instead slides the highlight
> 
> Yep. Good catch! This is only for the left & right arrows; up & down clear the selection as expected.
> 
> This is a direct result of the `MoveCaret(startDelta, endDelta int)` api. There's no value of endDelta you can send to extend the selection if you've pressed Shift _and_ to clear it if you haven't. A 0 will shrink/extend the selection, a +/-1 will slide/shrink/enlarge it, depending on the sign of startDelta. (Not noticing this bug was a direct result of me always either clicking the mouse or using the up/down arrows during testing, I guess. And insufficient test cases.)
> 
> @eliasnaur, what are your thoughts? As we've discussed, we want the api to be selection-agnostic as far down as possible. The options as I see them:
> 
>     2. Call ClearSelection before MoveCaret if Shift is pressed

I believe you mean when Shift is *not* pressed?

> I think I prefer option 2

I prefer option 2 as well. I don't see a need for MoveCaret to precisely map to keyboard combinations, so a ClearSelection when moving without shift is fine.

> with the option of changing to option 1 when/if we ever export the other movement functions (moveWord/moveLine/etc).

I think we can fold moveWord/Line into a generalization of MoveCaret. See my https://github.com/gioui/gio/pull/3#discussion_r559469869 sketch.

theclapp:
> I believe you mean when Shift is _not_ pressed?

Right.  Oops.

> option 2

Awesome.

> I think we can fold moveWord/Line into a generalization of MoveCaret

For a later PR.

theclapp:
> using arrow keys with highlighted text does not clear the highlight

Fixed.

eliasnaur:
Great work, all merged! Thank you for your patience.

ddkwork:
also，we need implement column edit model

PR #4, opened 2020-12-17
internal/glimpl: [wasm] remove syscall/js

inkeliz:
That change avoid JS calls through `syscall/js`. Instead we use the
WebAssembly imports, which is defined on the `gl_js.s` and `gl_js.js`.

That change is similar to the previous patch, but it's easier to read
the code and fix. It also have only five JavaScript functions, instead
of one function for each WebGL call.

Opera 72 (w/ AMD Ryzen 3900X): ~8.0ms per frame to ~3.4ms;
Chrome 87 (w/ Snapdragon 435): ~83.2ms per frame to ~39.8ms;

The `gl_js.js` file is embed by the `gogio` compiler.

### DIFF
--- a/internal/gl/gen/gen.go
+++ b/internal/gl/gen/gen.go
@@ -1,0 +1,341 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
+)
+
+func translateArgType(s string) (f string, offset string) {
+	switch s {
+	case "bool":
+		return "gioLoadBool", "8"
+	case "string":
+		return "gioLoadString", "16"
+	case "int", "int64", "uint", "Attrib", "Enum", "uintptr", "unit64":
+		return "gioLoadInt64", "8"
+	case "int32", "uint32":
+		return "gioLoadInt32", "8"
+	case "[]byte":
+		return "gioLoadSlice", "24"
+	case "float", "float64", "float32":
+		return "gioLoadFloat64", "8"
+	default:
+		return "gioLoadObject", "8"
+	}
+}
+
+func translateResultType(s string) (f string, offset string) {
+	switch s {
+	case "int", "int64", "uint", "Attrib", "Enum", "uintptr", "bool":
+		return "gioSetInt64", "8"
+	case "[4]float32", "[4]int":
+		return "gioSetArray4", "32"
+	default:
+		return "gioSetObject", "8"
+	}
+}
+
+func main() {
+
+	data, err := ioutil.ReadFile("Z:\\gio\\gio-3\\internal\\gl\\gl_syscall_js.go")
+	if err != nil {
+		panic(err)
+	}
+
+	js, err := os.Create("Z:\\gio\\gio-3\\internal\\gl\\gl_unsafe_js.js")
+	if err != nil {
+		panic(err)
+	}
+
+	golang, err := os.Create("Z:\\gio\\gio-3\\internal\\gl\\gl_unsafe_js.go")
+	if err != nil {
+		panic(err)
+	}
+
+	asm, err := os.Create("Z:\\gio\\gio-3\\internal\\gl\\gl_unsafe_js.s")
+	if err != nil {
+		panic(err)
+	}
+
+	findFunctions := regexp.MustCompile(`\(f \*FunctionCaller\) (\w+)\((.*)\) (.*?){`).FindAllSubmatch(data, -1)
+	findCalls := regexp.MustCompile(`f.Ctx.Call\((.*)\)`).FindAllSubmatch(data, -1)
+
+	typeRemover := regexp.MustCompile(`(.*)\((\w+)\)`)
+
+	writeHeader(js)
+	writeGoHeader(golang)
+
+	asm.WriteString(`// SPDX-License-Identifier: Unlicense OR MIT
+
+#include "textflag.h"
+
+`)
+
+	for i, v := range findFunctions {
+		if strings.ToUpper(string(v[1][0])) != string(v[1][0]) {
+			continue
+		}
+		if len(findCalls[i][1]) == 0 {
+			continue
+		}
+		if strings.Contains(string(v[0]), "panic") {
+			continue
+		}
+
+		for _, v := range v {
+			fmt.Println(string(v))
+		}
+		for _, v := range findCalls[i] {
+			fmt.Println(string(v))
+		}
+
+		argsRaw := strings.Split(string(v[2]), ", ")
+
+		args := make([][2]string, len(argsRaw))
+
+		for i, a := range argsRaw {
+			aa := strings.Split(a, " ")
+			args[i][0] = strings.TrimSpace(aa[0])
+
+			if len(aa) > 1 {
+				args[i][1] = aa[1]
+
+				for it := 0; it < i; it++ {
+					if args[it][1] == "" {
+						args[it][1] = strings.TrimSpace(aa[1])
+					}
+				}
+			}
+		}
+
+		var argsJS string
+		var offset = "0"
+
+		fmt.Println(args, len(args))
+
+		jsArgs := strings.Split(string(findCalls[i][1]), ",")
+
+		asmArgs := make([]string, 0, len(jsArgs))
+		goArgsNames := make([]string, 0, len(jsArgs))
+		for _, a := range jsArgs {
+			t := ""
+			for _, aa := range args {
+				if cast := typeRemover.FindAllStringSubmatch(a, -1); len(cast) >= 1 {
+					a = cast[0][2]
+				}
+				if a == "ba" {
+					a = "data"
+				}
+				a = strings.Trim(strings.TrimSpace(a), `)`)
+				fmt.Println(a, aa[0])
+				if aa[0] == a {
+					t = aa[1]
+					break
+				}
+			}
+			isArray := false
+			if a == "f.int32Buf" && string(v[1]) == "InvalidateFramebuffer" {
+				//specal case
+				t = "Enum"
+				a = "attachment"
+				isArray = true
+			}
+			if t == "" {
+				switch a {
+				case "nil":
+					argsJS += fmt.Sprintf("\t\t\t\tundefined,\n")
+				case "0":
+					argsJS += fmt.Sprintf("\t\t\t\t0,\n")
+				}
+				continue
+			}
+			f, o := translateArgType(t)
+
+			if t == "float32" {
+				t = "float64"
+			}
+			asmArgs = append(asmArgs, a+" "+t)
+
+			if t == "float64" {
+				a = "float64(" + a + ")"
+			}
+			goArgsNames = append(goArgsNames, a)
+
+			if isArray {
+				argsJS += fmt.Sprintf("\t\t\t\t[%s(sp+%s)],\n", f, offset)
+			} else {
+				argsJS += fmt.Sprintf("\t\t\t\t%s((sp)+%s),\n", f, offset)
+			}
+			offset += "+" + o
+		}
+
+		resultJS := ""
+		resultGo := ""
+		resultGoType := ""
+		if len(v) > 3 {
+			result := strings.TrimSpace(string(v[3]))
+			if len(result) > 0 {
+				f, _ := translateResultType(result)
+
+				resultGoType = result
+				resultJS = fmt.Sprintf("%s((go._inst.exports.getsp() >>> 0)+%s, r)", f, offset)
+				resultGo = "return "
+			}
+			if strings.Contains(string(v[1]), "Delete") {
+				resultJS = fmt.Sprintf("gioDeleteObject(sp)")
+			}
+		}
+
+		argsJS = strings.Trim(argsJS, ",")
+
+		call := strings.Split(string(findCalls[i][1]), `, `)
+
+		fmt.Fprintf(js, `
+         // %s
+		 "gioui.org/internal/gl.%s": (sp) => {
+            sp = (sp >>> 0);
+
+            let r = webgl.%s(
+%s			);
+
+            %s
+        },`, strings.TrimSpace(strings.TrimRight(string(v[0]), `{`)), "asm"+string(v[1]), strings.Replace(strings.Replace(call[0], `"`, "", -1), `)`, ``, -1), argsJS, resultJS)
+
+		fmt.Fprintf(golang, `
+//go:noescape
+func %s(%s) %s
+
+func %s
+	%s%s(%s)
+}
+`, "asm"+string(v[1]), strings.Join(asmArgs, ", "), resultGoType, v[0], resultGo, "asm"+string(v[1]), strings.Join(goArgsNames, ", "))
+
+		fmt.Fprintf(asm, `TEXT ·%s(SB), NOSPLIT, $0
+  CallImport
+  RET
+
+`, "asm"+string(v[1]))
+	}
+	writeEnd(js)
+}
+
+func writeHeader(f io.StringWriter) {
+	f.WriteString(`(() => {
+
+    // webgl is the array which handles the WebGL context. See InitGL function.
+    let webgl = 0;
+
+    // textDecoder holds the TextDecoder used for encode string.
+    let textDecoder = new TextDecoder("utf-8");
+
+    // invalidateBuffer is re-use when you call invalidateBuffer().
+    let invalidateBuffer = new Int32Array(1);
+
+    // hold values from JS
+    let values = [undefined]
+    let valuesPool = []
+
+    // Offset* is the byte-size of each type (matches with Reflect.Sizeof()).
+    const OffsetContextIndex = 8;
+    const OffsetInt64 = 8;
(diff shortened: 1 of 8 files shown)


Inkeliz:
I forgot to include the `gl_js.s` file, the assembly. 🤣
I hope that it will pass the tests now. 🤞

Inkeliz:
The `syscall/js` is [known to be slow](https://github.com/golang/go/issues/32591), I don't think it will improve soon, so that change is to "bypass" the `syscall/js` at all.

---------

No one asked, but I will answer...

**1. Why not`[]interface`?**
It is expensive. The GC triggers very often, and the `runtime.convTslice` (and such) is painfully slow.
Using `[]interface`, the time per frame drops to ~50ms (from ~80ms). Using `[]uintptr` the time per frame drops to ~40ms. 
In general, switching to `[]uintptr` saves around ~10ms on smartphone and ~0.7ms on a desktop.

**2. Why create `newProc`?**
That avoids the function name to be converted everytime. Yes, it is expensive. The original code (using `syscall/js`) needs to convert the golang-string to js-string EVERY CALL. Supposing that you are calling `window.console.log` you will use: `console.Call(“log”, …)`. That “log” must be converted to JS-string, each time that you call that. It is not cheap as you are supposing, the `LoadString` function is one of the slowest on JS side.

The `newProc` will re-use the same string. Also, since we are not using `interface`, we need to know the type of input. Usually, the `interface` provide us the type, but `interface` is slow.

Furthermore, the `ref` value varies their meaning accordingly with the `output` provided, as commented in the code. The `ref` could be either: the length of string, the reference of the js-object or the integer-value. 

_Using `Inkove` instead of  `Call` (on `syscal/js`) may improve that, but will be slower than this change anyway)._

**3. Why drop `js.Value`?**
Mixing `js.Value` and custom JS is dangerous. The older patch does that quite well, but at first it works on Go 1.15 and not on Go 1.14. So, trying to make it compatible with `js.Value` is not worth. 

**4. Where is the GC? Why not use ` runtime.SetFinalizer`?**
Seems that we did not need. We have the `delete*` function. Such as `DeleteBuffer`. So, it also calls the `free`, which flag that index as available for re-use. So, instead of increasing the list of `values`, it will re-use the same slot of the previous “free” value.

Yes, each `Delete*` will call JS twice, but relying on `runtime.SetFinalizer` will also call twice, and also lead to more inconsistent frame-time.

**5. Why have a shared string buffer?**
String is expensive. The original `syscall/js` makes almost 3 calls to create the string. I did it with a single call. (: 
The clever idea is to create an already-known buffer. If you use `_glTypeString` it will set that buffer and the `ref` will be the length of the string. So, when you call `Call` it already sets the buffer. The buffer (which is already on the Golagn size) will the read by `String()`. Exactly one single call to JS. 

The string cannot be larger than 4096 bytes, that seems enough for us. It is not threading safe, but WebGL does not seems to be thread-safe either. 

**6. That works?**
Yes. It is working fine on multiple versions of Golang, without any specific change for each version. I built it for Go 1.15 and it works on Go 1.14 fine. 

For browser, it requires the ES6 (2015). 

I test it against multiple devices and multiple conditions (WebGL 1-only, WebGL 2-desabled, WebGL 2) and on multiple devices that I have available, also using the BrowserStack. Far I tested, it works even on Firefox 53 (2017-04-18), and Chrome 63 (2017-12-05) and Safari 12. So far, the browser-support seems on par with the current Gio version.

_I tested with the "Kitchen" and my own app, only._

internal/glimpl/gl_js.go
```diff
@@ -6,345 +6,485 @@ import (
...
+	f.Ctx.Call(_glUniform1f, uintptr(unsafe.Pointer(&dst)), uintptr(unsafe.Pointer(&vv0)))
 }
 func (f *Functions) Uniform1i(dst Uniform, v int) {
-	f.Ctx.Call("uniform1i", js.Value(dst), v)
+	f.Ctx.Call(_glUniform1i, uintptr(unsafe.Pointer(&dst)), uintptr(unsafe.Pointer(&v)))
 }
 func (f *Functions) Uniform2f(dst Uniform, v0, v1 float32) {
-	f.Ctx.Call("uniform2f", js.Value(dst), v0, v1)
+	vv0, vv1 := float64(v0), float64(v1) // Float32 doesn't work
+	f.Ctx.Call(_glUniform2f, uintptr(unsafe.Pointer(&dst)), uintptr(unsafe.Pointer(&vv0)), uintptr(unsafe.Pointer(&vv1)))
```
eliasnaur, line 310: Uniform etc. are uints. Won't all these &value pointer taking expressions lead the values to be heap-allocated and thus subject to garbage collection?

eliasnaur (changes requested):
> **3. Why drop `js.Value`?**
> Mixing `js.Value` and custom JS is dangerous. The older patch does that quite well, but at first it works on Go 1.15 and not on Go 1.14. So, trying to make it compatible with `js.Value` is not worth.
> 

Can you elaborate what makes mixing js.Value and low-level javascript assembly dangerous, or at least more dangerous than your custom syscall approach? As far as I can tell, this argument blocks leaving performance insensitive function using the slower but safer syscall/js.

Inkeliz:
> > **3. Why drop `js.Value`?**
> > Mixing `js.Value` and custom JS is dangerous. The older patch does that quite well, but at first it works on Go 1.15 and not on Go 1.14. So, trying to make it compatible with `js.Value` is not worth.
> 
> Can you elaborate what makes mixing js.Value and low-level javascript assembly dangerous, or at least more dangerous than your custom syscall approach? As far as I can tell, this argument blocks leaving performance insensitive function using the slower but safer syscall/js.

Well, the `Call()` is by slow by default, since it needs to "translate" the golang-string to javascript-string (using `Invoke` instead of `Call` might mitigate that).

Anyway....

Let suppose that `NewFramebuffer` is fast using `syscall/js`, but `BindFramebuffer` is slow:

The `BindFramebuffer` (custom js) will take the `js.Value`. Then, we  need to read that value directly from `go._values`. If the `js.Value` changes (either new fields or change the order of the fields), your custom code will break...

The oposite is even worse, if `NewFramebuffer` is slow and  `BindFramebuffer` is fast: 

The `NewFramebuffer` (custom js) needs to generate a valid `js.Value`. So.... The `js.Value` created by the custom-function must have the same "signature" (the same flags) of the `syscall/js.Value`. That "signature"/"flags" changes from Go 1.14 to Go 1.14.1 to Go 1.15, and may change in Go 1.16 too.

The current patch doesn't have to match any flag with the `syscall/js`. The only need is to have the `go.mem` (on the JS side) exposed, that is only used to read/write the input/output from the function. The `go.mem` (js) is available on Go 1.14. 

----

In general currently we only need `go._mem`, and doesn't need anything from `syscall/js`.

internal/glimpl/gl_js.go
Inkeliz: I think you are right, did you have any idea to avoid it? Using `[]interface` seems to have the same issue, actually even worse.

About the `float64()` it's also done internally by the `syscall/js`, https://github.com/golang/go/blob/master/src/syscall/js/js.go#L162-L221. I'm trying to read the `float32` directly on the JS too.

Inkeliz:
I force-push to remove the `float64()`.

eliasnaur:
> > > **3. Why drop `js.Value`?**
> > > Mixing `js.Value` and custom JS is dangerous. The older patch does that quite well, but at first it works on Go 1.15 and not on Go 1.14. So, trying to make it compatible with `js.Value` is not worth.
> > 
> > 
> > Can you elaborate what makes mixing js.Value and low-level javascript assembly dangerous, or at least more dangerous than your custom syscall approach? As far as I can tell, this argument blocks leaving performance insensitive function using the slower but safer syscall/js.
> 
> Well, the `Call()` is by slow by default, since it needs to "translate" the golang-string to javascript-string (using `Invoke` instead of `Call` might mitigate that).
> 

Indeed, any syscall/js is slow. I hope to avoid it where the slowness is in the per-frame hot path.

> Anyway....
> 
> Let suppose that `NewFramebuffer` is fast using `syscall/js`, but `BindFramebuffer` is slow:
> 
> The `BindFramebuffer` (custom js) will take the `js.Value`. Then, we need to read that value directly from `go._values`. If the `js.Value` changes (either new fields or change the order of the fields), your custom code will break...
> 
> The oposite is even worse, if `NewFramebuffer` is slow and `BindFramebuffer` is fast:
> 
> The `NewFramebuffer` (custom js) needs to generate a valid `js.Value`. So.... The `js.Value` created by the custom-function must have the same "signature" (the same flags) of the `syscall/js.Value`. That "signature"/"flags" changes from Go 1.14 to Go 1.14.1 to Go 1.15, and may change in Go 1.16 too.
> 
> The current patch doesn't have to match any flag with the `syscall/js`. The only need is to have the `go.mem` (on the JS side) exposed, that is only used to read/write the input/output from the function. The `go.mem` (js) is available on Go 1.14.
> 
> In general currently we only need `go._mem`, and doesn't need anything from `syscall/js`.

To make the discussion clearer, I added a sketch of an approach that should be compatible with both syscall/js and whatever faster assembly approach we end up with, by storing both a js.Value and an integer reference for each object:

https://git.sr.ht/~eliasnaur/gio/commit/cf926a7491a63a7cfddb530f4f93bcabf6930700

WDYT?

Inkeliz:
I'm not sure if that is worth it. I don't think it solves any issue:

1. It doesn't remove the need of `go.mem` (which is a [DataView](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/DataView)).

But, additionaly:

2. It will call the slow function of `js.Global().Call("createRef", v).Int()`, and use the slow `syscall/js`.
3. It will need to delete the reference twice.

-----

I use something similar here: https://github.com/gioui/gio/pull/4/commits/0289d071a4e78c02e883e726d5a79d4a8b12d276#diff-bf9a8e77af361576a7335f1f62097cd39db03ab401616e81df00a3a9be5aa46dR168-R170

The `global.GlimpContext` get's the `js.Value` and create another reference. However, it's only called on the `NewContext`, which is infrequently called.

----

I'm not sure if I agree with the idea of "just replace slow functions". There's a lot of calls to WebGL, every single frame. For me, all functions must be as fast as possible. 

Also, let's consider that we have six slow functions, such as BufferData, CheckBufferframe, ReadPixel, EnableVertexAttribArray, VertexAttribPointer and Viewport. In the end, we will have 6 JS functions, 6 assemblies (and some additional overhead due to the another `js.Global().Call`) and worse performance than the current PR... Just replacing the `BufferData` saves ~1.5ms (of ~8ms), but changing everything saves ~4.5ms, so why not? 

I'm not sure why you are so worried about dropping the `syscall/js`, [the `syscall/js` also say:](https://godoc.org/syscall/js)
> This package is EXPERIMENTAL. Its current scope is only to allow tests to run, but not yet to provide a comprehensive API for users. It is exempt from the Go compatibility promise.

The current PR has only five functions. If it has some bug will be easier to spot than the older one, since all WebGL functions are using the same `call` function. It also passes all tests, including tests on real-browsers (on WebGL 1 and WebGL 2). I think it even possible to revert this commit if it breaks due to Golang changes (changes on the `wasm_exec` or the new `//go:wasmexport`). Also, using `syscall/js` is also susceptible to the same problem (Go1.13 to Go1.14 have breaking changes on `syscall/js`).

_This PR is working on the new [Go1.16Beta1](https://tip.golang.org/doc/go1.16), without any change._

eliasnaur:
> I'm not sure if that is worth it. I don't think it solves any issue:
> 
>     1. It doesn't remove the need of `go.mem` (which is a [DataView](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/DataView)).
> 

Sure, avoiding syscall/js means tinkering with internals. The goal is to minimize the tinkering.

> 
> But, additionaly:
> 
>     1. It will call the slow function of `js.Global().Call("createRef", v).Int()`, and use the slow `syscall/js`.
> 

Yes, but only on *creation* (and deletion) of objects, which is already expensive. The GPU backend takes care to re-use GPU objects.

>     2. It will need to delete the reference twice.
> 

Again, only during deletion which is a rare occurrence.

> 
> I use something similar here: [0289d07#diff-bf9a8e77af361576a7335f1f62097cd39db03ab401616e81df00a3a9be5aa46dR168-R170](https://github.com/gioui/gio/commit/0289d071a4e78c02e883e726d5a79d4a8b12d276#diff-bf9a8e77af361576a7335f1f62097cd39db03ab401616e81df00a3a9be5aa46dR168-R170)
> 
> The `global.GlimpContext` get's the `js.Value` and create another reference. However, it's only called on the `NewContext`, which is infrequently called.
> 

Indeed. I'm proposing to expand that facility to all GPU objects, and to keep a struct of both the js.Value and fast integer reference.

> I'm not sure if I agree with the idea of "just replace slow functions". There's a lot of calls to WebGL, every single frame. For me, all functions must be as fast as possible.
> 

It's true that a lot of calls are per-frame and need assembly implementation. However, with the dual js.Value/ref approach we're not bound to all-or-nothing. As a side bonus, I don't think any performance critical functions use string parameters, so we can omit string handling entirely.

> Also, let's consider that we have six slow functions, such as BufferData, CheckBufferframe, ReadPixel, EnableVertexAttribArray, VertexAttribPointer and Viewport. In the end, we will have 6 JS functions, 6 assemblies (and some additional overhead due to the another `js.Global().Call`) and worse performance than the current PR... Just replacing the `BufferData` saves ~1.5ms (of ~8ms), but changing everything saves ~4.5ms, so why not?
> 

I argue that the 4.5ms gain can be achieved with fewer functions replaced.

> I'm not sure why you are so worried about dropping the `syscall/js`, [the `syscall/js` also say:](https://godoc.org/syscall/js)
> 
> > This package is EXPERIMENTAL. Its current scope is only to allow tests to run, but not yet to provide a comprehensive API for users. It is exempt from the Go compatibility promise.
> 

The entire WASM port is experimental :) However, can we agree that syscall/js is safer than any other approach?

> The current PR has only five functions. If it has some bug will be easier to spot than the older one, since all WebGL functions are using the same `call` function. It also passes all tests, including tests on real-browsers (on WebGL 1 and WebGL 2). I think it even possible to revert this commit if it breaks due to Golang changes (changes on the `wasm_exec` or the new `//go:wasmexport`). Also, using `syscall/js` is also susceptible to the same problem (Go1.13 to Go1.14 have breaking changes on `syscall/js`).
> 
> _This PR is working on the new [Go1.16Beta1](https://tip.golang.org/doc/go1.16), without any change._

Ok. I admit that your latest revision does look a fair bit less scary than previous revisions, and I think I can be persuaded to replace all of syscall/js. However, there are two issues I'd like to see fixed:

- The uintptr(unsafe.Pointer()) conversions are both unsafe and creates garbage.
- The JNI-like approach is slower than straight up assembly/JS implementations, yet more complex.

I believe both issues can be solved by `go generate`'ing the assembly implementations. By generating you gain the maximal performance, avoid generic uintptr parameter types, and only need to ensure the generator is correct, not all the hand-written assembly stubs. I assume the generator will be about as complex as your JNI-like approach, except that the overhead is moved to generate time, not runtime.

internal/glimpl/gl_js.go
eliasnaur: See my comment about generating the assembly stubs.

Inkeliz:
> The uintptr(unsafe.Pointer()) conversions are both unsafe and creates garbage.

I tought about that, but the current `syscall/js` seems to create the same garbage or more.

> The JNI-like approach is slower than straight up assembly/JS implementations, yet more complex.

I'll try to build some proper generator.

eliasnaur:
> > The uintptr(unsafe.Pointer()) conversions are both unsafe and creates garbage.
> 
> I thought about that, but the current `syscall/js` seems to create the same garbage or more. Also, using the `func somefunction(arg int)` will not copy the `arg int` too? Just asking, but removing the `uintptr(unsafe.Pointer())` is better anyway.
> 

No, non-pointer parameters don't create garbage. `uintptr` is a value-type and doesn't create garbage either; taking the address (`&value`) is what forces heap allocation and garbage.

> > The JNI-like approach is slower than straight up assembly/JS implementations, yet more complex.
> 
> I'll try to build some proper generator.

eliasnaur:
While restructuring the GL backend for the new renderer, I discovered that the WebGL backend has been passing much more data than intended (the entire cached byte buffer, not just a slice of it). See https://gioui.org/commit/e012c3f393792 for the fix. It should speed up buffer(Sub)Data and friends.

Inkeliz:
I'm parking the code here. The generator is very poorly written, I tried to use the `go/package` and `go/ast` but it seem every more difficult, so I basically copy my old "generator" (which is bunch of regex).

--------

There's some changes, that I think not belong to here. I add some cache to `GetInteger`, `GetFloat` (...), since most of the values are constant and it's called every frame. It saves ~1.5ms, even using `syscall/js`. Maybe the gpu package could cache that values, instead of requesting them every frame.

To use the asm version, must use `-tags unsafe`, so both methods exists.

---------

I don't think it will be accepted in Gio core, but I'm using that patch. :\

PR #5, opened 2020-12-17
app/internal/window: fix support for Safari 11.1

Inkeliz:


eliasnaur:
Thanks! Merged.

Inkeliz:
It get's merged, but doesn't closes the PR. 👀

eliasnaur:
Can't explain why :/

whereswaldon:
@eliasnaur The auto-close feature will only work if you don't change the commit hash as part of merging the commit. The commit in this PR has hash `8123`, but the version in `main`'s history has a different hash. GitHub can't tell they're the same, so I think this is expected behavior.

PR #6, opened 2020-12-18
app/internal/window: [wasm] add support for insets and cache window-size

Inkeliz:
Currently, on iOS, the keyboard can block the content since it doesn't
trigger the "resize". Now, the `insets` will provide the size of the
keyboard (on iOS) and scrollbars. It's compatible with the Android (and
desktop), but Chrome automatic resizes the windows, so the `Inset` will
be 0.

The `Visual Viewport API` is also triggered when zooming, so the scale 
will change.

### DIFF
--- a/app/internal/window/os_js.go
+++ b/app/internal/window/os_js.go
@@ -30,12 +30,15 @@ type window struct {
 	clipboardCallback     js.Func
 	requestAnimationFrame js.Value
 	browserHistory        js.Value
+	visualViewport        js.Value
 	cleanfuncs            []func()
 	touches               []js.Value
 	composing             bool
 	requestFocus          bool
 
 	mu        sync.Mutex
+	size      f32.Point
+	viewport  f32.Point
 	scale     float32
 	animating bool
 }
@@ -56,6 +59,10 @@ func NewWindow(win Callbacks, opts *Options) error {
 	}
 	w.requestAnimationFrame = w.window.Get("requestAnimationFrame")
 	w.browserHistory = w.window.Get("history")
+	w.visualViewport = w.window.Get("visualViewport")
+	if w.visualViewport.IsUndefined() {
+		w.visualViewport = w.window
+	}
 	w.redraw = w.funcOf(func(this js.Value, args []js.Value) interface{} {
 		w.animCallback()
 		return nil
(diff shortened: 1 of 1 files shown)


app/internal/window/os_js.go
```diff
@@ -123,10 +129,19 @@ func (w *window) cleanup() {
...
+			w.draw(true)
+			return nil
+		})
+	} else {
+		w.addEventListener(w.visualViewport, "resize", func(this js.Value, args []js.Value) interface{} {
+			w.resize()
+			w.draw(true)
+			return nil
+		})
+	}
```
eliasnaur, line 144: Deduplicate the if and else branches by introducing a local:

```
viewport := w.visualViewport
if viewport.IsUndefined() {
    viewport = w.window
}
```

app/internal/window/os_js.go
```diff
@@ -466,43 +481,61 @@ func (w *window) ShowTextInput(show bool) {
...
-	if cfg == (unit.Metric{}) || width == 0 || height == 0 {
+	width, height, insets, metric := w.config()
+	if metric == (unit.Metric{}) || width == 0 || height == 0 {
 		return
 	}
-	w.mu.Lock()
-	w.scale = float32(scale)
-	w.mu.Unlock()
+
+	if w.resized {
```
eliasnaur, line 508: Why delay resizing the canvas, and not do it immediately in w.resize?

eliasnaur (changes requested):
Nice work.

app/internal/window/os_js.go
```diff
@@ -466,43 +475,59 @@ func (w *window) ShowTextInput(show bool) {
...
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.scale = float32(w.window.Get("devicePixelRatio").Float())
+
+	rect := w.cnv.Call("getBoundingClientRect")
+	w.size.X = float32(rect.Get("width").Float()) * w.scale
+	w.size.Y = float32(rect.Get("height").Float()) * w.scale
+
+	if vx, vy := w.visualViewport.Get("width"), w.visualViewport.Get("height"); !vx.IsUndefined() && !vy.IsUndefined() {
```
eliasnaur, line 488: Now that w.visualViewport is always set, is the IsUndefined tests still necessary?

Inkeliz: Yes. The `w.visualViewport` can be either "Window" or "Visual Viewport". The `w.window` doesn't have `window.width`/`window.height`.

eliasnaur:
Thanks! Merged.

PR #7, opened 2021-01-03
app/internal/window: [wasm] fix animation framerate

Inkeliz:
Before this change, Gio could spawn multiple `w.requestAnimationFrame.Invoke`,
which generates multiples `w.draw`, faster than the monitor refresh-rate.

This patch aims to call `w.requestAnimationFrame.Invoke` everyframe, but it
will only draw (`w.draw`) when animating.

### DIFF
--- a/app/internal/window/os_js.go
+++ b/app/internal/window/os_js.go
@@ -82,6 +82,7 @@ func NewWindow(win Callbacks, opts *Options) error {
 		w.w.Event(system.StageEvent{Stage: system.StageRunning})
 		w.resize()
 		w.draw(true)
+		w.animCallback()
 		select {}
 	}()
 	return nil
@@ -417,22 +418,17 @@ func (w *window) funcOf(f func(this js.Value, args []js.Value) interface{}) js.F
 func (w *window) animCallback() {
 	w.mu.Lock()
 	anim := w.animating
-	if anim {
-		w.requestAnimationFrame.Invoke(w.redraw)
-	}
 	w.mu.Unlock()
 	if anim {
 		w.draw(false)
 	}
+	w.requestAnimationFrame.Invoke(w.redraw)
 }
 
 func (w *window) SetAnimating(anim bool) {
 	w.mu.Lock()
-	defer w.mu.Unlock()
-	if anim && !w.animating {
-		w.requestAnimationFrame.Invoke(w.redraw)
-	}
 	w.animating = anim
+	w.mu.Unlock()
 }
 
 func (w *window) ReadClipboard() {


app/internal/window/os_js.go
```diff
@@ -417,22 +418,17 @@ func (w *window) funcOf(f func(this js.Value, args []js.Value) interface{}) js.F
...
 }
 
 func (w *window) SetAnimating(anim bool) {
 	w.mu.Lock()
-	defer w.mu.Unlock()
-	if anim && !w.animating {
-		w.requestAnimationFrame.Invoke(w.redraw)
-	}
 	w.animating = anim
+	w.mu.Unlock()
```
eliasnaur, line 431: Defer the Unlock.

app/internal/window/os_js.go
```diff
@@ -417,22 +418,17 @@ func (w *window) funcOf(f func(this js.Value, args []js.Value) interface{}) js.F
...
 		w.draw(false)
 	}
+	w.requestAnimationFrame.Invoke(w.redraw)
 }
 
 func (w *window) SetAnimating(anim bool) {
 	w.mu.Lock()
-	defer w.mu.Unlock()
-	if anim && !w.animating {
-		w.requestAnimationFrame.Invoke(w.redraw)
```
eliasnaur, line 433: Without requestAnimationFrame here, how will animation start? With your change, SetAnimating only sets w.animating, and that's it.

app/internal/window/os_js.go
```diff
@@ -417,22 +418,17 @@ func (w *window) funcOf(f func(this js.Value, args []js.Value) interface{}) js.F
...
 	w.mu.Lock()
 	anim := w.animating
-	if anim {
-		w.requestAnimationFrame.Invoke(w.redraw)
-	}
 	w.mu.Unlock()
 	if anim {
 		w.draw(false)
 	}
+	w.requestAnimationFrame.Invoke(w.redraw)
```
eliasnaur, line 425: This is called unconditionally now. When will animation stop?

eliasnaur (changes requested):

app/internal/window/os_js.go
Inkeliz: It starts inside ` NewWindow`, after the `w.draw(true)`, https://github.com/gioui/gio/pull/7/commits/41d91013e4a6e68c596b630a7c01873427e8e6bf#diff-abf70745e7c00ff72bc8babcc591453b6cde31a320dbb2bc50c13872115855faR85.

app/internal/window/os_js.go
Inkeliz: It starts inside NewWindow, after the w.draw(true), 41d9101#diff-abf70745e7c00ff72bc8babcc591453b6cde31a320dbb2bc50c13872115855faR85.

app/internal/window/os_js.go
Inkeliz: The `if anim {}` is responsable to prevent the redraw. The "animation callback" will never stop. The callback will be triggred each ~16ms (the refresh-rate of the monitor). However, it will not draw a new frame if it's not animating. 

The old way of calling `requestAnimationFrame` only when animate, triggers the callback faster than the refresh-rate.  I think it happens because `SetAnimation` can be true and false in one frame. Or, it can change one frame to another, which might call `requestAnimationFrame` inside the SetAnimation AND calls `requestAnimationFrame` in the current callback, so there's two calls in the same frame (or more).

eliasnaur: I don't think it's a good idea to always have requestAnimationFrame running, even if it is nearly a no-op. There must be some other way of fixing the piled up requests. How about tracking whether there is a pending requestAnimationFrame, and not submit another if so?

Inkeliz: First, I did some:
 ```
 t := time.Now()
	for e := range w.Events() {
		switch e := e.(type) {
		case system.FrameEvent:
			e.Frame(gtx.Ops)
			fmt.Println("FRAME", t.Sub(time.Now()))
                        t = time.Now()
			w.Invalidate()
```
The `fmt.Println` prints the time betwwen each frame. On Windows/Android it's 16ms. On WASM it's 6ms (or 2ms with #4 applied). Anyway, the WASM is creating frame faster than intended! 

I found that the `SetAnimating` calls the `w.requestAnimationFrame` and also it's called inside the callback, so I guess... It's calling twice. I insert some `fmt.Println`, the `w.SetAnimating` sometimes is `false` and them `true` (I don't know if it's one the same frame, or not), but... Since it's false > true, it could call `w.requestAnimationFrame` again, and again.

_If you test the same code (above) with that PR, the timing will be ~16ms._

Comment (eliasnaur) [reply app/internal/window/os_js.go:425]: I'm sure the PR fixes what it sets out to do (16 ms) frame, but I'm saying it's not worth the cost: always having requestAnimationFrame callbacks, even if the program is otherwise idle.

I suggested tracking whether a requestAnimationFrame has been called, and only submitting a new one if not. I believe that will achieve the same effect: reducing the updates to once every ~16ms, but without callbacks while idle (not animating).

**Review by eliasnaur — CHANGES_REQUESTED** [review]

File: app/internal/window/os_js.go (hunk 3 above)
Comment (Inkeliz) [reply app/internal/window/os_js.go:425]: I don't know to better fix that. In theory the old code was suppose to prevent multiples `w.requestAnimationFrame.Invoke`:
```
if anim && !w.animating {	
		w.requestAnimationFrame.Invoke(w.redraw)
``` 

But, doesn't seems to prevent such duplication. The another option might be adding some loop/select, but I think it's worse than call  `w.requestAnimationFrame.Invoke` everytime.

**Comment by eliasnaur** [comment]
I can't reproduce the < 16 ms frame times (on macOS Firefox), but I suspect what's fixing your case is the move of requestAnimationFrame after draw. Does

```
diff --git a/app/internal/window/os_js.go b/app/internal/window/os_js.go
index 6a35ae9..dc1f07b 100644
--- a/app/internal/window/os_js.go
+++ b/app/internal/window/os_js.go
@@ -417,12 +417,10 @@ func (w *window) funcOf(f func(this js.Value, args []js.Value) interface{}) js.F
 func (w *window) animCallback() {
        w.mu.Lock()
        anim := w.animating
-       if anim {
-               w.requestAnimationFrame.Invoke(w.redraw)
-       }
        w.mu.Unlock()
        if anim {
                w.draw(false)
+               w.requestAnimationFrame.Invoke(w.redraw)
        }
 }
```

fix your < 16 ms frame times?

**Comment by Inkeliz** [comment]
The suggestion of moving `w.requestAnimationFrame.Invoke(w.redraw)` doesn't fix. I tried it again (Chrome and Opera), and it doesn't work. I'm getting frames < 16ms.

------

I insert one `fmt.Println()` inside the `SetAnimation`, that is the result:

```
wasm.js:57 FRAME -6.610176ms
wasm.js:57 2021-01-06 13:24:08.449 +0000 UTC+0 m=+11.827149825 true
wasm.js:57 2021-01-06 13:24:08.456 +0000 UTC+0 m=+11.833569793 false
wasm.js:57 FRAME -6.834944ms
wasm.js:57 2021-01-06 13:24:08.457 +0000 UTC+0 m=+11.834279681 true
wasm.js:57 2021-01-06 13:24:08.463 +0000 UTC+0 m=+11.840424705 false
wasm.js:57 FRAME -6.56512ms
wasm.js:57 2021-01-06 13:24:08.463 +0000 UTC+0 m=+11.841059841 true
wasm.js:57 2021-01-06 13:24:08.47 +0000 UTC+0 m=+11.847494913 false
```

I think it's switching between `true` and `false` and causing to invoke again and again. _I test only on Chrome/Opera._

**Comment by eliasnaur** [comment]
> The suggestion of moving `w.requestAnimationFrame.Invoke(w.redraw)` doesn't fix. I tried it again (Chrome and Opera), and it doesn't work. I'm getting frames < 16ms.
> 
> I insert one `fmt.Println()` inside the `SetAnimation`, that is the result:
> 
> ```
> wasm.js:57 FRAME -6.610176ms
> wasm.js:57 2021-01-06 13:24:08.449 +0000 UTC+0 m=+11.827149825 true
> wasm.js:57 2021-01-06 13:24:08.456 +0000 UTC+0 m=+11.833569793 false
> wasm.js:57 FRAME -6.834944ms
> wasm.js:57 2021-01-06 13:24:08.457 +0000 UTC+0 m=+11.834279681 true
> wasm.js:57 2021-01-06 13:24:08.463 +0000 UTC+0 m=+11.840424705 false
> wasm.js:57 FRAME -6.56512ms
> wasm.js:57 2021-01-06 13:24:08.463 +0000 UTC+0 m=+11.841059841 true
> wasm.js:57 2021-01-06 13:24:08.47 +0000 UTC+0 m=+11.847494913 false
> ```
> 
> I think it's switching between `true` and `false` and causing to invoke again and again. _I test only on Chrome/Opera._

Indeed, SetAnimation may be called multiple times per frame with some efficiency cost. However, the important problem is whether _drawing_ happens more than once per monitor refresh. Even on my somewhat fast Linux machine running either Firefox or Chromium I get frame times ~16ms. I'm using the hello example to minimize drawing overhead, patched with

```
diff --git hello/hello.go hello/hello.go
index 7a328e7..75896d9 100644
--- hello/hello.go
+++ hello/hello.go
@@ -8,6 +8,7 @@ import (
        "image/color"
        "log"
        "os"
+       "time"
 
        "gioui.org/app"
        "gioui.org/io/system"
@@ -33,18 +34,23 @@ func main() {
 func loop(w *app.Window) error {
        th := material.NewTheme(gofont.Collection())
        var ops op.Ops
+       last := time.Now()
        for {
                e := <-w.Events()
                switch e := e.(type) {
                case system.DestroyEvent:
                        return e.Err
                case system.FrameEvent:
+                       now := time.Now()
+                       log.Println("elapsed", now.Sub(last))
+                       last = now
                        l := material.H1(th, "Hello, Gio")
                        maroon := color.NRGBA{R: 127, G: 0, B: 0, A: 255}
                        l.Color = maroon
                        l.Alignment = text.Middle
                        l.Layout(gtx)
+                       op.InvalidateOp{}.Add(gtx.Ops)
                        e.Frame(gtx.Ops)
                }
        }
```

Output:
```
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.920064ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.909824ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.499968ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.64512ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.934912ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.125184ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 17.32992ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.32512ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 17.11488ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.395008ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.014848ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 17.110272ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 17.149952ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 17.529856ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 15.195136ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 18.204928ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.300032ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 15.219968ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.90496ms
wasm_exec.js:51 2021/01/07 11:32:47 elapsed 16.630016ms
```

Writing this, I note that you're using Window.Invalidate to trigger a re-draw. Invalidate is only meant for externally triggered events, such as receiving a response to a network request, and not for animation. Use InvalidateOp for animation. If you'd like to make Window.Invalidate more robust against misuse, I think one way is to cancel any outstanding animation frame callback with [cancelAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame). Or better yet, understand why Window.Invalidate can avoid calling SetAnimating multiple times.

Inkeliz:
Using `op.InvalidateOp{}.Add(gtx.Ops)` seems to "fix". But, not so much.... If I leave the page (switch the tabs) and go back, it will be ~8ms. So, it starts as ~16ms... Switching the page/tab it drops to ~8ms.

----------

https://user-images.githubusercontent.com/4644332/103903531-eab5f380-50f3-11eb-8d47-594d8c573f5e.mp4

eliasnaur:
I still can't reproduce the issue, but it sounds like multiple requestAnimationCallbacks are triggered for each frame. Does https://gioui.org/commit/5f6fa25 help?

Inkeliz:
Yes, it fixes the issue.

theclapp:
Elias:
> I note that you're using Window.Invalidate to trigger a re-draw. Invalidate is only meant for externally triggered events, such as receiving a response to a network request, and not for animation. Use InvalidateOp for animation

Can you elaborate on why this is?  I might be using it wrong.

eliasnaur:
@theclapp see https://gioui.org/commit/759b796 where I clarify the difference between Invalidate/InvalidateOp, and add an opportunistic check for Invalidate to make it almost as good as an InvalidateOp.

theclapp:
👍 Thanks, that helps.

PR #8, opened 2021-01-04
cmd/gogio: remove windres dependency

Inkeliz:
Now, it's possible to compile to Windows (`-target windows`) without
having `windres`. The PNG icon, manifest and version info will be
generated and include using `gogio`.

### DIFF
--- a/cmd/gogio/windowsbuild.go
+++ b/cmd/gogio/windowsbuild.go
@@ -4,20 +4,23 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"image"
+	"github.com/akavel/rsrc/binutil"
+	"github.com/akavel/rsrc/coff"
+	"golang.org/x/text/encoding/unicode"
 	"image/png"
 	"io"
 	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"text/template"
 )
 
 func buildWindows(tmpDir string, bi *buildInfo) error {
-	builder := &windowsBuilder{TempDir: tmpDir, BuildInfo: bi}
+	builder := &windowsBuilder{TempDir: tmpDir}
 	builder.DestDir = *destPath
 	if builder.DestDir == "" {
 		builder.DestDir = bi.pkgPath
@@ -40,29 +43,36 @@ func buildWindows(tmpDir string, bi *buildInfo) error {
 		return fmt.Errorf("version (%d) is larger than the maximum (%d)", bi.version, math.MaxUint16)
 	}
 
-	builder.Resources.Name = name
-	builder.Manifest.Name = name
-	builder.Manifest.WindowsVersion = sdk
-	builder.Resources.Version = "1,0,0," + version
-	builder.Manifest.Version = "1.0.0." + version
+	for _, arch := range bi.archs {
+		builder.Coff = coff.NewRSRC()
+		builder.Coff.Arch(arch)
 
-	if err := builder.createIcon(); err != nil {
-		return err
-	}
+		if err := builder.embedIcon(bi.iconPath); err != nil {
+			return err
+		}
 
-	if err := builder.createManifest(); err != nil {
-		return fmt.Errorf("can't create manifest: %v", err)
-	}
-	if err := builder.createResource(); err != nil {
-		return fmt.Errorf("can't create resource: %v", err)
-	}
+		if err := builder.embedManifest(windowsManifest{
+			Version:        "1.0.0." + version,
+			WindowsVersion: sdk,
+			Name:           name,
+		}); err != nil {
+			return fmt.Errorf("can't create manifest: %v", err)
+		}
 
-	if err := builder.buildResource(); err != nil {
-		return fmt.Errorf("can't build the resources: %v", err)
-	}
+		if err := builder.embedInfo(windowsResources{
+			Version:      [2]uint32{uint32(1) << 16, uint32(bi.version)},
+			VersionHuman: "1.0.0." + version,
+			Name:         name,
+			Language:     0x0400, // Process Default Language: https://docs.microsoft.com/en-us/previous-versions/ms957130(v=msdn.10)
+		}); err != nil {
+			return fmt.Errorf("can't create info: %v", err)
+		}
 
-	for _, arch := range builder.BuildInfo.archs {
-		if err := builder.buildProgram(arch); err != nil {
+		if err := builder.buildResource(bi, name, arch); err != nil {
+			return fmt.Errorf("can't build the resources: %v", err)
+		}
+
+		if err := builder.buildProgram(bi, name, arch); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +237,156 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+type windowsInfoValue struct {
+	Length      uint16
+	ValueLength uint16
+	Type        uint16
+	Key         []byte
+	Value       []byte
+}
+
+func (v windowsInfoValue) WriteTo(w io.Writer) (_ int64, err error) {
+	// binary.Write doesn't support []byte inside struct. 0.o
```
eliasnaur, line 329: Drop emoji :)

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +237,156 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+	}
+
+	v.Length += uint16(len(v.Value))
+
+	return v
+}
+
+func stringBytes(s string, previous int) []byte {
+	r := []rune(s)
+	l := (len(r) * 2) + 2 // +2 due to null-termination
```
eliasnaur, line 383: The rest of this function looks like a poor man's https://pkg.go.dev/unicode/utf16#Encode or https://godoc.org/golang.org/x/text/encoding/unicode#UTF16 for []byte output. In particular, the unicode packages handle utf16 surrogate pairs.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +237,156 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+			FileFlags:      0,
+			FileOS:         0x40004,
+			FileType:       0x1,
+			FileSubType:    0,
+			FileDate:       [2]uint32{},
+		},
+		// https://docs.microsoft.com/pt-br/windows/win32/menurc/stringfileinfo
+		newValue(valueText, "StringFileInfo", []io.WriterTo{
+			// https://docs.microsoft.com/pt-br/windows/win32/menurc/stringtable
+			newValue(valueText, "04000400", []io.WriterTo{
```
eliasnaur, line 274: Perhaps name the magic value 04000400 with a nearby const.

cmd/gogio/windowsbuild.go
```diff
@@ -72,157 +77,125 @@ func buildWindows(tmpDir string, bi *buildInfo) error {
...
-}
-
-func (b *windowsBuilder) buildResource() error {
-	cmd := exec.Command(
-		"windres",
-		b.ResourcesPath,
-		filepath.Join(b.BuildInfo.pkgPath, "main_windows.syso"),
-	)
-	_, err := runCmd(cmd)
-	return err
```
eliasnaur, line 210: Yay!

eliasnaur (changes requested):
Thanks for removing the windres dependency.

cmd/gogio/windowsbuild.go
Inkeliz: I change it to use `unicode` and add some comments about the padding, making less magic inside `make([]byte)`.

cmd/gogio/windowsbuild.go
Inkeliz: It's not magic, it's Language (idiom) and the Page (index). It's possible to have multiples "pages" for each language. I added one `v.Language` and `page` to make it easier to understand and easier when it supports multiples languages. 

_Currently, `gogio` doesn't know the idiom, so I choose `Process Default Language` (which is the 0x0400)._

Inkeliz:
I did another force-push to fix the `do.mod`.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+		// https://docs.microsoft.com/pt-br/windows/win32/menurc/stringfileinfo
+		newValue(valueText, "StringFileInfo", []io.WriterTo{
+			// https://docs.microsoft.com/pt-br/windows/win32/menurc/stringtable
+			newValue(valueText, fmt.Sprintf("%04X%04X", v.Language, page), []io.WriterTo{
+				// https://docs.microsoft.com/pt-br/windows/win32/menurc/string-str
+				newValue(valueText, "ProductVersion", v.VersionHuman),
+				newValue(valueText, "FileVersion", v.VersionHuman),
+				newValue(valueText, "FileDescription", v.Name),
+				newValue(valueText, "ProductName", v.Name),
+				// @TODO include more data: gogio must have some way to provide such information (like Company Name, Copyright...)
```
eliasnaur, line 284: Delete "@".

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+	}
+
+	v.Length += uint16(len(v.Value))
+
+	return v
+}
+
+var utf16encoder = unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewEncoder()
+
+func stringBytes(s string, previous int) []byte {
```
eliasnaur, line 387: Rename stringBytes to utf16Encode, and comment that the encoded string is null-terminated.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+	b, err := utf16encoder.Bytes([]byte(s))
+	if err != nil {
+		panic(err)
+	}
+	b = append(b, 0x00, 0x00) // null-termination.
+	if (len(b)+previous)%4 != 0 {
```
eliasnaur, line 393: Padding doesn't seem right here, with the adding of previous and whatnot. Move padding to the one caller that need it.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+		b = append(b, 0x00, 0x00) // It must be 32-bit.
```
eliasnaur, line 394: What if len(b) + previous equals 5? All the more reason to move padding to the caller.

eliasnaur (changes requested):

cmd/gogio/windowsbuild.go
Inkeliz: I think it's impossible, contents are either uint16 or uint32. _All the `[]byte` are also padded._

cmd/gogio/windowsbuild.go
Inkeliz: I'll move that.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+)
+
+func newValue(valueType uint16, key string, input interface{}) windowsInfoValue {
+	v := windowsInfoValue{
+		Type: valueType,
+		Key:  utf16Encode(key),
+	}
+
+	v.Length = 6 + uint16(len(v.Key))
+	v.Key = append(v.Key, make([]byte, v.Length%4)...) // Padding, it must be 32-bit.
```
eliasnaur, line 358: The pading calculation is only correct if v.Length%2 == 0. The correct calculation is (4 - length%4).

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+	v.Length += v.Length % 4
```
eliasnaur, line 359: Ditto. Perhaps make a padding function, possibly local to newValue.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+
+	switch in := input.(type) {
+	case string:
+		v.Value = utf16Encode(in)
+		v.Value = append(v.Value, make([]byte, len(v.Value)%4)...) // Padding, it must be 32-bit.
```
eliasnaur, line 364: Ditto.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+		v.ValueLength = uint16(buff.Len())
+		v.Value = buff.Bytes()
+	}
+
+	v.Length += uint16(len(v.Value))
+
+	return v
+}
+
+var utf16encoder = unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewEncoder()
```
eliasnaur, line 388: Consider moving this variable into utf16Encode. It's neater and keeps utf16Encode race-free in case we split work across goroutines in future.

cmd/gogio/windowsbuild.go
```diff
@@ -264,53 +240,158 @@ func (f *windowsManifest) encode(w io.Writer) error {
...
+	// For some reason the ValueLength of the VS_VERSIONINFO must be the byte-length of the `windowsInfoValueFixed`:
+	t.ValueLength = 52
+
+	var verrsrc bufferCoff
+	if _, err := t.WriteTo(&verrsrc); err != nil {
 		return err
 	}
 
-	return template.Execute(w, f)
+	b.Coff.AddResource(16, 1, &verrsrc)
```
eliasnaur, line 302: What is the "16" number?

cmd/gogio/windowsbuild.go
```diff
@@ -72,157 +79,126 @@ func buildWindows(tmpDir string, bi *buildInfo) error {
...
-		}
-		// ICONDIRENTRY 4-6 structure
-		if err := binary.Write(w, binary.LittleEndian, [2]uint16{1, 32}); err != nil {
-			return err
-		}
-		// ICONDIRENTRY 8-12 structure
-		if err := binary.Write(w, binary.LittleEndian, [2]uint32{uint32(imageBuffer.Len() - imageOffset), uint32(headerOffset + imageOffset)}); err != nil {
+		b.Coff.AddResource(coff.RT_ICON, uint16(size), &iconBuffer)
+
+		if err = binary.Write(&iconHeader, binary.LittleEndian, struct {
```
eliasnaur, line 136: Use ":=" instead of "=".

cmd/gogio/windowsbuild.go
```diff
@@ -72,157 +79,126 @@ func buildWindows(tmpDir string, bi *buildInfo) error {
...
-		headerOffset = OffsetICONDIR + (OffsetICONDIRENTRY * len(sizes))
-		imageBuffer  bytes.Buffer
-	)
 	for _, size := range sizes {
-		imageOffset := imageBuffer.Len()
-		scaledImage := resizeIcon(iconVariant{size: size, fill: false}, img)
+		var iconBuffer bufferCoff
 
-		if err := png.Encode(&imageBuffer, scaledImage); err != nil {
+		if err = png.Encode(&iconBuffer, resizeIcon(iconVariant{size: size, fill: false}, iconImage)); err != nil {
```
eliasnaur, line 130: Use ":=" instead of "=".

cmd/gogio/windowsbuild.go
```diff
@@ -72,157 +79,126 @@ func buildWindows(tmpDir string, bi *buildInfo) error {
...
-	)
-
-	sizes := []int{16, 32, 48, 64, 128, 256}
+	var sizes = []int{16, 32, 48, 64, 128, 256}
+	var iconHeader bufferCoff
 
-	// ICONDIR structure
-	if err := binary.Write(w, binary.LittleEndian, [3]uint16{0, 1, uint16(len(sizes))}); err != nil {
+	// GRPICONDIR structure.
+	if err = binary.Write(&iconHeader, binary.LittleEndian, [3]uint16{0, 1, uint16(len(sizes))}); err != nil {
```
eliasnaur, line 123: Use ":=" instead of "=".

cmd/gogio/windowsbuild.go
```diff
@@ -72,157 +79,126 @@ func buildWindows(tmpDir string, bi *buildInfo) error {
...
-
-func convertPNGtoICO(w io.Writer, img image.Image) error {
-	// The file must be in .ICO format.
-	const (
-		OffsetICONDIR      int = 2 * 3
-		OffsetICONDIRENTRY int = (4 * 1) + (2 * 2) + (4 * 2)
-	)
-
-	sizes := []int{16, 32, 48, 64, 128, 256}
+	var sizes = []int{16, 32, 48, 64, 128, 256}
```
eliasnaur, line 119: Retain the shorter form:

sizes := []int{...

eliasnaur (changes requested):

eliasnaur:
Thank you. All merged.

PR #9, opened 2021-01-06
app/internal/window: [wasm] fix insets on Firefox

Inkeliz:
Insets must not be included if the `visualViewport` isn't available (Firefox).

### DIFF
--- a/app/internal/window/os_js.go
+++ b/app/internal/window/os_js.go
@@ -38,7 +38,7 @@ type window struct {
 
 	mu        sync.Mutex
 	size      f32.Point
-	viewport  f32.Point
+	inset     f32.Point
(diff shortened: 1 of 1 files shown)


Inkeliz:
Previously, the inset was `unit.Px(w.size.Y - w.viewport.Y)`. However, `w.viewport.Y` will be `0` if VisualViewport API isn't supported by the browser. In the end, the `inset` is equal to the size of the `w.size.Y`. Now, the `w.inset` will only set when `VisualViewport` is available. 

---

It's now working as intended on Safari, Chrome, Opera, Edge and Firefox.

eliasnaur:
Thanks. All merged.

PR #10, opened 2021-01-09
app, app/internal: [wasm,android] add navigation/status color changer

Inkeliz:
This changes adds the `window.SetColor` to programmatically change the color
of the Status-Bar and Navigation-Bar on Android and WASM (supported by Chrome
on Android).

Note: Windows, Linux, MacOS, iOS, FreeBSD, OpenBSD remains unchanged and
will be ignored.

### DIFF
--- a/app/internal/wm/GioView.java
+++ b/app/internal/wm/GioView.java
@@ -29,6 +29,8 @@ import android.view.WindowInsets;
 import android.view.Surface;
 import android.view.SurfaceView;
 import android.view.SurfaceHolder;
+import android.view.Window;
+import android.view.WindowInsetsController;
 import android.view.inputmethod.BaseInputConnection;
 import android.view.inputmethod.InputConnection;
 import android.view.inputmethod.InputMethodManager;
@@ -131,6 +133,70 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 		PointerIcon pointerIcon = PointerIcon.getSystemIcon(getContext(), id);
 		setPointerIcon(pointerIcon);
 	}
+
+    private enum Bar {
+        NAVIGATION,
+        STATUS,
+    }
+
+    private void setBarColor(Bar t, int color, int luminance) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+            return;
+        }
+
+        Window window = ((Activity) this.getContext()).getWindow();
+
+        int insetsMask;
+        int viewMask;
+
+        switch (t) {
+        case STATUS:
+            insetsMask = WindowInsetsController.APPEARANCE_LIGHT_STATUS_BARS;
+            viewMask = View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR;
+            window.setStatusBarColor(color);
+            break;
+        case NAVIGATION:
+            insetsMask = WindowInsetsController.APPEARANCE_LIGHT_NAVIGATION_BARS;
+            viewMask = View.SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR;
+            window.setNavigationBarColor(color);
+            break;
+        default:
+            throw new RuntimeException("invalid bar type");
+        }
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            return;
+        }
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+            int flags = this.getSystemUiVisibility();
+            if (luminance > 128) {
+               flags |=  viewMask;
+            } else {
+               flags &= ~viewMask;
+            }
+            this.setSystemUiVisibility(flags);
+            return;
+        }
+
+        WindowInsetsController insetsController = window.getInsetsController();
+        if (insetsController == null) {
+            return;
+        }
+        if (luminance > 128) {
+            insetsController.setSystemBarsAppearance(insetsMask, insetsMask);
+        } else {
+            insetsController.setSystemBarsAppearance(0, insetsMask);
+        }
+    }
+
+    private void setStatusColor(int color, int luminance) {
+        this.setBarColor(Bar.STATUS, color, luminance);
+    }
+
+    private void setNavigationColor(int color, int luminance) {
+        this.setBarColor(Bar.NAVIGATION, color, luminance);
+    }
 
 	private void dispatchMotionEvent(MotionEvent event) {
 		for (int j = 0; j < event.getHistorySize(); j++) {
--- a/app/internal/wm/os_android.go
+++ b/app/internal/wm/os_android.go
@@ -42,7 +42,9 @@ import "C"
 import (
 	"errors"
 	"fmt"
+	"gioui.org/internal/f32color"
 	"image"
+	"image/color"
 	"reflect"
 	"runtime"
 	"runtime/debug"
@@ -82,18 +84,22 @@ type window struct {
 // windowState tracks the View or Activity specific state lost when Android
 // re-creates our Activity.
 type windowState struct {
-	cursor *pointer.CursorName
+	cursor          *pointer.CursorName
+	navigationColor *color.NRGBA
+	statusColor     *color.NRGBA
 }
 
 // gioView hold cached JNI methods for GioView.
 var gioView struct {
-	once              sync.Once
-	getDensity        C.jmethodID
-	getFontScale      C.jmethodID
-	showTextInput     C.jmethodID
-	hideTextInput     C.jmethodID
-	postFrameCallback C.jmethodID
-	setCursor         C.jmethodID
+	once               sync.Once
+	getDensity         C.jmethodID
+	getFontScale       C.jmethodID
+	showTextInput      C.jmethodID
+	hideTextInput      C.jmethodID
+	postFrameCallback  C.jmethodID
+	setCursor          C.jmethodID
+	setNavigationColor C.jmethodID
+	setStatusColor     C.jmethodID
 }
 
 // ViewEvent is sent whenever the Window's underlying Android view
@@ -220,6 +226,8 @@ func Java_org_gioui_GioView_onCreateView(env *C.JNIEnv, class C.jclass, view C.j
 		m.hideTextInput = getMethodID(env, class, "hideTextInput", "()V")
 		m.postFrameCallback = getMethodID(env, class, "postFrameCallback", "()V")
 		m.setCursor = getMethodID(env, class, "setCursor", "(I)V")
+		m.setNavigationColor = getMethodID(env, class, "setNavigationColor", "(II)V")
+		m.setStatusColor = getMethodID(env, class, "setStatusColor", "(II)V")
 	})
 	view = C.gio_jni_NewGlobalRef(env, view)
 	wopts := <-mainWindow.out
@@ -235,6 +243,7 @@ func Java_org_gioui_GioView_onCreateView(env *C.JNIEnv, class C.jclass, view C.j
 	handle := C.jlong(view)
 	views[handle] = w
 	w.loadConfig(env, class)
+	w.Option(wopts.opts)
 	applyStateDiff(env, view, windowState{}, w.state)
 	w.setStage(system.StagePaused)
 	w.callbacks.Event(ViewEvent{View: uintptr(view)})
@@ -676,7 +685,18 @@ func (w *window) ReadClipboard() {
 	})
 }
 
-func (w *window) Option(opts *Options) {}
+func (w *window) Option(opts *Options) {
+	if o := opts.NavigationColor; o != nil {
+		w.setState(func(state *windowState) {
+			state.navigationColor = o
+		})
+	}
+	if o := opts.StatusColor; o != nil {
+		w.setState(func(state *windowState) {
+			state.statusColor = o
+		})
+	}
+}
 
 func (w *window) SetCursor(name pointer.CursorName) {
 	w.setState(func(state *windowState) {
@@ -703,6 +723,12 @@ func applyStateDiff(env *C.JNIEnv, view C.jobject, old, state windowState) {
 	if state.cursor != nil && old.cursor != state.cursor {
 		setCursor(env, view, *state.cursor)
 	}
+	if state.navigationColor != nil && old.navigationColor != state.navigationColor {
+		setNavigationColor(env, view, *state.navigationColor)
+	}
+	if state.statusColor != nil && old.statusColor != state.statusColor {
+		setStatusColor(env, view, *state.statusColor)
+	}
 }
 
 func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
@@ -726,6 +752,20 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
 		curID = 0 // TYPE_NULL
 	}
 	callVoidMethod(env, view, gioView.setCursor, jvalue(curID))
+}
+
+func setStatusColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setStatusColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(int(f32color.LinearFromSRGB(color).Luminance()*255)),
+	)
+}
+
+func setNavigationColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setNavigationColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(int(f32color.LinearFromSRGB(color).Luminance()*255)),
+	)
 }
 
 // Close the window. Not implemented for Android.
--- a/app/internal/wm/os_js.go
+++ b/app/internal/wm/os_js.go
@@ -3,7 +3,10 @@
 package wm
 
 import (
+	"fmt"
+	"gioui.org/internal/f32color"
 	"image"
+	"image/color"
 	"strings"
 	"sync"
 	"syscall/js"
@@ -22,6 +25,7 @@ import (
 type window struct {
 	window                js.Value
 	document              js.Value
+	head                  js.Value
 	clipboard             js.Value
 	cnv                   js.Value
 	tarea                 js.Value
@@ -61,6 +65,7 @@ func NewWindow(win Callbacks, opts *Options) error {
 		document:  doc,
 		tarea:     tarea,
 		window:    js.Global().Get("window"),
+		head:      doc.Get("head"),
 		clipboard: js.Global().Get("navigator").Get("clipboard"),
 	}
 	w.requestAnimationFrame = w.window.Get("requestAnimationFrame")
@@ -486,6 +491,9 @@ func (w *window) Option(opts *Options) {
 	if o := opts.WindowMode; o != nil {
 		w.windowMode(*o)
 	}
+	if o := opts.NavigationColor; o != nil {
+		w.navigationColor(*o)
+	}
 }
 
 func (w *window) SetCursor(name pointer.CursorName) {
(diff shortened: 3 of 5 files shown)


Inkeliz:
Using:
```
status := color.RGBA{R: 0, G: 0, B: 255, A: 128}
nav := color.RGBA{R: 255, G: 255, B: 0, A: 128}
w.SetColor(status, nav)
```

On Android:
Screenshot_2021-01-09-15-55-14-365_org gioui kitchen


On Chrome:
Screenshot_2021-01-09-16-26-36-773_com android chrome

Inkeliz:
For Windows, it's not possible to do on Win32. It may possible using [UWP](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.ViewManagement.ApplicationViewTitleBar?redirectedfrom=MSDN&view=winrt-19041), but [Golang seems to not handle UWP well](https://github.com/golang/go/issues/21805). Also, it should replace all Win32 calls to [CoreWindow](https://docs.microsoft.com/en-us/uwp/api/windows.ui.core.corewindow?view=winrt-19041), so it's not a easy task.

_I'm not sure if an Win32 app inside the [MSIX](https://docs.microsoft.com/en-us/windows/msix/packaging-tool/create-app-package) will make it capable to change the color, but I think it will not work._

app/window.go
```diff
@@ -235,6 +236,13 @@ func (w *Window) SetCursorName(name pointer.CursorName) {
 	})
 }
 
+// SetColor changes the current navigation/status bar color.
+func (w *Window) SetColor(top, bottom color.RGBA) {
```
eliasnaur, line 240: SetColor is not specific enough, SetNavigationColor and SetStatusColor are better. I suspect you added just one method because it's tedious to add new options. See my PR comment for a suggestion.

eliasnaur:
Nice work. However, I think it's time to unify options to app.Option instead of adding new methods to Window. That is, I think we should add a 

```
func (w *Window) SetOptions(opts ...Option)
```

and make navigation and status colors Options, along with the existing cursor name.

By using Option, we avoid adding new methods to Window for each new option, and we'll be closer to have the Title, Size, Min|MaxSize options be configurable after NewWindow.

eliasnaur (changes requested):

app/window.go
Inkeliz: Exactly. `SetColor` is not good, but `SetNavigationColor` and `SetStatusColor` seems worse. 🤦‍♂️
I think the `SetOptions` makes sense.

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,72 @@ private void setCursor(int id) {
 		setPointerIcon(pointerIcon);
 	}
 
+    private void setStatusColor(int color, int luminance) {
```
Inkeliz, line 137: Maybe both `setStatusColor` and  `setNavigationColor` can be replace by `setColor(bar String, ....)`, in order to re-use all the code. It may simplify the code, but adds a lot of `if (bar == "navigation)`, since some constants are only available in some versions (`WindowInsetsController.APPEARANCE_LIGHT_NAVIGATION_BAR`, for instance).

app/internal/wm/os_android.go
```diff
@@ -728,6 +753,20 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
 	callVoidMethod(env, view, gioView.setCursor, jvalue(curID))
 }
 
+func setStatusColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setStatusColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(int64((0.2126*float64(color.R))+(0.7152*float64(color.G))+(0.0722*float64(color.B)))),
```
Inkeliz, line 759: The "luminance" is used to switch the button/font color. Maybe add some comments about it? Or let it customizable, so you can choose between White or Black icons, on the Status-Bar?.

app/window.go
```diff
@@ -525,4 +526,20 @@ func MinSize(w, h unit.Value) Option {
...
+// StatusBarColor sets the color of the status bar (which usually contains notification icons
+// and system icons).
+func StatusBarColor(color color.NRGBA) Option {
+	return func(opts *wm.Options) {
+		opts.StatusColor = &color
+	}
+}
+
+// NavigationBarColor sets the color of the navigation bar (which contains the navigation controls,
+// such as "Back", "Home" or "Website Address Bar").
```
eliasnaur, line 538: I think it's better to be specific:

// NavigationBarColor sets the color of the navigation bar on Android, or the address bar in browsers.

Consider shortening to just "NavigationColor".

app/window.go
```diff
@@ -525,4 +526,20 @@ func MinSize(w, h unit.Value) Option {
 	}
 }
 
+// StatusBarColor sets the color of the status bar (which usually contains notification icons
+// and system icons).
```
eliasnaur, line 530: Be specific:

// StatusBarColor sets the color of the Android status bar.

Consider shortening to "StatusColor".

app/internal/wm/os_js.go
```diff
@@ -583,6 +590,29 @@ func (w *window) windowMode(mode WindowMode) {
 	}
 }
 
+func (w *window) navigationColor(c color.NRGBA) {
+	if c.A != 255 {
+		// Browser doesn't support alpha.
+		c = color.NRGBA{
+			R: c.A*c.R + (255 - c.A),
+			G: c.A*c.G + (255 - c.A),
+			B: c.A*c.B + (255 - c.A),
```
eliasnaur, line 599: This doesn't seem right. For example, if c.A = 254, c.R = 255, the result is much larger than 255. Except that it isn't, because the result overflows uint8.

app/internal/wm/os_android.go
eliasnaur: As long as you're ok with the luminance-based selection between dark and light, let's keep the API simple.

app/internal/wm/os_android.go
```diff
@@ -728,6 +753,20 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
...
+	)
+}
+
+func setNavigationColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setNavigationColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(int64((0.2126*float64(color.R))+(0.7152*float64(color.G))+(0.0722*float64(color.B)))),
```
eliasnaur, line 766: Extract into function that describes the luminance formula. Here and above.

app/internal/wm/GioView.java
eliasnaur: Java constants are resolved at compile time, so they don't need version guards.

Inkeliz: I change it to a single function `setBarColor` to reduce the duplicated code. 🎉

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,64 @@ private void setCursor(int id) {
 		setPointerIcon(pointerIcon);
 	}
 
+    private void setBarColor(String t, int color, int luminance) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+            return;
+        }
+
+        Window window = ((Activity) this.getContext()).getWindow();
+        if (window == null) {
```
eliasnaur, line 143: When is window null? If setBarColor is guaranteed to call while the activity is running, let's drop this check and let a NullPointerException alert us if we violate the guarantee.

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,64 @@ private void setCursor(int id) {
...
+            return;
+        }
+
+        int insetsMask = 0;
+        int viewMask = 0;
+
+        if (t == "Status") {
```
eliasnaur, line 150: This won't (shouldn't) work; you need to do `t.equals("Status")`. However, let's use the typesafer enums instead of free-form strings to distinguish the cases.

app/internal/wm/os_android.go
```diff
@@ -728,6 +753,24 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
...
+
+func setNavigationColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setNavigationColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(luminosity(color)),
+	)
+}
+
+func luminosity(color color.NRGBA) int64 {
+	return int64((0.2126 * float64(color.R)) + (0.7152 * float64(color.G)) + (0.0722 * float64(color.B)))
```
eliasnaur, line 771: Can you add a comment with a reference to the theory behind this formula?

app/internal/wm/os_android.go
```diff
@@ -728,6 +753,24 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
...
+}
+
+func setNavigationColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setNavigationColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(luminosity(color)),
+	)
+}
+
+func luminosity(color color.NRGBA) int64 {
```
eliasnaur, line 770: Why int64 specifically? Why isn't `int` good enough?

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,64 @@ private void setCursor(int id) {
...
+            window.setNavigationBarColor(color);
+        }
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            return;
+        }
+
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.Q) {
+            WindowInsetsController insetsController = window.getInsetsController();
+            if (insetsController == null) {
```
eliasnaur, line 168: Is insetsController ever null? If not, I'd rather leave a potential NPE than silently ignore a color change.

Inkeliz: The documentation say "This value may be null.", https://developer.android.com/reference/android/view/Window#getInsetsController(). 🤔

app/internal/wm/GioView.java
Inkeliz: It works. But, I replace with enums. 👍

app/internal/wm/os_android.go
Inkeliz: I used int64 because jvalue is int64. I switch to `int`.

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,66 @@ private void setCursor(int id) {
 		setPointerIcon(pointerIcon);
 	}
 
+	enum Bar {
```
eliasnaur, line 137: Indentation seems off.

app/internal/wm/os_android.go
```diff
@@ -728,6 +753,26 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
...
+}
+
+func setNavigationColor(env *C.JNIEnv, view C.jobject, color color.NRGBA) {
+	callVoidMethod(env, view, gioView.setNavigationColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(luminance(color)),
+	)
+}
+
+func luminance(color color.NRGBA) int64 {
```
eliasnaur, line 770: Still int64.

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,66 @@ private void setCursor(int id) {
...
+
+    private void setBarColor(Bar t, int color, int luminance) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+            return;
+        }
+
+        Window window = ((Activity) this.getContext()).getWindow();
+
+        int insetsMask = 0;
+        int viewMask = 0;
```
eliasnaur, line 150: Drop the initializers ("= 0") and let the Java compiler ensure they're always set.

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,66 @@ private void setCursor(int id) {
...
+
+        if (t == Bar.STATUS) {
```
eliasnaur, line 152: Nit: a switch is slightly clearer here.

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,66 @@ private void setCursor(int id) {
...
+            viewMask = View.SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR;
+
+            window.setNavigationBarColor(color);
+        }
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            return;
+        }
+
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.Q) {
```
eliasnaur, line 168: Nit: use Go return early idiom and keep all version comparisions less-than:

```
if (VERSION_SDK_INT < VERSION_CODES.R) {
   ...
   this.setSystemUiVisibility(flags);
   return;
}

WindowInsetsController insetsController = window.getInsetsController();
...
```

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,70 @@ private void setCursor(int id) {
...
+            window.setStatusBarColor(color);
+            break;
+        case NAVIGATION:
+            insetsMask = WindowInsetsController.APPEARANCE_LIGHT_NAVIGATION_BARS;
+            viewMask = View.SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR;
+            window.setNavigationBarColor(color);
+            break;
+        default:
+            return;
+        }
```
eliasnaur, line 165: Drop default case or throw RuntimeException. Ignoring unkown Bars frustrates future debugging efforts.

app/internal/wm/os_android.go
```diff
@@ -728,6 +753,26 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
...
+	callVoidMethod(env, view, gioView.setNavigationColor,
+		jvalue(uint32(color.A)<<24|uint32(color.R)<<16|uint32(color.G)<<8|uint32(color.B)),
+		jvalue(luminance(color)),
+	)
+}
+
+func luminance(color color.NRGBA) int {
+	// Relative Luminance.
+	// https://en.wikipedia.org/wiki/Relative_luminance.
+	return int((0.2126 * float64(color.R)) + (0.7152 * float64(color.G)) + (0.0722 * float64(color.B)))
```
eliasnaur, line 773: Add a +0.5 to round correctly?

app/internal/wm/os_js.go
```diff
@@ -583,6 +590,19 @@ func (w *window) windowMode(mode WindowMode) {
 	}
 }
 
+func (w *window) navigationColor(c color.NRGBA) {
+	cssColor := fmt.Sprintf("#%06X", []byte{c.R, c.G, c.B})
```
eliasnaur, line 594: What happened to the alpha?

Inkeliz: Google Chrome (which is the only one that supports `theme-color`) doesn't supports `alpha`. The best soution is to pre-compute the alpha, but my last attempt doesn't work. 

Maybe something like:

```
r, g, b, _ := color.RGBAModel.Convert(c).RGBA()
cssColor := fmt.Sprintf("#%06X", []uint8{uint8(r/255), uint8(g/255), uint8(b/255)})
```

Or...

```
r, g, b, a := uint32(c.R), uint32(c.G), uint32(c.B), uint32(c.A)
c.R = uint8(a * r / 255)
c.G = uint8(a * g / 255)
c.B = uint8(a * b / 255)
cssColor :=  fmt.Sprintf("#%06X", []byte{c.R, c.G, c.B})
```

eliasnaur: Perhaps take a look at package internal/f32color.

app/internal/wm/os_android.go
eliasnaur: Oh, and internal/f32color has a RGBA.Luminance method. Use that instead of duplicating the computation.

app/internal/wm/os_js.go
eliasnaur: Specifically, use f32color.LinearFromSRGB to convert the color to linear float32-based color channels. Then you can multiply each color channel with alpha. Finally, set alpha to 1.0 and convert back to sRGB with the SRGB method.

Inkeliz: I take a look at `f32color` and it contains `NRGBAToRGBA_PostAlpha` (https://github.com/gioui/gio/blob/4f9326c6d17c4ec1f5eb86afc0d50f6565c9b688/internal/f32color/rgba.go#L96-L112), can I use that directly? 🤔

Inkeliz:
I force-push multiple times to update test-check. It was falling for unknown reasons. 
Also, the "Resubmit Build" doesn't update the tests on GitHub. So, I need to push again. :S

app/internal/wm/GioView.java
```diff
@@ -132,6 +134,70 @@ private void setCursor(int id) {
...
+            break;
+        default:
+            throw new RuntimeException("invalid bar type");
+        }
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            return;
+        }
+
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q) {
```
eliasnaur, line 171: Nit: Use < to match the other checks above:

`if (SDK_INT < VERSION_CODES.R)`

app/internal/wm/os_js.go
```diff
@@ -583,6 +591,26 @@ func (w *window) windowMode(mode WindowMode) {
...
 
+func (w *window) navigationColor(c color.NRGBA) {
+	if c.A != 255 {
+		alpha := float32(c.A) / 255
+		rgba := f32color.LinearFromSRGB(c)
+		rgba.R *= alpha
+		rgba.G *= alpha
+		rgba.B *= alpha
+		rgba.A = 1
+		c = rgba.SRGB()
```
eliasnaur, line 602: I think you can just use f32color.NRGBAToRGBA and throw its alpha channel away. RGBA has alpha pre-multiplied, so the alpha channel can safely be ignored when not blending with the background.

```
rgba := f32color.NRGBAtoRGBA(c)
...
theme.Set("content", fmt.Sprintf("#%08X", []byte{rgba.R, rgba.G, rgba.B, 255}))
```

Please do test, say, white color with alpha = 180 and test that the resulting color has R,G,B = 180.

app/internal/wm/os_js.go
```diff
@@ -583,6 +591,26 @@ func (w *window) windowMode(mode WindowMode) {
...
+	}
+
+	theme := w.head.Call("querySelector", `meta[name="theme-color"]`)
+	if !theme.Truthy() {
+		theme = w.document.Call("createElement", "meta")
+		theme.Set("name", "theme-color")
+		w.head.Call("appendChild", theme)
+	}
+	theme.Set("content", fmt.Sprintf("#%08X", []byte{c.R, c.G, c.B, 255}))
```
eliasnaur, line 611: If theme-color doesn't support alpha, why specify it?

app/internal/wm/os_js.go
```diff
@@ -583,6 +591,26 @@ func (w *window) windowMode(mode WindowMode) {
 	}
 }
 
+func (w *window) navigationColor(c color.NRGBA) {
+	if c.A != 255 {
```
eliasnaur, line 595: Delete this check. The conversion routine should work regardless of the alpha value.

app/internal/wm/os_js.go
Inkeliz: The `f32color.NRGBAToLinearRGBA` returns `180`, when using `{R: 255, G: 255, B: 255, A: 180}`.

eliasnaur: NRGBAtoRGBA is the correct function, but my example was wrong because I forgot that alpha values are linear. You need to use A=128=0x80. I tested

NRGBAtoRGBA(color.NRGBA{A: 0x80, R: 0xff, G: 0xff, B: 0xff}) = color.RGBA{R:0xbc, G:0xbc, B:0xbc, A:0x80}

R,G,B = 0xbc = 188 is not quite 180, but that's the difference between gamma = 2.2 (R,G,B=188) and gamma = 2.0 (R,G,B=180).

Sorry for the confusion.

eliasnaur:
Thanks, merged.

PR #11, opened 2021-01-16
app, app/internal/window: add window.SetOptions for Options changes.

Inkeliz:
Currently, the Options can only be defined on the startup of the
window. This patch makes possible to change Options at anytime.

That patch also makes possible to change Size/MinSize/MaxSize/Title on
Windows and on Linux (x11).

Updates gio#193.

### DIFF
--- a/app/internal/window/os_macos.go
+++ b/app/internal/window/os_macos.go
@@ -128,6 +128,10 @@ func (w *window) SetCursor(name pointer.CursorName) {
 	w.cursor = windowSetCursor(w.cursor, name)
 }
 
+func (w *window) SetOption(option Option) {
+	// Not supported on MacOS.
+}
+
 func (w *window) ShowTextInput(show bool) {}
 
 func (w *window) SetAnimating(anim bool) {
@@ -315,7 +319,7 @@ func gio_onFinishLaunching() {
 	close(launched)
 }
 
-func NewWindow(win Callbacks, opts *Options) error {
+func NewWindow(win Callbacks, opts *Option) error {
 	<-launched
 	errch := make(chan error)
 	runOnMain(func() {
@@ -326,20 +330,26 @@ func NewWindow(win Callbacks, opts *Options) error {
 		}
 		screenScale := float32(C.gio_getScreenBackingScale())
 		cfg := configFor(screenScale)
-		width := cfg.Px(opts.Width)
-		height := cfg.Px(opts.Height)
+		width := cfg.Px(*opts.Width)
+		height := cfg.Px(*opts.Height)
 		// Window sizes is in unscaled screen coordinates, not device pixels.
 		width = int(float32(width) / screenScale)
 		height = int(float32(height) / screenScale)
-		minWidth := cfg.Px(opts.MinWidth)
-		minHeight := cfg.Px(opts.MinHeight)
+		var minWidth, minHeight int
+		if opts.MinWidth != nil && opts.MinHeight != nil {
+			minWidth = cfg.Px(*opts.MinWidth)
+			minHeight = cfg.Px(*opts.MinHeight)
+		}
 		minWidth = int(float32(minWidth) / screenScale)
 		minHeight = int(float32(minHeight) / screenScale)
-		maxWidth := cfg.Px(opts.MaxWidth)
-		maxHeight := cfg.Px(opts.MaxHeight)
+		var maxWidth, maxHeight int
+		if opts.MinWidth != nil && opts.MinHeight != nil {
+			maxWidth = cfg.Px(*opts.MaxWidth)
+			maxHeight = cfg.Px(*opts.MaxHeight)
+		}
 		maxWidth = int(float32(maxWidth) / screenScale)
 		maxHeight = int(float32(maxHeight) / screenScale)
-		title := C.CString(opts.Title)
+		title := C.CString(*opts.Title)
 		defer C.free(unsafe.Pointer(title))
 		errch <- nil
 		win.SetDriver(w)
@@ -357,7 +367,7 @@ func NewWindow(win Callbacks, opts *Options) error {
 	return <-errch
 }
 
-func newWindow(opts *Options) (*window, error) {
+func newWindow(opts *Option) (*window, error) {
 	view := viewFactory()
 	if view == 0 {
 		return nil, errors.New("CreateWindow: failed to create view")
--- a/app/internal/window/os_windows.go
+++ b/app/internal/window/os_windows.go
@@ -55,9 +55,10 @@ type window struct {
 	mu        sync.Mutex
 	animating bool
 
+	cfg    unit.Metric
 	minmax winConstraints
 	deltas winDeltas
-	opts   *Options
+	opts   *Option
 }
 
 const (
@@ -94,7 +95,7 @@ func Main() {
 	select {}
 }
 
-func NewWindow(window Callbacks, opts *Options) error {
+func NewWindow(window Callbacks, opts *Option) error {
 	cerr := make(chan error)
 	go func() {
 		// GetMessage and PeekMessage can filter on a window HWND, but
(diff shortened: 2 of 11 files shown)


app/internal/window/window.go
```diff
@@ -14,11 +14,61 @@ import (
...
+		}
+	case old.MinWidth != opts.MinWidth || old.MinHeight != opts.MinHeight:
+		if f := old.OptMinSize.optionCallback; f != nil {
+			f()
+		}
+	case old.MaxWidth != opts.MaxWidth || old.MaxHeight != opts.MaxHeight:
+		if f := old.OptMaxSize.optionCallback; f != nil {
+			f()
+		}
+	}
```
Inkeliz, line 71: That doesn't seems right. But we need to identify what was changed, otherwise it will call all functions (such as Size) even when it's not intended. _Must be a better alternative for that._

eliasnaur:
Haven't looked at the patch yet, but add "Updates gio#193".

eliasnaur (changes requested):

app/internal/window/window.go
eliasnaur: Let's drop the notion that an Option should be a function. I suggest

```
type Option struct {
    Title *string
    MaxWidth *int
    ...
}
```

So it's easy to tell which fields have changed. Option is not a pretty type, but it's only used internally.

Inkeliz: Switching to `struct` with pointer is better. But, why not create new interfaces? 

-----

Instead of change the `window.Window`, it's possible to add another interface:
```
type WindowBoundsSize interface {
SetMaxSize(x, y unit.Value)
SetMinSize(x, y unit.Value)
}
```

Then, create some `app.SetMaxSize(w window.Window, x, y unit.Value)`:

```
func SetMaxSize(w window.Window, x, y unit.Value) {
if ww, ok := w.(WindowBoundsSize); ok {
ww.SetMaxSize(x,y)
}
}
```

-----

In that case some platforms that doesn't support Resize/MaxSize/MinSize/Size (such as Android, JS, iOS) will never implement `WindowMaxSize`. I think it solves the problem of the forever growing `window.Window` interface. For instance (#10 ) adds alot of "no-op" functions just to implement the new window.Window. In that case, a new interface (such as `type WindowNavigationColor interface{}` and `type WindowStatusColor interface{}`) could be created.

eliasnaur: I believe my suggestion is simpler than interfaces, because you don't need new methods and the type-casting. Rather, backend use the settings field if they support it.

Another reason to use the Options struct is to support atomic settings. For example, setting dimensions and maxheight/width in one call. This is important to avoid flickering updates to the native window and for performance if the setting is expensive.

app/internal/window/os_macos.go
```diff
@@ -326,20 +330,26 @@ func NewWindow(win Callbacks, opts *Options) error {
...
-		height := cfg.Px(opts.Height)
+		width := cfg.Px(*opts.Width)
+		height := cfg.Px(*opts.Height)
 		// Window sizes is in unscaled screen coordinates, not device pixels.
 		width = int(float32(width) / screenScale)
 		height = int(float32(height) / screenScale)
-		minWidth := cfg.Px(opts.MinWidth)
-		minHeight := cfg.Px(opts.MinHeight)
+		var minWidth, minHeight int
+		if opts.MinWidth != nil && opts.MinHeight != nil {
```
eliasnaur, line 339: Why not check for MinWidth and MinHeight individually?

app/internal/window/os_macos.go
```diff
@@ -326,20 +330,26 @@ func NewWindow(win Callbacks, opts *Options) error {
...
+			minWidth = cfg.Px(*opts.MinWidth)
+			minHeight = cfg.Px(*opts.MinHeight)
+		}
 		minWidth = int(float32(minWidth) / screenScale)
 		minHeight = int(float32(minHeight) / screenScale)
-		maxWidth := cfg.Px(opts.MaxWidth)
-		maxHeight := cfg.Px(opts.MaxHeight)
+		var maxWidth, maxHeight int
+		if opts.MinWidth != nil && opts.MinHeight != nil {
```
eliasnaur, line 346: Ditto.

app/internal/window/os_macos.go
```diff
@@ -357,7 +367,7 @@ func NewWindow(win Callbacks, opts *Options) error {
 	return <-errch
 }
 
-func newWindow(opts *Options) (*window, error) {
+func newWindow(opts *Option) (*window, error) {
```
eliasnaur, line 370: Drop parameter if unused.

app/internal/window/os_windows.go
```diff
@@ -578,6 +588,48 @@ func (w *window) writeClipboard(s string) error {
 	return nil
 }
 
+func (w *window) SetOption(option Option) {
+	if option.Title != nil {
+		windows.SetWindowText(w.hwnd, *option.Title)
+		w.opts.Title = option.Title
+	}
+	if option.Height != nil && option.Width != nil {
```
eliasnaur, line 596: Check for Height and Width != nil individually. Or perhaps better, combine Width and Height to a single field, Size.

app/internal/window/os_windows.go
```diff
@@ -578,6 +588,48 @@ func (w *window) writeClipboard(s string) error {
...
+		w.opts.Width, w.opts.Height = option.Width, option.Height
+
+		windows.SetWindowPos(w.hwnd, w.hwnd, 0, 0, int32(w.cfg.Px(*option.Width)), int32(w.cfg.Px(*option.Height)), windows.SWP_ASYNCWINDOWPOS|windows.SWP_NOMOVE|windows.SWP_NOZORDER)
+	}
+	if option.MinWidth != nil && option.MinHeight != nil {
```
eliasnaur, line 601: Ditto. Combining MinWidth and MinHeight to MinSize seems better.

app/internal/window/os_windows.go
```diff
@@ -578,6 +588,48 @@ func (w *window) writeClipboard(s string) error {
...
+			cw = mw
+		}
+		if mh := w.cfg.Px(*option.MinHeight); ch < mh {
+			ch = mh
+		}
+		if cw != w.width || ch != w.height {
+			windows.SetWindowPos(w.hwnd, w.hwnd, 0, 0, int32(cw), int32(ch), windows.SWP_ASYNCWINDOWPOS|windows.SWP_NOMOVE|windows.SWP_NOZORDER)
+		}
+	}
+	if option.MaxWidth != nil && option.MaxHeight != nil {
```
eliasnaur, line 616: Ditto for MaxSize.

app/internal/window/window.go
```diff
@@ -64,6 +63,9 @@ type Driver interface {
 	// SetCursor updates the current cursor to name.
 	SetCursor(name pointer.CursorName)
 
+	// SetOption updates the current window Option.
+	SetOption(option Option)
```
eliasnaur, line 67: I prefer the "Options" name for the method; it matches Window.Options and "Set" implies replacing the options, not updating them.

// Options applies options to the window.
Options(options Option)

app/internal/window/window.go
```diff
@@ -14,13 +14,12 @@ import (
 	"gioui.org/unit"
 )
 
-type Options struct {
-	Width, Height       unit.Value
-	MinWidth, MinHeight unit.Value
-	MaxWidth, MaxHeight unit.Value
-	Title               string
+type Option struct {
+	Width, Height       *unit.Value
```
eliasnaur, line 18: Combining these fields seems better in a Point type, analogous to image|f32.Point. Point belongs in package unit, but it's ok to put it in the app/internal/window package for now.

app/internal/window/window.go
```diff
@@ -14,13 +14,12 @@ import (
 	"gioui.org/unit"
 )
 
-type Options struct {
-	Width, Height       unit.Value
-	MinWidth, MinHeight unit.Value
-	MaxWidth, MaxHeight unit.Value
-	Title               string
+type Option struct {
```
eliasnaur, line 17: I suggest keeping the "Options" name; the type does all possible options, even though not all may be set.

eliasnaur (changes requested):

pierrec:
@Inkeliz are you still working on this or can I pick it up where you left?

Inkeliz:
I'm not working on that anymore. It's not passing some tests, _and I can't test on MacOS/iOS_. Feel free to pick it up. 👍 
_I'm closing this PR._

PR #13, opened 2021-02-10
widget: fix bug how f.pos is calculated in widget.Float

vsariola:
setting minimum value for a Float never really worked, although min = 0 worked as intended which is why this probably went unnoticed.


eliasnaur:
Please sign off your change, as described in https://gioui.org/doc/contribute. Thanks, and sorry for the trouble.

eliasnaur:
Thanks, merged.

PR #14, opened 2021-02-20
internal/cmd/convertshaders: add Windows comp shader compilation

egonelbre:
Hi, I started looking into implementing comp shaders for Windows and hit a few errors with the shaders that I'm not sure how to solve.

> error X3663: thread sync operation found in varying flow control, consider reformulating your algorithm so all threads will hit the sync simultaneously:

backdrop.comp (and few others):

    void main() {
        if (mem_error != NO_ERROR) { // <---
            return;
        }
        ...
        barrier(); // <---

> error X3000: syntax error: unexpected token 'vector':

    struct TileSeg {
        vec2 origin;
        vec2 vector; // <--- it looks like "vector" is a reserved word.
        float y_edge;
        TileSegRef next;

> generate: Z:\Temp\shader-convert292189591\kernel4.comp(1000,17-25): error X3708: continue cannot be used in a switch

    case Cmd_Jump:
        cmd_ref = CmdRef(Cmd_Jump_read(cmd_alloc, cmd_ref).new_ref);
        cmd_alloc.offset = cmd_ref.offset;
        continue; // <---

> error X4026: thread sync operation must be in non-varying flow control, due to a potential race condition this sync is illegal, consider adding a sync after reading any values controlling shader execution at this point

    MallocResult alloc_clip_buf(uint link) {
        if (gl_LocalInvocationID.x == 0 && gl_LocalInvocationID.y == 0) {
            MallocResult m = malloc(CLIP_BUF_SIZE * 4);
            if (!m.failed) {
                write_mem(m.alloc, (m.alloc.offset >> 2) + CLIP_LINK_OFFSET, link);
            }
            sh_clip_alloc = m;
        }
        barrier(); // <---
        return sh_clip_alloc;
    }


eliasnaur:
> Hi, I started looking into implementing comp shaders for Windows and hit a few errors with the shaders that I'm not sure how to solve.
> 
> > error X3663: thread sync operation found in varying flow control, consider reformulating your algorithm so all threads will hit the sync simultaneously:
> 
> backdrop.comp (and few others):
> 
> ```
> void main() {
>     if (mem_error != NO_ERROR) { // <---
>         return;
>     }
>     ...
>     barrier(); // <---
> ```
> 

Ouch.

This (and the similar errors) seems be caused by fundamental limits in shader model 5.x. I couldn't find anything about control flow barriers in Microsoft's documentation, but I found http://developer.download.nvidia.com/compute/DevZone/docs/html/DirectCompute/doc/DirectCompute_Programming_Guide.pdf saying that "This function[GroupMemoryBarrierWithGroupSync] cannot appear within dynamic flow control".

I suppose it's not that surprising; Direct3D 11 and SM 5 are from around 2008 while OpenGL ES 3.1 is from 2014, and Vulkan/Metal are even younger.

I don't see a way around shader model 6 and D3D 12. At least backdrop.comp seems to successfully compile using DXC targeting shader model 6. I used http://shader-playground.timjones.io/fe2ad3d9372d4a20f0f7de065e95e139 to experiment.



> > error X3000: syntax error: unexpected token 'vector':
> 
> ```
> struct TileSeg {
>     vec2 origin;
>     vec2 vector; // <--- it looks like "vector" is a reserved word.

Fixed in [spirv-cross master](https://github.com/KhronosGroup/SPIRV-Cross/pull/1614).

>     float y_edge;
>     TileSegRef next;
> ```
> 
> > generate: Z:\Temp\shader-convert292189591\kernel4.comp(1000,17-25): error X3708: continue cannot be used in a switch
> 

Can be worked around (in kernel4.comp or in SPIRV-Cross) I suppose, but hopefully moot in SM 6.

> ```
> case Cmd_Jump:
>     cmd_ref = CmdRef(Cmd_Jump_read(cmd_alloc, cmd_ref).new_ref);
>     cmd_alloc.offset = cmd_ref.offset;
>     continue; // <---
> ```
> 
> > error X4026: thread sync operation must be in non-varying flow control, due to a potential race condition this sync is illegal, consider adding a sync after reading any values controlling shader execution at this point
> 

Hopefully gone in SM 6.

> ```
> MallocResult alloc_clip_buf(uint link) {
>     if (gl_LocalInvocationID.x == 0 && gl_LocalInvocationID.y == 0) {
>         MallocResult m = malloc(CLIP_BUF_SIZE * 4);
>         if (!m.failed) {
>             write_mem(m.alloc, (m.alloc.offset >> 2) + CLIP_LINK_OFFSET, link);
>         }
>         sh_clip_alloc = m;
>     }
>     barrier(); // <---
>     return sh_clip_alloc;
> }
> ```

Ditto.

eliasnaur:
Thanks for taking a stab at this by the way.

egonelbre:
> This (and the similar errors) seems be caused by fundamental limits in shader model 5.x. I couldn't find anything about control flow barriers in Microsoft's documentation, but I found http://developer.download.nvidia.com/compute/DevZone/docs/html/DirectCompute/doc/DirectCompute_Programming_Guide.pdf saying that "This function[GroupMemoryBarrierWithGroupSync] cannot appear within dynamic flow control".

I think this also applies to glsl not just DirectCompute. The only difference is that fxc and dxc complain about it. There seems to be a pending enhancement for adding it to glsllang https://github.com/KhronosGroup/glslang/issues/1389.

From https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/barrier.xhtml:

> For compute shaders, the barrier() function may be placed within flow control, but that flow control must be uniform flow control.

And in https://www.khronos.org/opengl/wiki/Compute_Shader#Shared_memory_coherency:

> There are limitations on how you can call barrier(). However, compute shaders are not as limited as Tessellation Control Shaders in their use of this function. barrier() can be called from flow-control, but it can only be called from uniform flow control. All expressions that lead to the evaluation of a barrier() must be dynamically uniform.

eliasnaur:
> > This (and the similar errors) seems be caused by fundamental limits in shader model 5.x. I couldn't find anything about control flow barriers in Microsoft's documentation, but I found http://developer.download.nvidia.com/compute/DevZone/docs/html/DirectCompute/doc/DirectCompute_Programming_Guide.pdf saying that "This function[GroupMemoryBarrierWithGroupSync] cannot appear within dynamic flow control".
> 
> I think this also applies to glsl not just DirectCompute. The only difference is that fxc complains about it. There seems to be a pending enhancement for adding it to glsllang [KhronosGroup/glslang#1389](https://github.com/KhronosGroup/glslang/issues/1389).
> 

The example program from the issue is indeed not uniform: some threads will not reach the barrier. The corresponding DXC issue has an interesting [comment](https://github.com/microsoft/DirectXShaderCompiler/issues/1306#issuecomment-392185302) (emphasis mine):

>I could see a useful feature in a front end like DXC where you can detect control dependence based on an expression that is **known to be varying, and issue that as a warning**. Because you don't have precise value information you get some **false positives as well, so at best it's a warning**.

If anything, it may be that the fxc compiler is too strict, at least for newer GPUs. Again, I couldn't find anything conclusive in the Direct3D documentation.

> From https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/barrier.xhtml:
> 
> > For compute shaders, the barrier() function may be placed within flow control, but that flow control must be uniform flow control.
> 
> And in https://www.khronos.org/opengl/wiki/Compute_Shader#Shared_memory_coherency:
> 
> > There are limitations on how you can call barrier(). However, compute shaders are not as limited as Tessellation Control Shaders in their use of this function. barrier() can be called from flow-control, but it can only be called from uniform flow control. All expressions that lead to the evaluation of a barrier() must be dynamically uniform.

I saw that as well, but I believe that all our barriers are dynamically uniform. There's a description of [dynamic uniform](https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#Dynamically_uniform_expression) in the wiki. See in particular the section that says

> However, if every shader in the rendering command is given the exact same input value, then it will be dynamically uniform. It is the value of the expression which must be uniform; every shader invocation must get the same value, even if that same value is being read from different memory location.

In other words, dynamically uniform is about runtime control flow, and not restricted to control flow a compiler can prove to be uniform.

egonelbre:
I was testing with dxc and it seems it accepts `cs_5_0` without throwing the error. So, maybe that'll work.

egonelbre:
So, I got `dxc` compilation working, however this bumps the minimum feature level to 11, which means the wine tests don't work anymore :(.

Adding dual compilation with fxc/dxc doesn't sound that pleasing. Also, compute shaders will require level 11 anyways.

eliasnaur:
> So, I got `dxc` compilation working, however this bumps the minimum feature level to 11, which means the wine tests don't work anymore :(.
> 
> Adding dual compilation with fxc/dxc doesn't sound that pleasing. Also, compute shaders will require level 11 anyways.

I can't see a way around keeping fxc for the non-compute shaders; the old renderer needs to support < d3d feature level 11, and the CPU fallback will still need a way to transform textures and compose CPU-rendered images.

egonelbre:
Very well, I guess both need to stay.

I'm not sure what the best approach architecturally is:

1. compile vert/frag shaders with both fxc and dxc... then choose the shaders based on feature level
2. compile vert/frag shaders only with fxc... then mix using fxc and dxc shaders when using compute shaders (I'm not sure whether they need to mix)
3. create a separate backend one for fxc d3d < feature level 11 and one for d3d level 11

I my gut feeling leans towards 1.

eliasnaur:
> Very well, I guess both need to stay.
> 
> I'm not sure what the best approach architecturally is:
> 
>     1. compile vert/frag shaders with both fxc and dxc... then choose the shaders based on feature level
> 
>     2. compile vert/frag shaders only with fxc... then mix using fxc and dxc shaders when using compute shaders (I'm not sure whether they need to mix)
> 
>     3. create a separate backend one for fxc d3d < feature level 11 and one for d3d level 11
> 
> 
> My gut feeling leans towards 1.

2. is simpler, because it leaves the compile path alone, and there's one less blob in shaders.go. I don't think our simple vert/frag shaders are sped up by feature level 5.

egonelbre:
Done. Now uses both fxc/dxc.

eliasnaur:
Thanks, looks good to me.

shader.go is getting quite large (~800kb) and unwieldy. Perhaps we should switch to binary output and use //go:embed? We'll have nicer diffs and the HLSL binary files would be much smaller. Perhaps drop the HLSL source code comment while we're at it.

egonelbre:
Sounds good. I'll need to bump go version for all the tests and then modifications to the generated code. But, yes, that sounds a better approach.

egonelbre:
Thinking about it a bit.

This would bump the minimum required go version to 1.16, maybe it's something that should be discussed before making that change? I'm not sure everyone has updated their local or test environments to use it.

I'll drop the comments in this PR.

eliasnaur:
> Thinking about it a bit.
> 
> This would bump the minimum required go version to 1.16, maybe it's something that should be discussed before making that change? I'm not sure everyone has updated their local or test environments to use it.
> 

Sure, a mailing list post and a Slack announcement should do. This change isn't needed before the rest of the Windows port anyway, so we could leave it out of the tree for a grace period while people upgrade. Or we could merge the change with a flag that disables the new output, leaving shaders.go untouched. Up to you.

egonelbre:
I'll add a flag to disable the new output, that way I have to do less commit juggling.

eliasnaur:
Great. If you feel adventurous, parallelizing shader compilation would also be wonderful.

egonelbre:
Sure, I'll cleanup/parallelize the shader generation in a separate PR. I was already feeling the generation perf issue myself :).

eliasnaur:
Thanks, merged. I amended the commit to include app/headless shaders as well.

PR #15, opened 2021-02-21
cmd/gogio: fix icon/resources for Android

Inkeliz:
Previously that patch, gogio unzip the `link.apk` (generated by AAPT2) to an temporary folder and then compress it again to a new `app.ap_` file.

For some unknown reason, that unzip-then-zip doesn't work. The resources are included but is corrupted in somehow. That PR aims to fix that by avoid the extraction to an temporary folder.


cmd/gogio/androidbuild.go
```diff
@@ -455,45 +455,82 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
...
 	}()
-	apkw := newZipWriter(ap_)
-	defer apkw.Close()
-	err = filepath.Walk(apkDir, func(path string, f os.FileInfo, err error) error {
+	unsignedAPKZip := zip.NewWriter(unsignedAPKFile)
+	defer unsignedAPKZip.Close()
+
+	// Copy files from linkAPK to unsignedAPK:
+	for _, f := range linkAPKZip.File {
+		h := f.FileHeader
```
eliasnaur, line 482: I think

```
h := zip.FileHeader{
    Name: f.Name,
    // Other required fields?
    ...
}
```

is safer than re-using and clearing fields from the original.

cmd/gogio/androidbuild.go
```diff
@@ -455,45 +455,82 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
 	// The Go standard library archive/zip doesn't support appending to zip
 	// files. Unpack the apk from aapt2 and re-zip its contents along with
```
eliasnaur, line 456: Update comment to match your new implementation.

cmd/gogio/androidbuild.go
```diff
@@ -455,45 +455,82 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
...
-			apkw.Add(zpath, path)
+		if _, err := io.Copy(w, r); err != nil {
+			return err
 		}
-		return nil
-	})
-	if err != nil {
+	}
+
+	// Append new files (that doesn't exists inside the link.apk):
```
eliasnaur, line 502: Nit: End comment sentences with a period, not colons. Here and below.

eliasnaur (changes requested):
I'd love to know what corrupts the resources, but the change looks good in any case.

Inkeliz:
I don't know what is the cause. Maybe it could be because Windows uses `\` as path, and that path is provide as "name" or some compression is wrong?! 

-----

That is what the Android Studio sees before:
gio-res-old

It's how it's detecting now:
gio-res-new

Since the "res" folder doesn't appears on Android Studio (on the previous version), I tried to "rewrite" the zip/unzip stuff and works. But, the root-cause is unknown. 😜

eliasnaur:
Thanks, merged.

PR #16, opened 2021-02-22
cmd/gogio: [android] allow custom signature-key

Inkeliz:
That change makes possible to provide custom PKCS#12/JavaKeystore using `-key` and the password with `-pass`. Also, it's possible to use PKCS#8 key along with X509 certificate, using `-key` and `-cert`.

By default the gogio will use the `debug.keystore`, if no key is provided.

### DIFF
--- a/cmd/gogio/androidbuild.go
+++ b/cmd/gogio/androidbuild.go
@@ -548,43 +548,57 @@ func signAPK(tmpDir string, tools *androidTools, bi *buildInfo) error {
 	if err != nil {
 		return err
 	}
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return err
-	}
-	keystore := filepath.Join(home, ".android", "debug.keystore")
-	if _, err := os.Stat(keystore); err != nil {
-		keystore = filepath.Join(tmpDir, "sign.keystore")
-		keytool, err := findKeytool()
-		if err != nil {
-			return err
-		}
-		_, err = runCmd(exec.Command(
-			keytool,
-			"-genkey",
-			"-keystore", keystore,
-			"-storepass", "android",
-			"-alias", "android",
-			"-keyalg", "RSA", "-keysize", "2048",
-			"-validity", "10000",
-			"-noprompt",
-			"-dname", "CN=android",
-		))
-		if err != nil {
+
+	if bi.key == "" {
+		if err := defaultAndroidKeystore(tmpDir, bi); err != nil {
 			return err
 		}
 	}
+
 	_, err = runCmd(exec.Command(
 		filepath.Join(tools.buildtools, "apksigner"),
 		"sign",
-		"--ks-pass", "pass:android",
-		"--ks", keystore,
+		"--ks-pass", "pass:"+bi.password,
+		"--ks", bi.key,
 		apkFile,
 	))
 	if err != nil {
 		return err
 	}
 	return nil
+}
+
+func defaultAndroidKeystore(tmpDir string, bi *buildInfo) error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+
+	// Use debug.keystore, if exists.
+	bi.key = filepath.Join(home, ".android", "debug.keystore")
+	bi.password = "android"
+	if _, err := os.Stat(bi.key); err == nil {
+		return nil
+	}
+
+	// Generate new key.
+	bi.key = filepath.Join(tmpDir, "sign.keystore")
+	keytool, err := findKeytool()
+	if err != nil {
+		return err
+	}
+	_, err = runCmd(exec.Command(
+		keytool,
+		"-genkey",
+		"-keystore", bi.key,
+		"-storepass", bi.password,
+		"-alias", "android",
+		"-keyalg", "RSA", "-keysize", "2048",
+		"-validity", "10000",
+		"-noprompt",
+		"-dname", "CN=android",
+	))
+	return err
 }
 
 func findNDK(androidHome string) (string, error) {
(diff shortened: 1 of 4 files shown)


cmd/gogio/androidbuild.go
```diff
@@ -548,43 +548,67 @@ func signAPK(tmpDir string, tools *androidTools, bi *buildInfo) error {
...
+	}
+
+	_, err = runCmd(exec.Command(filepath.Join(tools.buildtools, "apksigner"), append(buildParams, apkFile)...))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func defaultAndroidKeystore(tmpDir string, bi *buildInfo) error {
```
Inkeliz, line 581: I'm trying to add the "Android App Bundle", which may be the my next PR. The App Bundle uses `jarsigner` instead of `apksigner`, so the `defaultAndroidKeystore` function can be re-used.

eliasnaur:
I don't think this change is worth it. Why not have an extra line in my build script for release signing?

Inkeliz:
I'm not sure if I understand. But, I think it's simpler to just specify the `-key` and `-pass`. Considering that `gogio` already signs, it's not so difficult and I don't think it's a large change. If you already use `gogio` for testing, why not use that for release? 😐

----

Also, the same `-key`/`-pass` can be used to sign Windows binaries too, using the same `-key` and `-pass`. The only problem is that it requires the `signtool.exe` (but I'm looking into how it works and porting it to Golang). _For iOS/MacOS already requires some signature, but I don't know how it works and I can't test. But, seems reasoable to set some custom key._

whereswaldon:
@eliasnaur I understand your thinking around just managing this externally, but I think our current posture is inconsistent.

If we already have features to facilitate the signing of applications, it seems silly not to allow those features to be used to build signed releases. One of the killer features of gio is the simplicity and portability of the go-based toolchain. We lose some of those benefits if we force the user to install extra tooling for building releases when they didn't need those tools for development builds.

If gogio didn't already do signing, I could see arguing the other direction. As things stand now, I'd rather see this be incorporated.

Just my viewpoint though. Ultimately up to you!

cmd/gogio/main.go
```diff
@@ -34,7 +34,9 @@ var (
 	linkMode      = flag.String("linkmode", "", "set the -linkmode flag of the go tool")
 	extraLdflags  = flag.String("ldflags", "", "extra flags to the Go linker")
 	extraTags     = flag.String("tags", "", "extra tags to the Go tool")
-	iconPath      = flag.String("icon", "", "Specify an icon for iOS and Android")
+	iconPath      = flag.String("icon", "", "specify an icon for iOS and Android")
+	signKey       = flag.String("key", "", "specify the path of the keystore to be used to sign binaries.")
```
eliasnaur, line 38: Let's call it "signkey" to  emphasize the purpose. Add "Android" in the help text to make it clearer it only applied there.

Inkeliz:
I force-push to remove the `-cert`, and simply the code. So, only keystore (such as PFX/JKS) will be supported. I think that is enough, _since must of users will generate their keys using `keytool -genkey` (which is a keystore)._

cmd/gogio/main.go
```diff
@@ -34,7 +34,9 @@ var (
...
+	signPass      = flag.String("key_pass", "", "specify the password to decrypt the keystore.")
```
eliasnaur, line 39: I think "signpass" is prettier.

What about the key alias? Perhaps rename signpass to signcred ("credentials) and format it like this: "<alias>:<password>"?

eliasnaur (commented):
Ok, let's go ahead with this.

cmd/gogio/main.go
Inkeliz: I never used the `alias`. 🤔

But it should be fine to be `[<alias>]:<password>`, so either `alias:password` or `password` will be accepted. The only problem is if the password itself have `:` character. 🙄

eliasnaur: I suppose we could leave out alias for now, and add it when needed.

cmd/gogio/help.go
```diff
@@ -59,4 +59,8 @@ The -work flag prints the path to the working directory and suppress
 its deletion.
 
 The -x flag will print all the external commands executed by the gogio tool.
+
+The -signkey flag specifies the path of the keystore, used for signature for Android.
```
eliasnaur, line 63: Nit:

The -signkey flag specifies the path of the keystore, used for signing Android apk files.

cmd/gogio/help.go
```diff
@@ -59,4 +59,8 @@ The -work flag prints the path to the working directory and suppress
 its deletion.
 
 The -x flag will print all the external commands executed by the gogio tool.
+
+The -signkey flag specifies the path of the keystore, used for signature for Android. 
+
+The -signpass flag specifies the password of the signkey, ignored if no key is provided.
```
eliasnaur, line 65: Nit:

The -signpass flag specifies the password of the keystore, ignored if -signkey is not provided.

cmd/gogio/main.go
```diff
@@ -34,7 +34,9 @@ var (
 	linkMode      = flag.String("linkmode", "", "set the -linkmode flag of the go tool")
 	extraLdflags  = flag.String("ldflags", "", "extra flags to the Go linker")
 	extraTags     = flag.String("tags", "", "extra tags to the Go tool")
-	iconPath      = flag.String("icon", "", "Specify an icon for iOS and Android")
+	iconPath      = flag.String("icon", "", "specify an icon for iOS and Android")
+	signKey       = flag.String("signkey", "", "specify the path of the keystore to be used to sign Android APK.")
```
eliasnaur, line 38: Nit: APK => "apk files".

eliasnaur:
Thank you, merged.

PR #17, opened 2021-02-24
io/router: fix negative areas

egonelbre:


eliasnaur:
Thank you for the quick fix *and* test!

PR #18, opened 2021-02-25
cmd/gogio: [android] update target sdk version

Inkeliz:


cmd/gogio/androidbuild.go
```diff
@@ -386,9 +386,9 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
 	}
 
 	// Link APK.
-	// Currently, new apps must have a target SDK version of at least 28.
+	// Currently, new apps must have a target SDK version of at least 29.
```
eliasnaur, line 389: Why not level 30 while we're at it? https://developer.android.com/distribute/best-practices/develop/target-sdk states that level 30 is requires as of August.

eliasnaur:
Thanks, merged.

PR #20, opened 2021-02-27
Two tiny fixes

egonelbre:
* Fix hash to gioui.org which was breaking installation.
* Fix copy of lock in tests.


eliasnaur:
Thanks, merged with a `go mod tidy`'ed go.sum.

PR #21, opened 2021-02-28
layout: automatically add system inset for NewContext

egonelbre:
Currently every user needs to manually adjust for system insets.
This is rather verbose and most don't need to deviate from this behavior.

To disable the automatic adjustment, use:

```
e.Insets = system.Insets{}
ctx := layout.NewContext(ops, e)
```

### DIFF
--- a/layout/context.go
+++ b/layout/context.go
@@ -5,6 +5,7 @@ package layout
 import (
 	"time"
 
+	"gioui.org/f32"
 	"gioui.org/io/event"
 	"gioui.org/io/system"
 	"gioui.org/op"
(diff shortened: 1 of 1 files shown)


whereswaldon:
Before merging this, please send an email and post in slack. This change will break the layout of everyone who was previously compensating for the insets manually. That being said, I want this change. I think it removes a common footgun from gio.

egonelbre:
@whereswaldon I'm unsure whether to do it before or after it is "approved".

eliasnaur:
I think this is a net win, and not too weird to disable. Merged.

PR #22, opened 2021-02-28
layout: add Fit for widget scaling

egonelbre:
Currently adds three different variants `Contain`, `Cover`, `ScaleDown` and `Fill`. These behave similarly to object-fit.

This can be combined with other layouts. A common example could be:

```
return layout.Center.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
	return layout.Cover.Layout(gtx, image.Layout)
})
```

### DIFF
--- a/widget/fit.go
+++ b/widget/fit.go
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+package widget
+
+import (
+	"image"
+
+	"gioui.org/f32"
+	"gioui.org/layout"
+	"gioui.org/op"
+	"gioui.org/op/clip"
+)
+
+// Fit scales a widget to fit and clip to the constraints.
+type Fit uint8
+
+const (
+	// Unscaled does not alter the scale of a widget.
+	Unscaled Fit = iota
+	// Contain scales widget as large as possible without cropping
+	// and it preserves aspect-ratio.
+	Contain
+	// Cover scales the widget to cover the constraint area and
+	// preserves aspect-ratio.
+	Cover
+	// ScaleDown scales the widget smaller without cropping,
+	// when it exceeds the constraint area.
+	// It preserves aspect-ratio.
+	ScaleDown
+	// Fill stretches the widget to the constraints and does not
+	// preserve aspect-ratio.
+	Fill
+)
+
+// scale adds clip and scale operations to fit dims to the constraints.
+// It positions the widget to the appropriate position.
+// It returns dimensions modified accordingly.
+func (fit Fit) scale(gtx layout.Context, pos layout.Direction, dims layout.Dimensions) layout.Dimensions {
+	widgetSize := dims.Size
+
+	if fit == Unscaled || dims.Size.X == 0 || dims.Size.Y == 0 {
+		dims.Size = gtx.Constraints.Constrain(dims.Size)
+		clip.Rect{Max: dims.Size}.Add(gtx.Ops)
+
+		offset := pos.Position(widgetSize, dims.Size)
+		op.Offset(layout.FPt(offset)).Add(gtx.Ops)
+		dims.Baseline += offset.Y
+		return dims
+	}
+
+	scale := f32.Point{
+		X: float32(gtx.Constraints.Max.X) / float32(dims.Size.X),
+		Y: float32(gtx.Constraints.Max.Y) / float32(dims.Size.Y),
+	}
+
+	switch fit {
+	case Contain:
+		if scale.Y < scale.X {
+			scale.X = scale.Y
+		} else {
+			scale.Y = scale.X
+		}
+	case Cover:
+		if scale.Y > scale.X {
+			scale.X = scale.Y
+		} else {
+			scale.Y = scale.X
+		}
+	case ScaleDown:
+		if scale.Y < scale.X {
+			scale.X = scale.Y
+		} else {
+			scale.Y = scale.X
+		}
+
+		// The widget would need to be scaled up, no change needed.
+		if scale.X >= 1 {
+			dims.Size = gtx.Constraints.Constrain(dims.Size)
+			clip.Rect{Max: dims.Size}.Add(gtx.Ops)
+
+			offset := pos.Position(widgetSize, dims.Size)
+			op.Offset(layout.FPt(offset)).Add(gtx.Ops)
+			dims.Baseline += offset.Y
+			return dims
+		}
+	case Fill:
+	}
+
+	var scaledSize image.Point
+	scaledSize.X = int(float32(widgetSize.X) * scale.X)
+	scaledSize.Y = int(float32(widgetSize.Y) * scale.Y)
+	dims.Size = gtx.Constraints.Constrain(scaledSize)
+	dims.Baseline = int(float32(dims.Baseline) * scale.Y)
+
+	clip.Rect{Max: dims.Size}.Add(gtx.Ops)
+
+	offset := pos.Position(scaledSize, dims.Size)
+	op.Affine(f32.Affine2D{}.
+		Scale(f32.Point{}, scale).
+		Offset(layout.FPt(offset)),
+	).Add(gtx.Ops)
+
+	dims.Baseline += offset.Y
+
+	return dims
+}
--- a/layout/layout.go
+++ b/layout/layout.go
@@ -173,27 +173,37 @@ func (d Direction) Layout(gtx Context, w Widget) Dimensions {
 	if sz.Y < cs.Min.Y {
 		sz.Y = cs.Min.Y
 	}
-	var p image.Point
-	switch d {
-	case N, S, Center:
-		p.X = (sz.X - dims.Size.X) / 2
-	case NE, SE, E:
-		p.X = sz.X - dims.Size.X
-	}
-	switch d {
-	case W, Center, E:
-		p.Y = (sz.Y - dims.Size.Y) / 2
-	case SW, S, SE:
-		p.Y = sz.Y - dims.Size.Y
-	}
-	stack := op.Save(gtx.Ops)
+
+	defer op.Save(gtx.Ops).Load()
+	p := d.Position(dims.Size, sz)
 	op.Offset(FPt(p)).Add(gtx.Ops)
 	call.Add(gtx.Ops)
-	stack.Load()
+
 	return Dimensions{
 		Size:     sz,
 		Baseline: dims.Baseline + sz.Y - dims.Size.Y - p.Y,
 	}
+}
+
+// Position calculates widget position according to the direction.
+func (d Direction) Position(widget, bounds image.Point) image.Point {
+	var p image.Point
+
+	switch d {
+	case N, S, Center:
+		p.X = (bounds.X - widget.X) / 2
+	case NE, SE, E:
+		p.X = bounds.X - widget.X
+	}
+
+	switch d {
+	case W, Center, E:
+		p.Y = (bounds.Y - widget.Y) / 2
+	case SW, S, SE:
+		p.Y = bounds.Y - widget.Y
+	}
+
+	return p
 }
 
 // Spacer adds space between widgets.
--- a/widget/fit_test.go
+++ b/widget/fit_test.go
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+package widget
+
+import (
+	"bytes"
+	"encoding/binary"
+	"image"
+	"math"
+	"testing"
+
+	"gioui.org/f32"
+	"gioui.org/layout"
+	"gioui.org/op"
+)
+
+func TestFit(t *testing.T) {
+	type test struct {
+		Dims   image.Point
+		Scale  f32.Point
+		Result image.Point
+	}
+
+	fittests := [...][]test{
+		Unscaled: {
+			{
+				Dims:   image.Point{0, 0},
+				Scale:  f32.Point{X: 1, Y: 1},
+				Result: image.Point{X: 0, Y: 0},
+			}, {
+				Dims:   image.Point{50, 25},
+				Scale:  f32.Point{X: 1, Y: 1},
+				Result: image.Point{X: 50, Y: 25},
+			}, {
+				Dims:   image.Point{50, 200},
+				Scale:  f32.Point{X: 1, Y: 1},
+				Result: image.Point{X: 50, Y: 100},
+			}},
+		Contain: {
+			{
+				Dims:   image.Point{50, 25},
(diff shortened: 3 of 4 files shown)


layout/example_test.go
```diff
@@ -127,6 +127,30 @@ func ExampleList() {
 	// 5
 }
 
+func ExampleContain() {
+	gtx := layout.Context{
+		Ops: new(op.Ops),
+		// Rigid constraints with both minimum and maximum set.
```
eliasnaur, line 133: These constraints don't seem rigid to me.

layout/example_test.go
```diff
@@ -127,6 +127,30 @@ func ExampleList() {
...
+		Constraints: layout.Constraints{
+			Max: image.Point{X: 100, Y: 100},
+		},
+	}
+
+	dims := layout.Contain.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
```
eliasnaur, line 139: This is an example; can you describe what it intends to show?

layout/scale.go
```diff
@@ -0,0 +1,79 @@
...
+	"gioui.org/f32"
+	"gioui.org/op"
+	"gioui.org/op/clip"
+)
+
+// Fit scales the content to fit to the constraints.
+type Fit uint8
+
+const (
+	Contain Fit = iota
```
eliasnaur, line 15: Add comments to each Fit; they're not obvious to me.

eliasnaur (changes requested):
Fit is general and seems useful but does it belong in gioui.org/layout? If so, do all variants belong? A way to find out would be to convert (some of) gio-example or some other representative project to use them. Another is to put Fit in gioui.org/x and let time answer.

egonelbre:
> Fit is general and seems useful but does it belong in gioui.org/layout?

This has come up a few times in relation to image sizing.

And all the variants are useful:

1. Contain - scale an icon (either vector or pixel) into a particular rectangle.
2. Cover - use images as backgrounds for headers.
3. Fill - use gradient images as the background, which can be stretched arbitrarily.
4. ScaleDown - showing a preview of an image.
 
> Another is to put Fit in gioui.org/x and let time answer.

Sure, I'm also fine to put it into gioui.org/x, however I'm relatively confident that it'll be needed any time you need to interact with images.

layout/scale.go
```diff
@@ -0,0 +1,79 @@
...
+	case Fill:
+	}
+
+	dims.Size.X = int(float32(dims.Size.X) * scale.X)
+	dims.Size.Y = int(float32(dims.Size.Y) * scale.Y)
+	dims.Size = gtx.Constraints.Constrain(dims.Size)
+	dims.Baseline = int(float32(dims.Baseline) * scale.Y)
+
+	stack := op.Save(gtx.Ops)
+	defer stack.Load()
```
eliasnaur, line 68: Nit: put on one line?

layout/scale.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		if scale.X >= 1 {
+			call.Add(gtx.Ops)
+			return dims
+		}
+	case Fill:
+	}
+
+	dims.Size.X = int(float32(dims.Size.X) * scale.X)
+	dims.Size.Y = int(float32(dims.Size.Y) * scale.Y)
```
eliasnaur, line 63: Add 0.5 to properly round sizes? Same goes for Baseline.

eliasnaur (commented):
> > Fit is general and seems useful but does it belong in gioui.org/layout?
> 
> This has come up a few times in relation to image sizing.
> 
> And all the variants are useful:
> 
>     1. Contain - scale an icon (either vector or pixel) into a particular rectangle.
> 
>     2. Cover - use images as backgrounds for headers.
> 
>     3. Fill - use gradient images as the background, which can be stretched arbitrarily.
> 
>     4. ScaleDown - showing a preview of an image.
> 
> 
> > Another is to put Fit in gioui.org/x and let time answer.
> 
> Sure, I'm also fine to put it into gioui.org/x, however I'm relatively confident that it'll be needed any time you need to interact with images.

Ok. Does layout.Fit come up in any other case than images? And what about widget.Image, can it change to work with Fit? Be replaced by Fit?

layout/example_test.go
egonelbre: Yeah, this didn't make sense as an example, I converted it into a proper test.

layout/example_test.go
egonelbre: Fixed.

layout/scale.go
egonelbre: I currently added this, however I'm not quite certain about it. I suspect this could end up causing a "half-pixel" gap in some scenarios.

eliasnaur: Good point, leave it out. Dimensions shouldn't be larger than the scaled size, even half a pixel.

egonelbre:
> Ok. Does layout.Fit come up in any other case than images?

I would say, yes, but less commonly.

You can have a vector "images"/icons (widget.Icon) that you may want to scale to fit. 

However, you could also have text that should fit into some specific area. While it can be done by computing, sometimes it's nicer to automatically size it.

> And what about widget.Image, can it change to work with Fit.

It should already work with widget.Image, such as:

```
layout.Center.Layout(gtx, func(gtx layout.Context) layout.Dimenstions{
	return layout.Cover.Layout(gtx, widget.Image{Src: src}.Layout)
})
```

> Be replaced by Fit?

Not quite, since the paint.ImageOp does not have the "Layout" operation. `widget.Image` also handles DPI scaling, which would be lost.

egonelbre:
>> And what about widget.Image, can it change to work with Fit? 
> It should already work with widget.Image, such as:

Ah, I forgot about the clipping inside widget.Image.

egonelbre:
To make it work with widget.Image it would need to be a property, where the default zero value is `layout.Unscaled`. An untested implementation for the rough idea:

```
// Image is a widget that displays an image.
type Image struct {
	// Src is the image to display.
	Src paint.ImageOp
	// Fit is the image scaling in relation to
	// maximum constraints.
	Fit layout.Fit
	// Scale is the ratio of image pixels to
	// dps. If Scale is zero Image falls back to
	// a scale that match a standard 72 DPI.
	Scale float32
}

const defaultScale = float32(160.0 / 72.0)

func (im Image) Layout(gtx layout.Context) layout.Dimensions {
	scale := im.Scale
	if scale == 0 {
		scale = defaultScale
	}
	return im.Fit.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
		size := im.Src.Size()
		wf, hf := float32(size.X), float32(size.Y)
		w, h := gtx.Px(unit.Dp(wf*scale)), gtx.Px(unit.Dp(hf*scale))

		pixelScale := scale * gtx.Metric.PxPerDp
		op.Affine(f32.Affine2D{}.Scale(f32.Point{}, f32.Pt(pixelScale, pixelScale))).Add(gtx.Ops)
		im.Src.Add(gtx.Ops)
		paint.PaintOp{}.Add(gtx.Ops)

		return layout.Dimensions{Size: image.Pt(w, h)}
	})
}
```

eliasnaur:
> To make it work with widget.Image it would need to be a property, where the default zero value is `layout.Unscaled`. An untested implementation for the rough idea:
>
> ```
> // Image is a widget that displays an image.
> type Image struct {
> 	// Src is the image to display.
> 	Src paint.ImageOp
> 	// Fit is the image scaling in relation to
> 	// maximum constraints.
> 	Fit layout.Fit
> 	// Scale is the ratio of image pixels to
> 	// dps. If Scale is zero Image falls back to
> 	// a scale that match a standard 72 DPI.
> 	Scale float32
> }
>
> const defaultScale = float32(160.0 / 72.0)
>
> func (im Image) Layout(gtx layout.Context) layout.Dimensions {
> 	scale := im.Scale
> 	if scale == 0 {
> 		scale = defaultScale
> 	}
> 	return im.Fit.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
> 		size := im.Src.Size()
> 		wf, hf := float32(size.X), float32(size.Y)
> 		w, h := gtx.Px(unit.Dp(wf*scale)), gtx.Px(unit.Dp(hf*scale))
>
> 		pixelScale := scale * gtx.Metric.PxPerDp
> 		op.Affine(f32.Affine2D{}.Scale(f32.Point{}, f32.Pt(pixelScale, pixelScale))).Add(gtx.Ops)
> 		im.Src.Add(gtx.Ops)
> 		paint.PaintOp{}.Add(gtx.Ops)
>
> 		return layout.Dimensions{Size: image.Pt(w, h)}
> 	})
> }
> ```
>

Thanks. Your example brought up a few points:

First, it's unfortunate the Fit field is needed; layout.Fit is separate,
why not require users to use both? The answer is that without an
explicit Fit, widget.Image likely breaks the implicit contract that
widgets never return dimensions outside their constraints.

In fact, layout.Fit *requires* widgets to break this contract, splitting
widgets in two classes: those that honor the contract and those that
must be decorated by something like Fit.

Can we make the distinction clear? If we had UnconstraintedContext then
yes:

	package layout

	type UnconstrainedWidget func(gtx UnconstrainedContext) Dimensions

	...

	func (fit Fit) Layout(gtx Context, w UnconstrainedWidget) Dimensions

However, I can't see an elegant way to introduce UnconstrainedContext.

The other option is to put Fit in package widget, unexport its Layout
method and use it as fields on Image and Icon to describe their
constrained behaviour.

WDYT?

Elias

egonelbre:
> The other option is to put Fit in package widget, unexport its Layout
> method and use it as fields on Image and Icon to describe their
> constrained behaviour.

This sounds good to me, since fit is going to be most useful for images. The corner cases can be handled by custom code.

eliasnaur:
Yet another option for a generally useful Fit would be to replace Layout with something lower level:

    // Scale adds clip and scale operations to fit dims to the constraints. It
    // returns dimensions modified accordingly.
    func (f Fit) Scale(gtx Context, dims Dimensions) Dimensions

forcing widgets to use Fit internally.

However, it's probably prudent to start with Fit internally to package widget.

egonelbre:
I do like that `scale` function version more, I'll change the api, but I'll keep it private regardless.

egonelbre:
One of the things I realized while testing was that even when we add Fit, the direction layout won't work as intended when using widget.Cover and layout.Center. I created a small program to experiment with the values: https://gist.github.com/egonelbre/b4fb1e812864f9ff16d58391fa0fd8ad.

This is the same issue about the unconstrained sizes. Since layout.Center doesn't know the size of the full size of the child, it cannot transform it to center it. Note, this issue actually already happens with "Unscaled" currently.

Maybe there needs to be an additional property on layout.Context that is "Unconstrained" and then widgets can draw according to that property? Then implementing the `Fit` as `layout.Fit` should be possible.

Note: I'm fine to holding off merging this until we have a good overall design.

widget/fit.go
```diff
@@ -0,0 +1,84 @@
...
+package widget
+
+import (
+	"gioui.org/f32"
+	"gioui.org/layout"
+	"gioui.org/op"
+	"gioui.org/op/clip"
+)
+
+// Fit scales a widget to fit to the constraints.
```
eliasnaur, line 12: Mention clipping in addition to scaling?

widget/fit.go
```diff
@@ -0,0 +1,84 @@
...
+type Fit uint8
+
+const (
+	// Unscaled does not alter the scale of a widget.
+	Unscaled Fit = iota
+	// Contain scales the widget to fit exactly constraints and
+	// preserves aspect-ratio.
```
eliasnaur, line 19: Can we make it clearer that "fitting" means making as large as possible? Here and below.

widget/fit.go
```diff
@@ -0,0 +1,84 @@
...
+	Contain
+	// Cover scales the widget to cover the constraints and
+	// preserves aspect-ratio.
+	Cover
+	// ScaleDown scales the widget down, when needed, to
+	// fit into the constraints and preserves aspect-ratio.
+	ScaleDown
+	// Cover scales the widget to fill the constraints and
```
eliasnaur, line 27: Cover=>Fill.

eliasnaur (commented):
> One of the things I realized while testing was that even when we add Fit, the direction layout won't work as intended when using widget.Cover and layout.Center. I created a small program to experiment with the values: https://gist.github.com/egonelbre/b4fb1e812864f9ff16d58391fa0fd8ad.
> 
> This is the same issue about the unconstrained sizes. Since layout.Center doesn't know the size of the full size of the child, it cannot transform it to center it. Note, this issue actually already happens with "Unscaled" currently.
> 

I'm probably being dense, but what exactly is the problem with Cover/Unscaled+Center? I ran your nice example program, but the behaviour seemed ok to me.

egonelbre:
> I'm probably being dense, but what exactly is the problem with Cover/Unscaled+Center? I ran your nice example program, but the behaviour seemed ok to me.

The issue is that the image is placed at the top, not Center or South as the direction would imply.

The Fit quite closely matches object-fit and background-position in CSS. Depending on the image you may want to "zoom into" the center area, the top area or bottom area -- or in an alternatively, cut off the top, bottom or both.

eliasnaur:
> > I'm probably being dense, but what exactly is the problem with Cover/Unscaled+Center? I ran your nice example program, but the behaviour seemed ok to me.
> 
> The issue is that the image is placed at the top, not Center or South as the direction would imply.

I see, thanks. Having layout.Direction and widget.Fit work together like CSS would be nice, but I think it's ok to say that it's the job of Fit to align the image as well as scaling and cropping it. FWIW, Android's ImageView also encode the alignment in its [scaleType](https://developer.android.com/reference/android/widget/ImageView#attr_android:scaleType) property.

egonelbre:
For me having them separate sounds clearer, such as

```
 type Image struct {
 	Src      paint.ImageOp
 	Fit      Fit
 	Position layout.Direction
 }
```

Something about the Android enums feels confusing.

eliasnaur:
> For me having them separate sounds clearer, such as
> 
> ```
>  type Image struct {
>  	Src      paint.ImageOp
>  	Fit      Fit
>  	Position layout.Direction
>  }
> ```
> 
> Something about the Android enums feels confusing.

Sure. I mentioned Android only to say that other GUIs also don't treat image positioning as part of their layout system.

widget/fit.go
egonelbre: Tried my best.

egonelbre:
I ended up adding layout.Direction to image as well to make the positioning possible. I ended up exposing some things from layout.Direction to avoid reimplementing the position calculation logic in Fit.

eliasnaur:
Looks good, merged.

PR #23, opened 2021-03-01
Pinch Gesture for OSX

wrnrlr:
More platfroms will come later


wrnrlr:
I don't understand why this is failing, thought it was the use of old api in gesture example, but fails to install gogio

PR #24, opened 2021-03-01
Pinch gesture for osx (cleanup)

wrnrlr:
title says it all

### DIFF
--- a/io/pointer/pointer.go
+++ b/io/pointer/pointer.go
@@ -36,6 +36,10 @@ type Event struct {
 	Position f32.Point
 	// Scroll is the scroll amount, if any.
 	Scroll f32.Point
+	// Scale is the rate of a pinch gesture
+	Magnification float32
+	// Rotation is the rate of a rotate gesture
+	Rotation float32
 	// Modifiers is the set of active modifiers when
 	// the mouse button was pressed.
 	Modifiers key.Modifiers
@@ -73,7 +77,7 @@ type PassOp struct {
 type ID uint16
 
 // Type of an Event.
-type Type uint8
+type Type uint16
 
 // Priority of an Event.
 type Priority uint8
@@ -127,6 +131,10 @@ const (
 	Leave
 	// Scroll of a pointer.
 	Scroll
+	// Pinch gesture
+	Pinch
+	// Rotate gesture
+	Rotate
 )
 
 const (
@@ -228,6 +236,10 @@ func (t Type) String() string {
 		return "Leave"
 	case Scroll:
 		return "Scroll"
+	case Pinch:
+		return "Pinch"
+	case Rotate:
+		return "Rotate"
 	default:
 		panic("unknown Type")
 	}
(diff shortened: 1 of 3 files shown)


io/pointer/pointer.go
```diff
@@ -73,7 +77,7 @@ type PassOp struct {
 type ID uint16
 
 // Type of an Event.
-type Type uint8
+type Type uint16
```
eliasnaur, line 80: What motivates this change?

io/pointer/pointer.go
```diff
@@ -36,6 +36,10 @@ type Event struct {
 	Position f32.Point
 	// Scroll is the scroll amount, if any.
 	Scroll f32.Point
+	// Scale is the rate of a pinch gesture
+	Magnification float32
+	// Rotation is the rate of a rotate gesture
+	Rotation float32
```
eliasnaur, line 42: I don't "rate" is right. According to the [documentation](https://developer.apple.com/documentation/appkit/nsevent/1526249-rotation) it seems rotation is a property of a tablet pen, and that it is in absolute degrees. If so, it seems more like a physical property (the angle of the pen) than a gesture (say, rotating a map with a two-finger gesture).

io/pointer/pointer.go
```diff
@@ -36,6 +36,10 @@ type Event struct {
 	Position f32.Point
 	// Scroll is the scroll amount, if any.
 	Scroll f32.Point
+	// Scale is the rate of a pinch gesture
+	Magnification float32
```
eliasnaur, line 40: "Rate" doesn't sound right. Better to say

// Magnification is the change of scale from a pinch gesture.

eliasnaur (changes requested):
I'm sceptical of this change. It seems better (albeit more work) to introduce gesture.Scale and gesture.Rotate similar to gesture.Drag. The advantages are portability, testability and better fit with Gio's pointer.Areas (for example, a pinch shouldn't register if the two fingers are in two different pointer.Areas).

eliasnaur:
> I'm sceptical of this change. It seems better (albeit more work) to introduce gesture.Scale and gesture.Rotate similar to gesture.Drag. The advantages are portability, testability and better fit with Gio's pointer.Areas (for example, a pinch shouldn't register if the two fingers are in two different pointer.Areas).

Oh, and we don't have to add the new fields and types to pointer.Event, along with rules for routing the new event types.

io/pointer/pointer.go
wrnrlr: With this change there are more then 8 pointer Types

wrnrlr:
True it is a lot of work. The osx implementation if particularly confusing because then we need to handle NSEvent and NSTouch together. It is not clear to me how existing multitouch events like the two finger scroll and click should work in gesture, while also supporting those events on pointer.

eliasnaur:
> True it is a lot of work. The osx implementation if particularly confusing because then we need to handle NSEvent and NSTouch together. It is not clear to me how existing multitouch events like the two finger scroll and click should work in gesture, while also supporting those events on pointer.

Two-finger scroll and (right?)-click cleanly map to events we have to handle from mice anyway, but it's true that it would be nice to have them in gesture instead of the low-level pointer interface. Pinch zoom and scroll, on the other hand, are more clearly gestures; they entail expanding pointer.Type and pointer.Event to support them, yet they're rarely used.

wrnrlr:
ok we will do it like that

PR #25, opened 2021-03-05
add pointer.Buttons to MouseClick events

mixmasala:
this commit extends ClickEvent to include the corresponding button.

### DIFF
--- a/gesture/gesture.go
+++ b/gesture/gesture.go
@@ -37,6 +37,8 @@ type Click struct {
 	clicks int
 	// pressed tracks whether the pointer is pressed.
 	pressed bool
+	// Buttons tracks which buttons are pressed.
+	Buttons pointer.Buttons
 	// entered tracks whether the pointer is inside the gesture.
 	entered bool
 	// pid is the pointer.ID.
@@ -53,6 +55,7 @@ type ClickEvent struct {
 	Position  f32.Point
 	Source    pointer.Source
 	Modifiers key.Modifiers
+	Buttons   pointer.Buttons
 	// NumClicks records successive clicks occurring
 	// within a short duration of each other.
 	NumClicks int
@@ -131,7 +134,7 @@ func (c *Click) Hovered() bool {
 	return c.entered
 }
 
-// Pressed returns whether a pointer is pressing.
+// Pressed returns whether a pointer is pressing the left mouse button.
 func (c *Click) Pressed() bool {
 	return c.pressed
 }
@@ -146,10 +149,12 @@ func (c *Click) Events(q event.Queue) []ClickEvent {
 		}
 		switch e.Type {
 		case pointer.Release:
-			if !c.pressed || c.pid != e.PointerID {
+			if c.pid != e.PointerID {
 				break
 			}
-			c.pressed = false
+			b := c.Buttons ^ e.Buttons // the released button
+			c.pressed = e.Buttons&pointer.ButtonPrimary > 0
+			c.Buttons = e.Buttons
 			if c.entered {
 				if e.Time-c.clickedAt < doubleClickDuration {
 					c.clicks++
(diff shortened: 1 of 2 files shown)


mixmasala:
I'm not sure why go fmt is failing

gesture/gesture.go
```diff
@@ -146,33 +149,34 @@ func (c *Click) Events(q event.Queue) []ClickEvent {
 		}
 		switch e.Type {
 		case pointer.Release:
-			if !c.pressed || c.pid != e.PointerID {
+			if c.pid != e.PointerID {
 				break
 			}
-			c.pressed = false
+			b := c.Buttons ^ e.Buttons // the released button
```
eliasnaur, line 155: Did you mean "buttons" in plural?

Use full comment sentences: // The released buttons.

gesture/gesture.go
```diff
@@ -146,33 +149,34 @@ func (c *Click) Events(q event.Queue) []ClickEvent {
...
+			c.pressed = e.Buttons&pointer.ButtonPrimary > 0
```
eliasnaur, line 156: Use "!=" for bitmask operations. A ">" looks like an ordered operation.

gesture/gesture.go
```diff
@@ -181,17 +185,17 @@ func (c *Click) Events(q event.Queue) []ClickEvent {
 			if c.pid != e.PointerID {
 				break
 			}
-			c.pressed = true
-			events = append(events, ClickEvent{Type: TypePress, Position: e.Position, Source: e.Source, Modifiers: e.Modifiers})
+			c.pressed = e.Buttons&pointer.ButtonPrimary > 0
```
eliasnaur, line 188: !=

gesture/gesture.go
```diff
@@ -37,6 +37,8 @@ type Click struct {
 	clicks int
 	// pressed tracks whether the pointer is pressed.
 	pressed bool
+	// Buttons tracks which buttons are pressed.
```
eliasnaur, line 40: This doesn't seem quite right. Did you mean

// Buttons specify the set of buttons to track. The zero value is equivalent to the
// primary button, not the empty set.

?

Ah, I see from the rest of your change that I wasn't clear enough in my description. What I meant by the Buttons field is that it should be set by the *user* from Click, not Click itself. The purpose of (the zero value of) Buttons is maintain the old behaviour where only the primary mouse button would be tracked. Users such as yourself that need to track other buttons can include them in Buttons.

gesture/gesture.go
```diff
@@ -37,6 +37,8 @@ type Click struct {
 	clicks int
 	// pressed tracks whether the pointer is pressed.
```
eliasnaur, line 38: Should this change to something like

// pressed tracks whether one or more buttons from the Buttons set is pressed.

?

gesture/gesture.go
```diff
@@ -131,7 +134,7 @@ func (c *Click) Hovered() bool {
 	return c.entered
 }
 
-// Pressed returns whether a pointer is pressing.
+// Pressed returns whether a pointer is pressing the left mouse button.
```
eliasnaur, line 137: "primary mouse button". Sorry for renaming the buttons while you're working on this change :)

gesture/gesture.go
```diff
@@ -146,33 +149,34 @@ func (c *Click) Events(q event.Queue) []ClickEvent {
...
 		switch e.Type {
 		case pointer.Release:
-			if !c.pressed || c.pid != e.PointerID {
+			if c.pid != e.PointerID {
 				break
 			}
-			c.pressed = false
+			b := c.Buttons ^ e.Buttons // the released button
+			c.pressed = e.Buttons&pointer.ButtonPrimary > 0
+			c.Buttons = e.Buttons
```
eliasnaur, line 157: Click shouldn't update its Buttons field. See my comment on the field above.

eliasnaur (changes requested):

PR #26, opened 2021-03-06
gpu/internal/rendertest: test depth buffer overlaps

egonelbre:
Regression test for 44991f355cc9938d02dc500062a20a8a61ac2cfa


eliasnaur:
Thank you very much!

PR #28, opened 2021-03-07
Radial Gradient

egonelbre:
Implementation of RadialGradient, I'm unsure whether we wanted to merge this or not. But I updated it, such that people can try and experiment if they want to.


eliasnaur (commented):
Indeed, I'd like to hold off features not implemented in the compute renderer.

FWIW, the rendering of materials have been split out to a separate (fixed function) pass, which may be easier to modify for gradients. See [renderMaterials](https://git.sr.ht/~eliasnaur/gio/tree/main/item/gpu/compute.go#L438).

eliasnaur:
> @eliasnaur I took a look at the `renderMaterials`, but doesn't this imply it'll create a texture for every gradient?
> 

A atlas reservation per gradient, but otherwise yes. I just pushed gioui.org/commit/3a3ec711d388a7 that caches materials off a (transformation, image handle) key. If two gradients are equal, they can re-use that space.

> For example in https://github.com/egonelbre/expgio/blob/master/box-shadows/surface.go, it currently creates 8*3 = 24 gradients for a single surface. Those gradients would add up really fast for anything more complex. While I understand that a specialized "round-rect-shadow" would reduce it to a single additional tex per atlas, it could still be problematic in other scenarios.

Indeed your example is expensive in texture memory, but I'll gladly settle for "it works, wasting memory" than not implemented at all.

Alternatives include: adding gradient support to piet-gpu (Raph is open to other material types); adding a "drop-shadow" material to Gio or effect to piet-gpu (we're likely going to need that eventually). Both options seem like more work, and I don't see myself working on either in the near future.

PR #29, opened 2021-03-13
op/clip: ensure that roundRect is always closed

egonelbre:
Floating point error may accumulate and the round rect may not
necessarily close up entirely. Add an additional "Close" to ensure
it's properly closed.


eliasnaur:
Thanks, merged.

PR #30, opened 2021-03-13
Add clip.Circle for drawing circles.

egonelbre:
Also update widget/material to use specific type.


egonelbre:
I noticed few other widgets using UniformRRect for drawing circles, I'll replace those as well.

eliasnaur:
Thanks, merged.

PR #31, opened 2021-03-13
Refactor clip.RRect

egonelbre:
Unfortunately, Github doesn't support separate stacked PR-s so the changes from https://github.com/gioui/gio/pull/30 are included. The only relevant change here is the last commit.

This uses absolute coordinates for RRect to avoid floating point drift. This is probably a nicer fix than adding a tiny segment to patch the issue.

Also, I realized that it's possible to remove clip.Border by exposing the Path of clip.RRect, e.g:

```
Stroke{
	Path: UniformRect(r, 3).Path(ops),
	Style: StrokeStyle{Width: 10},
}.Op()
```

Since this is a larger API change, I didn't do it.

### DIFF
--- a/op/clip/clip.go
+++ b/op/clip/clip.go
@@ -294,12 +294,15 @@ func (p *Path) arc(alpha float64, c f32.Point, rx, ry, beg, delta float64) {
 // Cube records a cubic Bézier from the pen through
 // two control points ending in to.
 func (p *Path) Cube(ctrl0, ctrl1, to f32.Point) {
-	if ctrl0 == (f32.Point{}) && ctrl1 == (f32.Point{}) && to == (f32.Point{}) {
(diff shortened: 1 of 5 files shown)


eliasnaur:
Thank you, merged.

> Unfortunately, Github doesn't support separate stacked PR-s so the changes from #30 are included. The only relevant change here is the last commit.
>

What I usually do is have a single PR with all my pending work, with self-contained commits sorted in increasing order of controversy.
 
> This uses absolute coordinates for RRect to avoid floating point drift. This is probably a nicer fix than adding a tiny segment to patch the issue.
> 
> Also, I realized that it's possible to remove clip.Border by exposing the Path of clip.RRect, e.g:
> 
> ```
> Stroke{
> 	Path: UniformRect(r, 3).Path(ops),
> 	Style: StrokeStyle{Width: 10},
> }.Op()
> ```
> 

LGTM.  Seems more composable your way.

PR #32, opened 2021-03-14
Expose Path for Circle, Rect and RRect

egonelbre:
I'm not sure whether clip.Border should be deprecated or not, but I added the commit just in case.

### DIFF
--- a/op/clip/shapes.go
+++ b/op/clip/shapes.go
@@ -51,11 +51,16 @@ type RRect struct {
 
 // Op returns the op for the rounded rectangle.
 func (rr RRect) Op(ops *op.Ops) Op {
-	return Outline{Path: rr.path(ops)}.Op()
+	return Outline{Path: rr.Path(ops)}.Op()
 }
 
-// path returns the path spec for the rounded rectangle.
-func (rr RRect) path(ops *op.Ops) PathSpec {
+// Add the rectangle clip.
+func (rr RRect) Add(ops *op.Ops) {
+	rr.Op(ops).Add(ops)
+}
+
+// Path returns the PathSpec for the rounded rectangle.
+func (rr RRect) Path(ops *op.Ops) PathSpec {
 	var p Path
 	p.Begin(ops)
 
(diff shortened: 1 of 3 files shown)


op/clip/shapes.go
```diff
@@ -26,6 +26,23 @@ func (r Rect) Add(ops *op.Ops) {
 	r.Op().Add(ops)
 }
 
+// Path returns the PathSpec for the rectangle.
+func (r Rect) Path(ops *op.Ops) PathSpec {
+	var p Path
```
eliasnaur, line 31: This results in a generic path clip, whereas Rect.Op records a very cheap pixel-aligned clip. One solution (other than leaving out Rect.Path altogether) is to move the bounds field from Op to Path, so that this Rect.Path method can simply return

```
PathSpec{
    bounds: image.Rectangle(r),
}
```

I'm going to need PathSpec bounds for all paths for compute anyway, so moving the field is probably going to happen regardless.

op/clip/shapes.go
```diff
@@ -91,12 +113,9 @@ func (rr RRect) path(ops *op.Ops) PathSpec {
...
 }
 
-// Add the rectangle clip.
-func (rr RRect) Add(ops *op.Ops) {
-	rr.Op(ops).Add(ops)
-}
-
 // Border represents a rectangular border.
+//
+// Deprecated: use clip.Stroke{Path: clip.UniformRRect(r, rr).Path(ops), Style: clip.StrokeStyle{Width: width}}.
```
eliasnaur, line 118: Please remove Border altogether. It is easy to replace, perhaps even with a `go fmt -r`.

op/clip/shapes.go
egonelbre: I'll remove it from this changeset for now.

I'm not yet quite sure how that change would work together with `clip.Stroke`.

op/clip/shapes.go
egonelbre: I'm not sure about the `go fmt -r` part since it doesn't seem to handle optional fields.

eliasnaur:
Thank you, merged.

PR #33, opened 2021-03-18
widget: ensure Border doesn't overlap content

egonelbre:
When using larger border width it would overlap the widget content.


egonelbre:
Yeah, I just realized that this change would be problematic, because if you add/remove or change the border width, this would move the internal content around.

PR #34, opened 2021-03-24
cmd/gogio: [android] add support for AAB

Inkeliz:
That patch makes possible to generate Android App Bundle (`.aab`) instead
of APK. In order to generate AAB use `-buildmode "bundle"`.


cmd/gogio/androidbuild.go
```diff
@@ -153,10 +154,15 @@ func buildAndroid(tmpDir string, bi *buildInfo) error {
 	case "archive":
 		return archiveAndroid(tmpDir, bi, perms)
 	case "exe":
-		if err := exeAndroid(tmpDir, tools, bi, extraJars, perms); err != nil {
+		if err := exeAndroid(tmpDir, tools, bi, extraJars, perms, false); err != nil {
 			return err
 		}
 		return signAPK(tmpDir, tools, bi)
+	case "bundle":
```
eliasnaur, line 161: Can we avoid another buildmode for what is arguably a variation of "exe"? How about choosing aab mode according to the output filename, so that `gogio -o blah.aab` will output in AAB format? There's precedent in that `-target ios` selects ".ipa" vs ".app" format according to output filename.

cmd/gogio/androidbuild.go
```diff
@@ -569,6 +593,94 @@ func signAPK(tmpDir string, tools *androidTools, bi *buildInfo) error {
...
+		return err
+	}
+
+	_, err = runCmd(exec.Command(
+		filepath.Join(tools.buildtools, "zipalign"),
+		"-f",
+		"4", // 32-bit alignment.
+		filepath.Join(tmpDir, "app.aa_"),
+		aabFile,
+	))
```
eliasnaur, line 639: Extract to function and share with signAPK.

cmd/gogio/androidbuild.go
```diff
@@ -569,6 +593,94 @@ func signAPK(tmpDir string, tools *androidTools, bi *buildInfo) error {
 	return nil
 }
 
+func signAAB(tmpDir string, tools *androidTools, bi *buildInfo) error {
+	aabFile := *destPath
+	if aabFile == "" {
+		dir, _ := os.Getwd()
+		aabFile = filepath.Join(dir, fmt.Sprintf("%s.aab", bi.name))
```
eliasnaur, line 600: Use filepath.Abs?

cmd/gogio/androidbuild.go
```diff
@@ -569,6 +593,94 @@ func signAPK(tmpDir string, tools *androidTools, bi *buildInfo) error {
...
+		"-keystore", bi.key,
+		"-list",
+		"-keypass", bi.password,
+		"-v",
+	))
+	if err != nil {
+		return err
+	}
+
+	matchAlias, err := regexp.Compile("Alias name:(.*)")
```
eliasnaur, line 661: This doesn't seem worth using the regexp package for. I suggest fmt.Sscanf or just strings.HasPrefix and slicing.

eliasnaur:
Thank you, merged.

PR #36, opened 2021-03-25
app: [js] move redraw out of js.FuncOf

Inkeliz:
Currently, the redraw is called inside js.FuncOf. That PR moves the
redraw to the main function, using channels inside the FuncOf, instead.

The current method (of calling inside the FuncOf) seems to be responsable
to generate `deadlock` errors. It happens even when the wrapped in
goroutines.

### DIFF
--- a/app/internal/wm/os_js.go
+++ b/app/internal/wm/os_js.go
@@ -77,7 +77,9 @@ func NewWindow(win Callbacks, opts *Options) error {
 	})
 	w.addEventListeners()
 	w.addHistory()
+	w.windowMode(opts.WindowMode)
 	w.w = win
+
 	go func() {
 		defer w.cleanup()
 		w.w.SetDriver(w)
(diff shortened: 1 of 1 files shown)


Inkeliz:
It's kind related to https://github.com/golang/go/issues/37136.  

My application freezes (and crashes) due to "all goroutines are asleep - deadlock!" on WASM. I tried to wrap my functions into `go func(){}` and channels. It fixes some problems, but sometimes it crashes due to "deadlock", at random.

Then, I notice that Gio seems to call the redraw inside the `js.FuncOf`. So, I think my functions are also called inside the `js.FuncOf`. _I'm not sure._

-----------

Maybe there's a better way to fix that problem. But using the channels and calling the `redraw` from the main function (outside of `js.FuncOf`) fixes my `goroutine` problems, for now. 😍

app/internal/wm/os_js.go
```diff
@@ -36,6 +36,9 @@ type window struct {
 	composing             bool
 	requestFocus          bool
 
+	chanAnimation chan bool
+	chanRedraw    chan bool
```
eliasnaur, line 40: I use `chan struct{}` everywhere else for notification channels. Please do that here as well.

eliasnaur:
Merged, thanks.

PR #37, opened 2021-04-03
app/internal/wm: [js] avoid duplicated-paste

Inkeliz:


widget/editor.go
```diff
@@ -426,6 +426,10 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
 		if k.Modifiers != key.ModShortcut {
 			return false
 		}
+		if runtime.GOOS == "js" {
+			// Prevents duplicated-paste. The HTMLInputElement, used as a input for JS, already handles CTRL + V.
+			return false
+		}
```
eliasnaur, line 432: It would be unfortunate to need platform specific code in otherwise portable widgets. Can we filter the duplicate pastes at a lower level?

Inkeliz: Well, the another alternative is using the `preventDefault` on [`Paste Event`](https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event). 

_The only caveat is that using `preventDefault` would require "clipboard-read" permission (because the Clipboard API would be responsable to read the content, instead of the InputElement)._

eliasnaur:
Thanks, merged.

PR #38, opened 2021-04-19
widget: make Editor lose focus when escape is pressed

vsariola:
Signed-off-by: vsariola <5684185+vsariola@users.noreply.github.com>

### DIFF
--- a/widget/editor.go
+++ b/widget/editor.go
@@ -444,6 +444,9 @@ func (e *Editor) command(gtx layout.Context, k key.Event) bool {
 			return false
 		}
 		e.caret.end, e.caret.start = e.offsetToScreenPos2(0, e.Len())
+	// Pressing escape removes focus from the editor
+	case key.NameEscape:
+		key.FocusOp{}.Add(gtx.Ops)
 	default:
 		return false
 	}


whereswaldon:
@vsariola Hi, thanks for the PR! We appreciate your desire to contribute. However, this feature is nonstandard, and I don't think it belongs in the core editor widget. Many text editors do not drop focus when escape is pressed, and implementing this here would prevent such usage.

Instead, I suggest that you try using the [`eventx`](https://pkg.go.dev/gioui.org/x/eventx#Spy) package to detect the escape keypress and drop focus from outside of the editor. That would look something like this (untested pseudocode):

```go
spy, spiedGtx := eventx.Enspy(gtx)
dims := editor.Layout(spiedGtx)
for _, group := range spy.AllEvents() {
    for _, event := range group.Items {
        switch e := event.(type) {
        case key.Event:
            if e.Name == key.NameEscape {
                key.FocusOp{}.Add(gtx.Ops)
            }
        }
    }
}
return dims
```

vsariola:
Got it; did not have any idea about the existence of EventSpy. Closing this as unnecessary.

BTW pretty good "pseudocode" as it worked right off the bat 👍 😄

whereswaldon:
I'm glad it worked!

PR #39, opened 2021-04-22
app, app/internal: [wasm,android] new Option to lock orientation

Inkeliz:


app/window.go
```diff
@@ -464,13 +464,31 @@ const (
...
 }
 
+const (
+	// UnlockedOrientation is the default behavior, allowing the rotate of the screen.
+	UnlockedOrientation = wm.UnlockedOrientation
+	// LockedPortrait locks the screen to portrait.
+	LockedPortrait = wm.LockedPortrait
+	// LockedLandscape locks the screen to landscape.
+	LockedLandscape = wm.LockedLandscape
+)
```
Inkeliz, line 481: I don't know if the names are right...

I choose `UnlockedOrientation` instead of `Any`, because `app.Any` seems wrong, since it doesn't means anything. Maybe `AnyOrientation` ? The `Locked` and `Unlocked` is to make obivous that the screen can/can't rotate.

Inkeliz: Every device (at least mobile devices) has "2x Portrait" and "2x Landscape". I mean, you can hold your smartphone at portrait mode or in a portrait mode upside down (180º rotated from the stardand portrait).

Maybe it could be extended to:

- UnlockedOrientation
- LockedPortrait               // Allow the rotation in 180º (primary or upside down; the sensor of the device can still rotate).
- LockedPrimaryPortrait   // Lock into the "primary" portrait.
- LockedReversedPortrait // Lock into the "reversed portrait (upside down).
- LockedLandscape
- LockedPrimaryLandscape
- LockedReversedLandscape

Currently the `LockedPortrait` and `LockedLandscape` behaves as `LockedPrimaryPortrait` and `LockedPrimaryLandscape`.

-----------

However, I think it's better to keep it simple and extend it once iOS (and maybe others OSes) gets supported. :)

app/internal/wm/window.go
```diff
@@ -37,6 +38,14 @@ const (
 	Fullscreen
 )
 
+type OrientationMode uint8
```
eliasnaur, line 41: Nit: "Orientation" is better.

app/window.go
eliasnaur: Thanks for working on this. I think "locked" is too similar to "lock screen". How about:

```
        // AnyOrientation allows the window to be freely orientated.
	AnyOrientation
	// PortraitOrientation constrains the window to portrait orientations.
	PortraitOrientation
	// LandscapeOrientation constrains the window to landscape orientations.
	LandscapeOrientation
```

app/internal/wm/os_android.go
```diff
@@ -754,6 +765,20 @@ func setCursor(env *C.JNIEnv, view C.jobject, name pointer.CursorName) {
...
 }
 
+func setOrientation(env *C.JNIEnv, view C.jobject, mode OrientationMode) {
+	var id int
+	// Constants defined at https://developer.android.com/reference/android/content/pm/ActivityInfo.
+	switch mode {
+	case UnlockedOrientation:
+		id = 2 // SCREEN_ORIENTATION_USER;
+	case LockedLandscape:
+		id = 0 // SCREEN_ORIENTATION_LANDSCAPE;
```
eliasnaur, line 775: If we're limiting the options to just landscape + portrait for now, let's use the variants that allow free movement, SCREEN_ORIENTATION_USER_LANDSCAPE and SCREEN_ORIENTATION_USER_PORTRAIT.

app/internal/wm/os_js.go
```diff
@@ -591,6 +598,21 @@ func (w *window) windowMode(mode WindowMode) {
...
+func (w *window) orientation(mode OrientationMode) {
+	if j := w.screenOrientation; !j.Truthy() || !j.Get("unlock").Truthy() || !j.Get("lock").Truthy() {
+		return // Browser don't support Screen Orientation API.
+	}
+
+	switch mode {
+	case UnlockedOrientation:
+		w.screenOrientation.Call("unlock")
+	case LockedLandscape:
+		w.screenOrientation.Call("lock", "landscape-primary").Call("then", w.redraw)
```
eliasnaur, line 610: I'm not familiar with the Screen Orientation API, but can we use values that allow both variants (normal and upside down) for each orientation, like for Android?

app/window.go
```diff
@@ -464,13 +464,31 @@ const (
...
+
+// Orientation sets the orientation of the app.
+//
+// Supported platforms are Android and JS (JS may requires to run on Fullscreen).
+func Orientation(mode wm.OrientationMode) Option {
```
eliasnaur, line 486: It occurred to me that a function for enum options is redundant. Why not make the enums themselves Options?

```
var (
    AnyOrientation = orientation(wm.AnyOrientation)
   LandscapeOrientation = orientation(wm.LandscapeOrientation)
   ...
)

func orientation(mode wm.Orientation) Option
```

See https://gioui.org/commit/a06a7a4b3c107 for a similar change to WindowMode.
The same applies to WindowMode.

app/internal/wm/os_js.go
Inkeliz: `landscape-primary` = `SCREEN_ORIENTATION_LANDSCAPE`.
`landscape` = `SCREEN_ORIENTATION_USER_LANDSCAPE`.

I update both: JS and Android. Now the user can rotate it. 🎉

app/internal/wm/os_android.go
Inkeliz: It's only supported by API Level 18. Should it fallback to SCREEN_ORIENTATION_USER if the device is using SDK 16? 😶

eliasnaur: Yes please, but fall back to SCREEN_ORIENTATION_LANDSCAPE and SCREEN_ORIENTATION_PORTRAIT which is API 1.

app/window.go
```diff
@@ -464,13 +464,31 @@ var (
...
+	AnyOrientation = orientation(wm.AnyOrientation)
+	// LandscapeOrientation constrains the window to landscape orientations.
+	LandscapeOrientation = orientation(wm.LandscapeOrientation)
+	// PortraitOrientation constrains the window to portrait orientations.
+	PortraitOrientation = orientation(wm.PortraitOrientation)
+)
+
+// orientation sets the orientation of the app.
+//
+// Supported platforms are Android and JS (JS may requires to run on Fullscreen).
```
eliasnaur, line 485: There's a stray word here: "may".

How about

// Supported on Android and some browsers.

?

eliasnaur:
Thanks! Merged.

PR #40, opened 2021-04-30
app, io: [wasm, android] add support for numeric/email keyboard mode

Inkeliz:
Previously, the on-screen-keyboard always displays the text keyboard,
QWERTY or equivalent. However, for a better user experience,
it is better for the keyboard to restrict, or facilitate, what
can be filled in the Editor.

Now, it's possible to set the `Mode` of the keyboard on `key.SoftKeyboardOp`.

Due to some limitations (gio#116), only numeric and text keyboards are
supported on Android.

### DIFF
--- a/app/internal/wm/GioView.java
+++ b/app/internal/wm/GioView.java
@@ -36,6 +36,7 @@ import android.view.inputmethod.BaseInputConnection;
 import android.view.inputmethod.InputConnection;
 import android.view.inputmethod.InputMethodManager;
 import android.view.inputmethod.EditorInfo;
+import android.text.InputType;
 
 import java.io.UnsupportedEncodingException;
 
@@ -47,6 +48,7 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 	private final InputMethodManager imm;
 	private final float scrollXScale;
 	private final float scrollYScale;
+	private int keyboardHint;
 
 	private long nhandle;
 
@@ -247,8 +249,18 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 		}
 	}
 
-	@Override public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
+	@Override public InputConnection onCreateInputConnection(EditorInfo editor) {
+		editor.inputType = this.keyboardHint;
+		editor.imeOptions = EditorInfo.IME_FLAG_NO_FULLSCREEN | EditorInfo.IME_FLAG_NO_EXTRACT_UI;
 		return new InputConnection(this);
+	}
+
+	void setInputHint(int hint) {
+	    if (hint == this.keyboardHint) {
+	        return;
+	    }
+	    this.keyboardHint = hint;
+	    imm.restartInput(this);
 	}
 
 	void showTextInput() {
--- a/app/internal/wm/os_android.go
+++ b/app/internal/wm/os_android.go
@@ -152,6 +152,7 @@ var gioView struct {
 	getFontScale       C.jmethodID
 	showTextInput      C.jmethodID
 	hideTextInput      C.jmethodID
+	setInputHint       C.jmethodID
 	postFrameCallback  C.jmethodID
 	setCursor          C.jmethodID
 	setOrientation     C.jmethodID
@@ -281,6 +282,7 @@ func Java_org_gioui_GioView_onCreateView(env *C.JNIEnv, class C.jclass, view C.j
 		m.getFontScale = getMethodID(env, class, "getFontScale", "()F")
 		m.showTextInput = getMethodID(env, class, "showTextInput", "()V")
 		m.hideTextInput = getMethodID(env, class, "hideTextInput", "()V")
+		m.setInputHint = getMethodID(env, class, "setInputHint", "(I)V")
 		m.postFrameCallback = getMethodID(env, class, "postFrameCallback", "()V")
 		m.setCursor = getMethodID(env, class, "setCursor", "(I)V")
 		m.setOrientation = getMethodID(env, class, "setOrientation", "(II)V")
@@ -606,6 +608,28 @@ func (w *window) ShowTextInput(show bool) {
 		} else {
 			callVoidMethod(env, w.view, gioView.hideTextInput)
 		}
+	})
+}
+
+func (w *window) SetInputHint(mode key.InputHint) {
+	// Constants defined at https://developer.android.com/reference/android/text/InputType.
+	const (
+		TYPE_NULL                = 0
+		TYPE_CLASS_NUMBER        = 2
+		TYPE_NUMBER_FLAG_DECIMAL = 8192
+		TYPE_NUMBER_FLAG_SIGNED  = 4096
+	)
+
+	runInJVM(javaVM(), func(env *C.JNIEnv) {
+		var m jvalue
+		switch mode {
+		case key.HintNumeric:
+			m = TYPE_CLASS_NUMBER | TYPE_NUMBER_FLAG_DECIMAL | TYPE_NUMBER_FLAG_SIGNED
+		default:
+			// TYPE_NULL, since TYPE_CLASS_TEXT isn't currently supported (gio#116), so TYPE_NULL is used instead.
+			m = TYPE_NULL
+		}
+		callVoidMethod(env, w.view, gioView.setInputHint, m)
 	})
 }
 
--- a/app/internal/wm/os_js.go
+++ b/app/internal/wm/os_js.go
@@ -313,6 +313,27 @@ func (w *window) focus() {
 	w.requestFocus = true
 }
 
+func (w *window) keyboard(hint key.InputHint) {
+	var m string
+	switch hint {
+	case key.HintAny:
+		m = "text"
+	case key.HintText:
+		m = "text"
+	case key.HintNumeric:
+		m = "decimal"
+	case key.HintEmail:
+		m = "email"
+	case key.HintURL:
+		m = "url"
+	case key.HintTelephone:
+		m = "tel"
+	default:
+		m = "text"
+	}
+	w.tarea.Set("inputMode", m)
+}
+
 func (w *window) keyEvent(e js.Value, ks key.State) {
 	k := e.Get("key").String()
 	if n, ok := translateKey(k); ok {
@@ -522,6 +543,10 @@ func (w *window) ShowTextInput(show bool) {
 			w.blur()
 		}
 	}()
+}
+
+func (w *window) SetInputHint(mode key.InputHint) {
+	w.keyboard(mode)
 }
 
 // Close the window. Not implemented for js.
--- a/app/window.go
+++ b/app/window.go
@@ -185,6 +185,9 @@ func (w *Window) processFrame(frameStart time.Time, size image.Point, frame *op.
 	case router.TextInputClose:
 		go w.Run(func() { w.driver.ShowTextInput(false) })
 	}
+	if hint, ok := w.queue.q.TextInputHint(); ok {
+		go w.Run(func() { w.driver.SetInputHint(hint) })
+	}
 	if txt, ok := w.queue.q.WriteClipboard(); ok {
 		go w.WriteClipboard(txt)
 	}
--- a/io/key/key.go
+++ b/io/key/key.go
@@ -22,7 +22,8 @@ import (
 // Key events are in general only delivered to the
 // focused key handler.
 type InputOp struct {
-	Tag event.Tag
+	Tag  event.Tag
+	Hint InputHint
 }
 
 // SoftKeyboardOp shows or hide the on-screen keyboard, if available.
@@ -63,6 +64,25 @@ type Event struct {
 type EditEvent struct {
 	Text string
 }
+
+// InputHint changes the on-screen-keyboard type. That hints the
+// type of data that might be entered by the user.
+type InputHint uint8
+
+const (
+	// HintAny hints that any input is expected.
+	HintAny InputHint = iota
+	// HintText hints that text input is expected. It may activate auto-correction and suggestions.
+	HintText
+	// HintNumeric hints that numeric input is expected. It may activate shortcuts for 0-9, "." and ",".
+	HintNumeric
+	// HintEmail hints that email input is expected. It may activate shortcuts for common email characters, such as "@" and ".com".
+	HintEmail
+	// HintURL hints that URL input is expected. It may activate shortcuts for common URL fragments such as "/" and ".com".
+	HintURL
+	// HintTelephone hints that telephone number input is expected. It may activate shortcuts for 0-9, "#" and "*".
+	HintTelephone
+)
 
 // State is the state of a key during an event.
 type State uint8
@@ -127,6 +147,7 @@ func (h InputOp) Add(o *op.Ops) {
 	}
 	data := o.Write1(opconst.TypeKeyInputLen, h.Tag)
 	data[0] = byte(opconst.TypeKeyInput)
+	data[1] = byte(h.Hint)
 }
 
 func (h SoftKeyboardOp) Add(o *op.Ops) {
--- a/io/router/key.go
+++ b/io/router/key.go
@@ -17,6 +17,7 @@ type keyQueue struct {
 	handlers map[event.Tag]*keyHandler
 	reader   ops.Reader
 	state    TextInputState
+	hint     key.InputHint
 }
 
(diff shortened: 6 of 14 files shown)


app/internal/wm/os_js.go
```diff
@@ -308,6 +308,25 @@ func (w *window) focus() {
...
+	case key.KeyboardEmail:
+		m = "email"
+	case key.KeyboardUrl:
+		m = "url"
+	case key.KeyboardTelephone:
+		m = "tel"
+	default:
+		m = "text"
+	}
+	w.tarea.Set("inputMode", m)
```
Inkeliz, line 327: `inputMode` isn't supported in some browsers (Firefox and old Safari). However, changing the `type` also affects the physical keyboard (not only the on-screen-keyboard).

It's possible to detect if `inputMode` isn't supported and use `type`. But, I think it's better to not change the `type` at all.

app/internal/wm/GioView.java
```diff
@@ -242,13 +245,20 @@ private void dispatchMotionEvent(MotionEvent event) {
...
+        return new InputConnection(this);
 	}
 
-	void showTextInput() {
-		GioView.this.requestFocus();
-		imm.showSoftInput(GioView.this, 0);
+	void showTextInput(int mode) {
+        this.keyboardMode = mode;
+        if (this.lastKeyboardMode != mode) {
+            imm.restartInput(this);
```
Inkeliz, line 257: _I didn't find anyway to change the type of the keyboard without restart it. The restart calls the `onCreateInputConnection` which sets the keyboard mode._

io/key/key.go
```diff
@@ -64,6 +65,23 @@ type EditEvent struct {
 	Text string
 }
 
+// VirtualKeyboardMode changes the on-screen-keyboard type. That hints the
+// type of data that might be entered by the user.
+type KeyboardMode uint8
```
Inkeliz, line 70: I don't know if the name is right. Maybe `KeyboardHint` or `KeyboardFlag`? 😶

eliasnaur: How about InputMode?

io/key/key.go
```diff
@@ -29,6 +29,7 @@ type InputOp struct {
 // It replaces any previous SoftKeyboardOp.
 type SoftKeyboardOp struct {
 	Show bool
+	Mode KeyboardMode
```
eliasnaur, line 32: What if an editor changes input mode but not focus? Input mode seems to belong to InputOp.

Inkeliz: The `KeyboardMode` only affects the on-screen-keyboard, which is only usefull when the `Show` is `true`. I actually think about replace the `Show: false` to `InvisibleKeyboard`. However, it would be a API change, so I avoid to change that.

----------

I don't know if it belongs to `InputOp` because it only affects the on-screen-keyboard. Who is responsable to show the virtual-keyboard is the `SoftKeyboardOp`. _That doesn't prevent the user to type anything on the physical keyboard._

eliasnaur: SoftKeyboardOp{Show:true} is not the only way to show the keyboard. The keyboard may appear on screen at all time, the Android Activity has a setting for displaying the keyboard when starting. Specifically, Android may call onCreateInputConnection at any time, but there may not have been a SoftKeyboardOp to update the InputMode.

Taking a step back, SoftKeyboardOp is an action op (show or hide the keyboard), whereas InputOp is a passive description of the widget. InputMode is a description of the widget, and therefore belongs in InputOp.

Inkeliz: I change it to `InputOp`. _If Android calls `onCreateInputConnection` at random, it will set the lastest mode._

app/internal/wm/GioView.java
```diff
@@ -242,13 +244,20 @@ private void dispatchMotionEvent(MotionEvent event) {
...
+	void setInputMode(int mode) {
+        this.keyboardMode = mode;
+        imm.restartInput(this);
 	}
 
 	void showTextInput() {
-		GioView.this.requestFocus();
-		imm.showSoftInput(GioView.this, 0);
+        GioView.this.requestFocus();
+        imm.showSoftInput(GioView.this, 0);
```
eliasnaur, line 260: Are you using spaces for indentation? I think we should leave it at tabs.

io/router/key.go
```diff
@@ -38,6 +40,24 @@ func (q *keyQueue) InputState() TextInputState {
 	return q.state
 }
 
+// InputState returns the last input mode state as
+// determined in Frame. It returns false if no change
+// happens.
+func (q *keyQueue) InputMode() (key.InputMode, bool) {
+	if q.focus == nil {
+		return key.KeyboardDefault, false
```
eliasnaur, line 48: I would expect `q.mode, false` here. See also my comment below for (*Router).TextInputMode.

io/router/key.go
```diff
@@ -38,6 +40,24 @@ func (q *keyQueue) InputState() TextInputState {
...
+	}
+	focused, ok := q.handlers[q.focus]
+	if !ok {
+		return key.KeyboardDefault, false
```
eliasnaur, line 52: Ditto.

io/router/key.go
```diff
@@ -38,6 +40,24 @@ func (q *keyQueue) InputState() TextInputState {
...
+	}
+	if focused.mode == q.mode {
+		return focused.mode, false
+	}
+	q.mode = focused.mode
+	return q.mode, true
```
eliasnaur, line 58: Nit:

```
old := q.mode
q.mode = focused.mode
return q.mode, old != q.mode
```

io/router/router.go
```diff
@@ -102,6 +102,12 @@ func (q *Router) TextInputState() TextInputState {
 	return q.kqueue.InputState()
 }
 
+// TextInputMode returns the input state from the most recent
+// call to Frame.
```
eliasnaur, line 106: Is it better to say

// TexInputMode returns the input mode from the most recent key.InputOp.

? 

Then there won't be a sudden switch to KeyboardDefault just because a Frame contains no Editors.

Remember to change the comment for (*keyQueue).InputMode as well.

io/key/key.go
```diff
@@ -64,6 +65,23 @@ type EditEvent struct {
...
+	// KeyboardDefault is the default keyboard (alphanumeric).
+	KeyboardDefault InputMode = iota
+	// KeyboardNumeric is an keyboard containing 0-9 Digits and decimal separator (`.` or `,`).
+	KeyboardNumeric
+	// KeyboardEmail is an keyboard containing "@", ".com" and other optimizations.
+	KeyboardEmail
+	// KeyboardUrl is an keyboard containing "/", ".com" and other optimizations.
+	KeyboardUrl
+	// KeyboardTelephone is an keyboard containing 0-9 Digits and the pound (#) and the asterisk (*).
+	KeyboardTelephone
```
eliasnaur, line 82: It's weird to have a common prefix `Keyboard` that doesn't match the type name. I suggest `Hint` (or `Mode` if we go with your original name)?

If so, we should change the comments to refer to the mode rather than a keyboard. For example:

// HintURL hints that URL input is expected. It may activate shortcuts for common URL fragments such as "/" and ".com".

io/key/key.go
```diff
@@ -64,6 +65,23 @@ type EditEvent struct {
...
+const (
+	// KeyboardDefault is the default keyboard (alphanumeric).
+	KeyboardDefault InputMode = iota
+	// KeyboardNumeric is an keyboard containing 0-9 Digits and decimal separator (`.` or `,`).
+	KeyboardNumeric
+	// KeyboardEmail is an keyboard containing "@", ".com" and other optimizations.
+	KeyboardEmail
+	// KeyboardUrl is an keyboard containing "/", ".com" and other optimizations.
+	KeyboardUrl
+	// KeyboardTelephone is an keyboard containing 0-9 Digits and the pound (#) and the asterisk (*).
```
eliasnaur, line 81: an => a.

io/key/key.go
```diff
@@ -64,6 +65,23 @@ type EditEvent struct {
 	Text string
 }
 
+// InputMode changes the on-screen-keyboard type. That hints the
+// type of data that might be entered by the user.
+type InputMode uint8
+
+const (
+	// KeyboardDefault is the default keyboard (alphanumeric).
+	KeyboardDefault InputMode = iota
```
eliasnaur, line 74: How about `HintAlphanumeric`, so that detail is moved from the comment to the name? If `alphanumeric` is too specific for you, I suggest `HintNone`.

Speaking of which, is there a need for both `HintNone` and `HintAlphanumeric`? The former represents no hints (TYPE_NULL in Android) while the latter explicitly asks for the alphanumeric input type.

io/key/key.go
```diff
@@ -64,6 +65,23 @@ type EditEvent struct {
...
+	// KeyboardNumeric is an keyboard containing 0-9 Digits and decimal separator (`.` or `,`).
+	KeyboardNumeric
+	// KeyboardEmail is an keyboard containing "@", ".com" and other optimizations.
+	KeyboardEmail
+	// KeyboardUrl is an keyboard containing "/", ".com" and other optimizations.
+	KeyboardUrl
```
eliasnaur, line 80: Url => URL is the Go style.

io/key/key.go
```diff
@@ -64,6 +65,23 @@ type EditEvent struct {
 	Text string
 }
 
+// InputMode changes the on-screen-keyboard type. That hints the
+// type of data that might be entered by the user.
+type InputMode uint8
```
eliasnaur, line 70: I think your "hint" is a better name than "mode", not least because the underlying platform may well ignore it. So how about `InputHint`?

io/key/key.go
Inkeliz: That makes sense. I also thought about others settings such as "autocapitalize", "grammar-correction" and "auto-complete", which could be disabled in some circumstances.

Since JS is the only OS that can support such feature, for now, maybe we can use as such:

- HintNone: Basic keyboard (without autocapitalize; without grammar-correction;  without autocomplete...). 
On Android it will use the TYPE_NULL.

- HintText: Alphanumeric keyboard (with all features). 
On Android it will use TYPE_TEXT (once it gets supported :S ).

app/internal/wm/os_android.go
```diff
@@ -579,6 +581,23 @@ func (w *window) ShowTextInput(show bool) {
...
+func (w *window) SetInputHint(mode key.InputHint) {
+	runOnMain(func(env *C.JNIEnv) {
+		if w.view == 0 {
+			return
+		}
+		var m jvalue
+		// Constants defined at https://developer.android.com/reference/android/text/InputType.
+		switch mode {
+		case key.HintNumeric:
+			m = 2 | 8192 | 4096 // TYPE_CLASS_NUMBER | TYPE_NUMBER_FLAG_DECIMAL | TYPE_NUMBER_FLAG_SIGNED
```
eliasnaur, line 593: Let's start declaring consts for the Java constants such as TYPE_*.

app/internal/wm/os_android.go
```diff
@@ -579,6 +581,23 @@ func (w *window) ShowTextInput(show bool) {
...
+		default:
+			m = 0 // TYPE_NULL, since TYPE_CLASS_TEXT isn't currently supported (gio#116), so TYPE_NULL is used instead.
```
eliasnaur, line 595: Constants.

io/key/key.go
```diff
@@ -64,6 +65,25 @@ type EditEvent struct {
...
 }
 
+// InputHint changes the on-screen-keyboard type. That hints the
+// type of data that might be entered by the user.
+type InputHint uint8
+
+const (
+	// HintAny hints that any input is expected.
+	HintAny InputHint = iota
+	// HintText hints that text input is expected. It may activate shortcut for grammar-corrector and suggestions.
```
eliasnaur, line 75: Nit: "It may activate auto-correction and suggestions"

io/key/key.go
```diff
@@ -64,6 +65,25 @@ type EditEvent struct {
...
+	HintText
+	// HintNumeric hints that numeric input is expected. It may activate shortcuts for 0-9 digits, "." and ",".
+	HintNumeric
+	// HintEmail hints that email input is expected. It may activate shortcut for common email characters, such as "@" and ".com".
```
eliasnaur, line 79: "shortcuts"

io/key/key.go
```diff
@@ -64,6 +65,25 @@ type EditEvent struct {
...
+	HintEmail
+	// HintURL hints that URL input is expected. It may activate shortcuts for common URL fragments such as "/" and ".com".
+	HintURL
+	// HintTelephone hints that telephone number input is expected. It may activate shortcuts for 0-9 digits, "#" and "*".
```
eliasnaur, line 83: Nit: drop "digits".

io/router/key.go
```diff
@@ -17,13 +17,15 @@ type keyQueue struct {
 	handlers map[event.Tag]*keyHandler
 	reader   ops.Reader
 	state    TextInputState
+	mode     key.InputHint
```
eliasnaur, line 20: Rename to "hint".

io/router/key.go
```diff
@@ -17,13 +17,15 @@ type keyQueue struct {
...
 }
 
 type keyHandler struct {
 	// visible will be true if the InputOp is present
 	// in the current frame.
 	visible bool
 	new     bool
+	mode    key.InputHint
```
eliasnaur, line 28: Rename to "hint".

Inkeliz:
The build seems to fail, but it's not related to that change. :|

whereswaldon:
Looks like we need to update to depend on Android 30, which is the effective minimum in the play store. [This type](https://developer.android.com/reference/android/view/WindowInsetsController) was only introduced in Android API Level 30.

whereswaldon:
Ah, Elias has already done the update [here](https://git.sr.ht/~eliasnaur/gio/commit/a7be7cc22bd027fcbad357ee5573bf512750f3aa). Can you rebase your PR on top of the latest main branch?

app/internal/wm/GioView.java
```diff
@@ -247,10 +249,20 @@ private void dispatchMotionEvent(MotionEvent event) {
...
+		editor.imeOptions = EditorInfo.IME_FLAG_NO_FULLSCREEN | EditorInfo.IME_FLAG_NO_EXTRACT_UI;
 		return new InputConnection(this);
 	}
 
+	void setInputHint(int hint) {
+	    if (hint == this.keyboardHint) {
+	        return;
+	    }
+		this.keyboardHint = hint;
+		imm.restartInput(this);
```
eliasnaur, line 263: The indentation differs in this section.

app/window.go
```diff
@@ -185,6 +185,9 @@ func (w *Window) processFrame(frameStart time.Time, size image.Point, frame *op.
 	case router.TextInputClose:
 		go w.Run(func() { w.driver.ShowTextInput(false) })
 	}
+	if hint, ok := w.queue.q.TextInputHint(); ok {
+		w.driver.SetInputHint(hint)
```
eliasnaur, line 189: Use a goroutine and Window.Run like for ShowTextInput above. See also https://gioui.org/commit/8611894b4bb36629b1255ee30989d38438274fe9.

app/internal/wm/os_js.go
```diff
@@ -524,6 +545,12 @@ func (w *window) ShowTextInput(show bool) {
 	}()
 }
 
+func (w *window) SetInputHint(mode key.InputHint) {
+	go func() {
+		w.keyboard(mode)
+	}()
```
eliasnaur, line 551: SetInputHint should be called from in Window.Run and you don't need a goroutine here anymore.

app/internal/wm/os_android.go
```diff
@@ -609,6 +611,31 @@ func (w *window) ShowTextInput(show bool) {
...
+func (w *window) SetInputHint(mode key.InputHint) {
+	// Constants defined at https://developer.android.com/reference/android/text/InputType.
+	const (
+		TYPE_NULL                = 0
+		TYPE_CLASS_NUMBER        = 2
+		TYPE_NUMBER_FLAG_DECIMAL = 8192
+		TYPE_NUMBER_FLAG_SIGNED  = 4096
+	)
+
+	runOnMain(func(env *C.JNIEnv) {
```
eliasnaur, line 623: SetInputInput should be called through Window.Run and you can assume the main thread here. See window.ShowTextInput in this file.

eliasnaur:
Looks good, merged. Thank you for your patience.

PR #42, opened 2021-06-07
internal/stroke, op/clip: remove unused variables

Inkeliz:


eliasnaur:
Add "Fixes gio#xxx" to commit message.

eliasnaur:
Merged, thanks.

PR #46, opened 2021-06-07
internal/wm: [wasm] add Title support

Inkeliz:


app/internal/wm/os_js.go
```diff
@@ -602,6 +605,10 @@ func (w *window) config() (int, int, system.Insets, unit.Metric) {
 		}
 }
 
+func (w *window) title(s string) {
+	w.document.Set("title", s)
+}
+
```
eliasnaur, line 611: Is it worth adding a method for?

eliasnaur:
Thanks, merged.

PR #47, opened 2021-06-08
app/internal/wm: [android] fix key.Event and key.EditEvent conflict

Inkeliz:
Fixes gio#224

### DIFF
--- a/app/internal/wm/os_android.go
+++ b/app/internal/wm/os_android.go
@@ -546,7 +546,7 @@ func Java_org_gioui_GioView_onKeyEvent(env *C.JNIEnv, class C.jclass, handle C.j
 	if n, ok := convertKeyCode(keyCode); ok {
 		w.callbacks.Event(key.Event{Name: n})
 	}
-	if r != 0 {
+	if r != 0 && r != '\n' { // Checking for "\n" to prevent duplication with key.NameEnter (gio#224).
 		w.callbacks.Event(key.EditEvent{Text: string(rune(r))})
 	}
 }


app/internal/wm/os_android.go
```diff
@@ -545,6 +545,7 @@ func Java_org_gioui_GioView_onKeyEvent(env *C.JNIEnv, class C.jclass, handle C.j
 	w := views[handle]
 	if n, ok := convertKeyCode(keyCode); ok {
 		w.callbacks.Event(key.Event{Name: n})
+		return
```
eliasnaur, line 548: You're trying to filter out the EditText{"\n"}, right? If so, I think it's safer to just add that condition to the "if r!= 0" check.

Inkeliz: Yes, since the `\n` is sent from both key.Event and key.EditEvent. 

 thought about the `if` stuff. But is there any reason to use `key.EditEvent` and `key.Event` at the same time?  I mean, on Windows it never happens, even under that situation:
https://github.com/gioui/gio/blob/910fa30edf05a6a402a7036e551fc9e442a4972c/app/internal/wm/os_windows.go#L682-L685
https://github.com/gioui/gio/blob/910fa30edf05a6a402a7036e551fc9e442a4972c/app/internal/wm/os_windows.go#L240-L259

Even in that case it only creates the `key.Event` but doesn't create the `key.EditEvent`.

eliasnaur: I'd expect both events for, say, the 'A' key.

Inkeliz: So, do you think that it should check `if r != 0 && string(r) == "\n" {}`? Or, should we change the Editor itself, and detects "\n" on EditKey (when `Submit` is enabled)?

eliasnaur: Check r != '\n' and add a comment saying why we check it.

eliasnaur:
Thanks. Merged.

PR #49, opened 2021-07-06
add centered option for windows OS

audrenbdb:


app/window.go
```diff
@@ -596,6 +596,13 @@ func Title(t string) Option {
 	}
 }
 
+// Centered set the centered position of the VM
+func Centered(c bool) Option {
```
Inkeliz, line 600: I think it could be `var Centered` ? 

Similar of `var Windowed` or `var AnyOrientation`:

https://github.com/gioui/gio/blob/d5f8c860af591025fc1e27174a2e31400492cf1b/app/window.go#L558-L590

app/internal/wm/os_windows.go
```diff
@@ -555,6 +534,49 @@ func (w *window) Option(opts *Options) {
 	}
 }
 
+//SetWindowLayout set size and position of window
```
eliasnaur, line 537: Nit: missing space after //.

app/window.go
eliasnaur: What about top-left or just left? I suggest adding a Direction type like layout.Direction and add the option

// Gravity controls the direction a window is placed towards.
func Gravity(dir layout.Direction)

Note in the comments that only Windows is supported.

app/internal/wm/os_windows.go
```diff
@@ -555,6 +534,49 @@ func (w *window) Option(opts *Options) {
...
+	windows.MoveWindow(w.hwnd, 0, 0, width, height, true)
+}
+
+func (w *window) CenterWindow(width, height int32) {
+	screenWidth := int32(windows.GetSystemMetrics(w.hwnd, windows.SM_CXSCREEN))
+	screenHeight := int32(windows.GetSystemMetrics(w.hwnd, windows.SM_CYSCREEN))
+	x := screenWidth/2 - width/2
+	y := screenHeight/2 - height/2
+	windows.MoveWindow(w.hwnd, x, y, width, height, true)
+}
```
eliasnaur, line 578: Inline these functions into SetWindowLayout; they're not worth splitting out. If you do want to split them, at least move the MoveWindow call to SetWindowLayout and separate only the width/height computations.

eliasnaur:
I believe this functionality have been implemented in Gio.

PR #51, opened 2021-07-26
app: [android] destroy gpu when change view

Inkeliz:
That patch restores the old behavior for Android, and fixes
the gio#248.

------------

I don't think that code is good, and may have a nicer way of fixing it. Feel free to change the code (and refuse this patch). I'm just sending it here because it fixes the [gio#248](https://todo.sr.ht/~eliasnaur/gio/248). 👍

### DIFF
--- a/app/window.go
+++ b/app/window.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"image"
 	"image/color"
+	"runtime"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Inkeliz:
I'm closing in favor of https://git.sr.ht/~eliasnaur/gio/commit/0dc99ab.

PR #54, opened 2021-08-28
[WASM] Fix PACK_ROW_LENGTH, UNPACK_ROW_LENGTH for WebGL 1

Inkeliz:
Fix gio#259

### DIFF
--- a/internal/gl/gl_js.go
+++ b/internal/gl/gl_js.go
@@ -262,6 +262,12 @@ func (f *Functions) GetBindingi(pname Enum, idx int) Object {
 	return Object(obj)
 }
 func (f *Functions) GetInteger(pname Enum) int {
+	if !f.isWebGL2 {
+		switch pname {
+		case PACK_ROW_LENGTH, UNPACK_ROW_LENGTH:
+			return 0 // PACK_ROW_LENGTH and UNPACK_ROW_LENGTH is only available on WebGL 2
+		}
+	}
 	return paramVal(f.Ctx.Call("getParameter", int(pname)))
 }
 func (f *Functions) GetFloat(pname Enum) float32 {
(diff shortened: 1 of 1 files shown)


internal/gl/gl_js.go
```diff
@@ -351,6 +357,12 @@ func (f *Functions) LinkProgram(p Program) {
 	f.Ctx.Call("linkProgram", js.Value(p))
 }
 func (f *Functions) PixelStorei(pname Enum, param int) {
+	if !f.isWebGL2 {
+		switch pname {
+		case PACK_ROW_LENGTH, UNPACK_ROW_LENGTH:
+			return // PACK_ROW_LENGTH and UNPACK_ROW_LENGTH is only available on WebGL 2
```
eliasnaur, line 363: We can't just ignore the PixelStorei call. Panic?

Inkeliz: Technically, the `PixelStorei` is supported on WebGL 1. I don't knwo if `PixelStorei` is called with other  types of `pname`. I'll test it later. :)

eliasnaur: Both PACK_ROW_LENGTH and UNPACK_ROW_LENGTH are unsupported in webgl 1 according to https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/pixelStorei. Not sure why PACK... worked until now. In any case, you can probably get away with skipping the pixelstorei call if the row length is equal to the image stride.

internal/gl/gl_js.go
```diff
@@ -351,6 +357,9 @@ func (f *Functions) LinkProgram(p Program) {
 	f.Ctx.Call("linkProgram", js.Value(p))
 }
 func (f *Functions) PixelStorei(pname Enum, param int) {
+	if !f.isWebGL2 {
+		return // PACK_ROW_LENGTH and UNPACK_ROW_LENGTH is only available on WebGL 2
```
eliasnaur, line 361: You're still ignoring PixelStorei which may corrupt texture downloads and uploads. How about skipping the call to PixelStore if row length == image width? Or, replace value with 0 if so.

eliasnaur:
Superseded by https://gioui.org/commit/f7975024735b0413dad635f5d23987380378676b.

PR #55, opened 2021-09-19
app: [wasm] fix crash when send events

Inkeliz:
The Configure function must be called after the SetDriver.

Fix gio#269


eliasnaur:
Thanks. I squashed the fix with my change.

PR #56, opened 2021-09-19
app: [wasm] fix scale

Inkeliz:


eliasnaur:
Thanks. Squashed along with a fix for sending ConfigEvent on resize.

PR #57, opened 2021-09-27
internal/gl: [js] adds cache to GetInteger/GetFloat and bind functions

Inkeliz:
In the beginning of each frame, multiple calls to
`GetInteger`/`GetFloat` is performed, taking around ~2ms out of ~8ms.

Now, those data are cache at first usage, avoiding calls to JS. It saves
around ~2ms out of ~8ms.

Due to `syscall/js` usage of `TextEncoder`, which is slow on Chrome, a
new cache was created, to bind function and use `Invoke` instead. It
saves around ~1ms out of ~8ms.

In total, this patch improves the performance from ~8ms to ~5ms. It
still slower than https://github.com/gioui/gio/pull/4.

### DIFF
--- a/gpu/internal/opengl/opengl.go
+++ b/gpu/internal/opengl/opengl.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"image"
+	"runtime"
 	"strings"
 	"time"
 	"unsafe"
@@ -78,6 +79,7 @@ type glState struct {
 	viewport          [4]int
 	unpack_row_length int
 	pack_row_length   int
+	isInitized        bool
 }
 
 type state struct {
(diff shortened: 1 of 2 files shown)


eliasnaur:
The reason we ask for those values at each frame is to restore the gl state at the end of a frame. This is important when mixing gio and foreign rendering.

This change defeats that.

Inkeliz:
It's possible to mix gio and another renderer on WebAssembly?

If so, I can change it to only use the `.Call("bind")` to avoid the TextEncoder, saving some ~1ms. :\

eliasnaur:
> It's possible to mix gio and another renderer on WebAssembly?
> 

Not that I know of.

> If so, I can change it to only use the `.Call("bind")` to avoid the TextEncoder, saving some ~1ms. :\

How about skipping state save/load when GOOS=="js"? That's just as fast, and clearer than returning cached values from Get* functions.

eliasnaur:
Superseded by https://gioui.org/commit/30ecf75a.

Inkeliz:
Thanks. 

It doesn't include the "Bind"/"Invoke" for JavaScript. 

Should I submit a patch for that, or do you don't like the idea of Invoke?

The bind will save the function and the invoke use the saved function. 

The saved function doesn't include the arguments, so it calls the function as normally would. The difference is that you don't need to translate the string (from the function name) to JS string. 

In my test (on Kitchen) it save 1ms (from 8ms to 7ms). It may save more time on mobile devices. There's no side effect.

eliasnaur:
Sure, bind/invoke sounds good in a separate PR. Or re-open this one with a different subject.

PR #59, opened 2021-10-03
cmd/gogio: compile when no icon is found

Inkeliz:
If the icon isn't found, it will continue to build the program, without the icon.


eliasnaur:
Thanks!

PR #60, opened 2021-10-06
app: [wasm] fix deadlock when send events

inkeliz:


eliasnaur (commented):
Add

Fixes gio#283.

Using goroutines to work around what I believe is a Go/WASM limitation is inefficient but also introduces non-determinism: with goroutines you can no longer be sure events delivered in order are processed in order.

A deterministic solution would be to queue up events in callbacks and then schedule the window event loop to process them.

Inkeliz:
I think that is a major issue for JS, for long time. Could you implement the event queue or accept the goroutine? It needs rebase, anyway. 😬😬

eliasnaur:
I don't think the goroutine approach is the right fix, but I'm happy to accept a patch that adds incoming events to a queue to be processed by the event loop.

inkeliz:
I'm updating my app to the latest Gio, so I'm taking a look on some changes that I have in my forked version. That deadlock-fix is one of them.

However, "adds incoming events to a queue to be processed by the event loop", will not cause another deadlock if the queue is full? Or did you mean some "slice-based" queue, instead of channels?

eliasnaur:
>  Or did you mean some "slice-based" queue, instead of channels?

Yes, I meant a "slice-based" queue.

PR #61, opened 2021-10-15
app: [android] handle touch events from stylus

dstaley:
This PR treats touch events from a stylus as if they were from a finger. I'm sure there's a philosophical debate as to whether a stylus is more like a mouse pointer than a finger, but I figured this would at least be a good first step to getting apps built with Gio to work mostly correctly with a stylus on Android devices.

Tested with the widgets in the `kitchen` example.


eliasnaur:
Merged. Thank you!

PR #64, opened 2021-10-26
app,internal/gl: [wasm] fix context lost

inkeliz:


gpu/internal/opengl/opengl.go
```diff
@@ -685,10 +685,6 @@ func (t *texture) ensureFBO() gl.Framebuffer {
 		panic(err)
 	}
 	b.funcs.FramebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, t.obj, 0)
-	if st := b.funcs.CheckFramebufferStatus(gl.FRAMEBUFFER); st != gl.FRAMEBUFFER_COMPLETE {
-		b.funcs.DeleteFramebuffer(fb)
-		panic(fmt.Errorf("incomplete framebuffer, status = 0x%x, err = %d", st, b.funcs.GetError()))
-	}
```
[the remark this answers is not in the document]
Inkeliz: I change the `gl` function, to return `COMPLETE` if the current context is lost:

https://github.com/gioui/gio/pull/64/commits/8d099023d8ff52a48e7ac6719b28a1be88d2005d#diff-0f89f88088151ae0c3098bf34cc6c2af613a37bb1e72ce62ed6e41177c21a0e6R121-R126

app/os_js.go
```diff
@@ -162,6 +165,22 @@ func (w *window) cleanup() {
 }
 
 func (w *window) addEventListeners() {
+	w.addEventListener(w.cnv, "webglcontextlost", func(this js.Value, args []js.Value) interface{} {
+		args[0].Call("preventDefault")
+		w.contextLost = true
+		w.cnv.Set("width", 0)
+		w.cnv.Set("height", 0)
```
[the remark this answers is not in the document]
Inkeliz: I added some comment and also move the resize to `webglcontextrestored`:

https://github.com/gioui/gio/commit/8d099023d8ff52a48e7ac6719b28a1be88d2005d#diff-3948540b9786c362b9acf669b24a166ef3bebb8ef23d1a1b0a7892c84e293d6aR181-R190

app/os_js.go
```diff
@@ -54,6 +54,9 @@ type window struct {
 	// is pending.
 	animRequested bool
 	wakeups       chan struct{}
+
+	contextLost          bool
+	contextLostRecovered bool
```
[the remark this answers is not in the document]
Inkeliz: I put some new `Enums` and comments.

app/os_js.go
```diff
@@ -24,6 +24,14 @@ import (
 
 type ViewEvent struct{}
 
+type contextStatus uint8
```
eliasnaur, line 27: FWIW, using `uint8` for enums is (my) bad idea. A plain `int` is clearer and just as performant in most cases.

app/gl_js.go
```diff
@@ -65,5 +74,8 @@ func (c *glContext) Refresh() error {
 }
 
 func (w *window) NewContext() (context, error) {
+	if w.contextStatus != contextStatusOkay {
+		return nil, gpu.ErrDeviceLost
```
eliasnaur, line 78: Why can NewContext return ErrDeviceLost? NewContext is only ever called as a result of a FrameEvent, which is sent by the driver. In other words, the driver should suppress FrameEvents while the context is lost.

app/window.go
```diff
@@ -192,6 +192,10 @@ func (w *Window) validateAndProcess(frameStart time.Time, size image.Point, sync
 					w.ctx, err = d.NewContext()
 				})
 				if err != nil {
+					if errors.Is(err, gpu.ErrDeviceLost) {
```
eliasnaur, line 195: This change shouldn't be necessary if NewContext never returns ErrDeviceLost, as commented above.

eliasnaur (commented):
Much nicer, thank you.

app/gl_js.go
Inkeliz: Removing it doesn't work, or will require more changes. Seems that new frames is generated as response from clicks or other events.

eliasnaur: app.Window only renders in response to FrameEvents, and os_js.go is the only source of those. Can't you check the context status before invoking requestAnimationFrame, and before sending a FrameEvent in draw? You also need to restart the animation loop when the context recovers.

internal/gl/gl_js.go
```diff
@@ -299,7 +299,12 @@ func (f *Functions) BufferSubData(target Enum, offset int, src []byte) {
 	f._bufferSubData.Invoke(int(target), offset, f.byteArrayOf(src))
 }
 func (f *Functions) CheckFramebufferStatus(target Enum) Enum {
-	return Enum(f._checkFramebufferStatus.Invoke(int(target)).Int())
+	status := Enum(f._checkFramebufferStatus.Invoke(int(target)).Int())
+	if status != FRAMEBUFFER_COMPLETE && f.Ctx.Call("isContextLost").Bool() {
```
eliasnaur, line 303: Nit: The status check is redundant. Checking `isContextLost` is enough.

app/os_js.go
```diff
@@ -162,6 +176,24 @@ func (w *window) cleanup() {
 }
 
 func (w *window) addEventListeners() {
+	w.addEventListener(w.cnv, "webglcontextlost", func(this js.Value, args []js.Value) interface{} {
+		args[0].Call("preventDefault")
+		w.contextStatus = contextStatusLost
+		w.SetAnimating(false)
```
eliasnaur, line 182: app.Window is in control of SetAnimating; it seems inappropriate to force it here.

app/os_js.go
```diff
@@ -162,6 +176,24 @@ func (w *window) cleanup() {
...
+		w.w.Event(system.StageEvent{Stage: system.StagePaused})
+		return nil
+	})
+	w.addEventListener(w.cnv, "webglcontextrestored", func(this js.Value, args []js.Value) interface{} {
+		w.contextStatus = contextStatusRestored
+		w.SetAnimating(true)
```
eliasnaur, line 188: Ditto.

app/gl_js.go
eliasnaur: I don't think this is resolved yet; NewContext can still return ErrDeviceLost.

app/os_js.go
```diff
@@ -162,6 +176,24 @@ func (w *window) cleanup() {
...
+		// Resize is required to force update the canvas content when restored.
+		w.cnv.Set("width", 0)
+		w.cnv.Set("height", 0)
+		w.resize()
+		w.w.Event(system.StageEvent{Stage: system.StageRunning})
+		w.chanRedraw <- struct{}{}
```
eliasnaur, line 194: This is unrelated to this change, but I worry that these unconditional channel sends risks deadlock. Why not

```
select {
case w.chanRedraw <- struct{}{}
default:
}
```

for chanRedraw and chanAnimation?

internal/gl/gl_js.go
eliasnaur: Good point. Disregard my comment.

app/gl_js.go
inkeliz: That was removed. However, the code is a little bit more complex. :S

app/gl_js.go
```diff
@@ -32,11 +33,15 @@ func newContext(w *window) (*glContext, error) {
...
 		cnv: w.cnv,
+		w:   w,
 	}
 	return c, nil
 }
 
 func (c *glContext) RenderTarget() (gpu.RenderTarget, error) {
+	if c.w.contextStatus != contextStatusOkay {
+		return nil, gpu.ErrDeviceLost
+	}
```
inkeliz, line 44: That is required to prevent new frames to be generated with the old content, because was already lost.

app/gl_js.go
```diff
@@ -61,7 +63,21 @@ func (c *glContext) Lock() error {
...
+	if c.w.contextStatus&contextStatusDestroyed == 0 {
+		c.w.contextStatus |= contextStatusDestroyed
+		return gpu.ErrDeviceLost
+	}
+	if c.w.contextStatus&contextStatusRestored != 0 {
+		c.w.contextStatus = contextStatusOkay
+		return nil
+	}
+	return gpu.ErrDeviceLost
```
inkeliz, line 80: That is chaotic, but the reason is: I decide to not change the `window.go` file, or add new errors, which would be exclusively for WASM.

--------

The `Refresh` is called when the context is already lost and a new GPU is created. Because of that, we have two issues:

1. It should not continue if the Status is Lost (that is unlike, since new frames are not generated).
2. It should destroy the current context (force Gio to recreate everything).
3. It should continue if the context is already destroyed (and then will be re-created).

app/os_js.go
```diff
@@ -178,14 +209,12 @@ func (w *window) addEventListeners() {
...
-		ev := system.StageEvent{}
 		switch w.document.Get("visibilityState").String() {
 		case "hidden", "prerender", "unloaded":
-			ev.Stage = system.StagePaused
+			w.w.Event(system.StageEvent{Stage: system.StagePaused})
 		default:
-			ev.Stage = system.StageRunning
+			w.w.Event(system.StageEvent{Stage: system.StageRunning})
 		}
-		w.w.Event(ev)
```
eliasnaur, line 188: Unrelated change?

app/os_js.go
eliasnaur: I don't think this is resolved. You did implement non-blocking redraw for the `resize` event below, so maybe add a new method for requesting redraw?

app/os_js.go
```diff
@@ -24,6 +24,15 @@ import (
...
 type ViewEvent struct{}
 
+type contextStatus int
+
+const (
+	contextStatusOkay contextStatus = 0
+	contextStatusLost contextStatus = 1 << (iota + 1)
+	contextStatusRestored
+	contextStatusDestroyed
+)
```
eliasnaur, line 34: Do you ever need more than one context status at the same time? I don't see anywhere. If not, convert to regular enum and avoid bit set operations.

inkeliz: On `func (c *glContext) Refresh() error` it checks if it's destroyed and it's restored or not. So, it gives `gpu.ErrDeviceLost` and flags as `Destroyed`. Then, when restored, it will check if it's `Restored` and `Destroyed`, otherwise it will destroy first.

app/gl_js.go
eliasnaur: I don't see why you need contextStatusDestroyed. How about

```
	switch c.w.contextStatus {
	case contextStatusLost:
		return errOutOfDate
	case contextStatusRestored:
		c.w.contextStatus = contextStatusOkay
		return gpu.ErrDeviceLost
	default:
		return nil
	}
```

? If so, I believe you can demote context status from a bitfield to a state enum.

app/os_js.go
eliasnaur: Indeed, I had missed the bit manipulations happening in Refresh. See my other comment to why I still believe a state enum is enough.

eliasnaur:
Merged! Thank you for your patience :)

PR #65, opened 2021-11-05
app: [iOS] add ViewEvent

Inkeliz:
Fix #305

### DIFF
--- a/app/os_ios.go
+++ b/app/os_ios.go
@@ -86,7 +86,10 @@ import (
 	"gioui.org/unit"
 )
 
-type ViewEvent struct{}
+type ViewEvent struct {
+	// View is a CFTypeRef for the UIView of GioView.
+	View uintptr
+}
 
 type window struct {
 	view        C.CFTypeRef
@@ -125,6 +128,7 @@ func onCreate(view C.CFTypeRef) {
 	w.w.SetDriver(w)
 	views[view] = w
 	w.w.Event(system.StageEvent{Stage: system.StagePaused})
+	w.w.Event(ViewEvent{View: uintptr(view)})
 }
 
 //export gio_onDraw
@@ -177,6 +181,7 @@ func onStop(view C.CFTypeRef) {
 func onDestroy(view C.CFTypeRef) {
 	w := views[view]
 	delete(views, view)
+	w.w.Event(ViewEvent{})
 	w.w.Event(system.DestroyEvent{})
 	w.displayLink.Close()
 	w.view = 0


app/os_ios.go
```diff
@@ -86,10 +86,17 @@ import (
 	"gioui.org/unit"
 )
 
-type ViewEvent struct{}
+type ViewEvent struct {
+	// View is a CFTypeRef for the NSView for the window.
+	View uintptr
+
+	// Window is a CFTypeRef for the NSView for the window.
+	Window uintptr
```
eliasnaur, line 94: Window is a confusing name for a NSView. Can't client code fetch the parent NSView (or NSWindow) through ViewEvent.View?

Inkeliz: I think it's possible. At least, using `[view superview]` seems to work. _Considering that `view` is the C.CFTypeRef  of `ViewEvent.View`._

eliasnaur:
> I my case I use:
> [[view superview] addSubview:myCustomView];

This is too brittle; if we ever drop the parent view, your code will silently break.
Can you confirm something like https://developer.apple.com/documentation/appkit/nswindow/1419160-contentview works too?

Inkeliz:
There's no `contentView` on UIView (https://developer.apple.com/documentation/uikit/uiview?language=objc). It gives a error, "no visible @interface for UIView".

----

In the previous PR (with View and "Window"), both are UIView, however one of them is "GioView" and another one is the "RootView" (I don't know the name of that, but is here GioView is added to):

https://github.com/gioui/gio/blob/d2dc0d52495efc6ddb476745638fa162205bba04/app/os_ios.m#L24-L33

eliasnaur:
I confused UIView and NSView, sorry.

I'm still worried we can't ever remove the parent view. Why doesn't `[view addSubView:...]` directly on the ViewEvent.View work?

Inkeliz:
Using `[view addSubView:...]` works too. 🎉

eliasnaur:
Thanks for confirming. Can you add the missing empty ViewEvent in onDestroy (see os_android.go for an example)? Otherwise, I think the change is good to go.

Inkeliz:
I add the `w.w.Event(ViewEvent{})` on `onDestroy`, but I don't think it makes sense. On iOS the `onDestroy` sends `system.DestroyEvent`, where Android doesn't. 🤔

eliasnaur:
> I add the `w.w.Event(ViewEvent{})` on `onDestroy`, but I don't think it makes sense. On iOS the `onDestroy` sends `system.DestroyEvent`, where Android doesn't. thinking

Android is special in that the GioView may be destroyed and recreated, for example during rotations. And yes, the Android backend should probably send a DestroyEvent when the GioView is gone for good.

However, the empty ViewEvent merely notifies the Gio program that the view handle is no longer valid, which is not the same as the app.Window being destroyed.

I've merged your patch with a short description in the commit message; Chris and I want to improve commit messages so they can be understood by casual Gio users, and to automate the newsletter change logs.

PR #66, opened 2021-11-15
app/internal/windows: [windows/386] fix fullscreen

Inkeliz:
Before that patch, the usage of `app.Fullscreen` may result on "Failed to
find SetWindowLongPtrW procedure in user32.dll" when running on 32Bits.


eliasnaur:
Thank you. Merged.

PR #67, opened 2021-11-18
app, cmd/gogio: add android foreground permission and service

mixmasala:
This adds the permission android.permission.FOREGROUND_SERVICE and adds
GioForegroundService which creates the tray Notification necessary to
implement the Foreground Service. The package foreground includes the method
StartForeground, which on android, notifies the system that the program
will perform background work and that it shouldn't be killed. It returns
a channel that should be closed when the background work is complete.

See https://developer.android.com/guide/components/foreground-services
and https://developer.android.com/training/notify-user/build-notification

### DIFF
--- a/app/Gio.java
+++ b/app/Gio.java
@@ -5,6 +5,8 @@ package org.gioui;
 import android.content.ClipboardManager;
 import android.content.ClipData;
 import android.content.Context;
+import android.content.Intent;
+import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
 
@@ -65,4 +67,21 @@ public final class Gio {
 	}
 
 	static private native void scheduleMainFuncs();
+
+	static Intent startForegroundService(Context ctx, String title, String text) throws ClassNotFoundException {
+		Intent intent = new Intent();
+		intent.setClass(ctx, ctx.getClassLoader().loadClass("org/gioui/GioForegroundService"));
+		intent.putExtra("title", title);
+		intent.putExtra("text", text);
+
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+			// Use startForegroundService for API level 26 and later
+			ctx.startForegroundService(intent);
+		} else {
+			// Use startService for API level 25 and earlier
+			ctx.startService(intent);
+		}
+
+		return intent;
+	}
 }
--- a/app/GioForegroundService.java
+++ b/app/GioForegroundService.java
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+package org.gioui;
+import android.app.Notification;
+import android.app.Service;
+import android.app.Notification;
+import android.app.Notification.Builder;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.ComponentName;
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.os.IBinder;
+import android.os.Build;
+import android.os.Bundle;
+
+// GioForegroundService implements a Service required to use the FOREGROUND_SERVICE
+// permission on Android, in order to run an application in the background.
+// See https://developer.android.com/guide/components/foreground-services for
+// more details. To add this permission to your application, import
+// gioui.org/app/permission/foreground and use the Start method from that
+// package to control this service.
+public class GioForegroundService extends Service {
+	private String channelName;
+
+	// ForegroundNotificationID is a default unique ID for the tray Notification of this service, as it must be nonzero.
+	private int ForegroundNotificationID = 0x42424242;
+
+	@Override public int onStartCommand(Intent intent, int flags, int startId) {
+		// Get the channel parameters from intent extras and package metadata.
+		Bundle extras = intent.getExtras();
+		String title = extras.getString("title");
+		String text = extras.getString("text");
+		Context ctx = getApplicationContext();
+		try {
+			ComponentName svc = new ComponentName(this, this.getClass());
+			Bundle metadata = getPackageManager().getServiceInfo(svc, PackageManager.GET_META_DATA).metaData;
+			if (metadata == null) {
+				throw new RuntimeException("No ForegroundService MetaData found");
+			}
+			channelName = metadata.getString("org.gioui.ForegroundChannelName");
+			String channelDesc = metadata.getString("org.gioui.ForegroundChannelDesc", "");
+			String channelID = metadata.getString("org.gioui.ForegroundChannelID");
+			int notificationID = metadata.getInt("org.gioui.ForegroundNotificationID", ForegroundNotificationID);
+			this.createNotificationChannel(channelDesc, channelID, channelName);
+			Intent launchIntent = getPackageManager().getLaunchIntentForPackage(ctx.getPackageName());
+
+			PendingIntent pending = null;
+			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+				pending = PendingIntent.getActivity(ctx, notificationID, launchIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK|PendingIntent.FLAG_IMMUTABLE);
+			} else {
+				pending = PendingIntent.getActivity(ctx, notificationID, launchIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
+			}
+			Notification.Builder builder = new Notification.Builder(ctx, channelID)
+				.setContentTitle(title)
+				.setContentText(text)
+				.setSmallIcon(getResources().getIdentifier("@mipmap/ic_launcher_adaptive", "drawable", getPackageName()))
+				.setContentIntent(pending)
+				.setPriority(Notification.PRIORITY_MIN);
+			startForeground(notificationID, builder.build());
+		} catch (PackageManager.NameNotFoundException e) {
+			throw new RuntimeException(e);
+		} catch (java.lang.SecurityException e) {
+			// XXX: notify the caller of Start that the service has failed
+			throw new RuntimeException(e);
+		}
+		return START_NOT_STICKY;
+	}
+
+	@Override public IBinder onBind(Intent intent) {
+		return null;
+	}
+
+	@Override public void onCreate() {
+		super.onCreate();
+	}
+
+	@Override
+	public void onTaskRemoved(Intent rootIntent) {
+		super.onTaskRemoved(rootIntent);
+		this.deleteNotificationChannel();
+		stopForeground(true);
+		this.stopSelf();
+	}
+
+	@Override public void onDestroy() {
+		this.deleteNotificationChannel();
+	}
+
+	private void deleteNotificationChannel() {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+			NotificationManager notificationManager = getSystemService(NotificationManager.class);
+			notificationManager.deleteNotificationChannel(channelName);
+		}
+	}
+
+	private void createNotificationChannel(String channelDesc, String channelID, String channelName) {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+			// https://developer.android.com/training/notify-user/build-notification#java
+			NotificationChannel channel = new NotificationChannel(channelID, channelName, NotificationManager.IMPORTANCE_LOW);
+			channel.setDescription(channelDesc);
+			NotificationManager notificationManager = getSystemService(NotificationManager.class);
+			notificationManager.createNotificationChannel(channel);
+		}
+	}
+}
--- a/app/os_android.go
+++ b/app/os_android.go
@@ -40,6 +40,10 @@ static jclass jni_GetObjectClass(JNIEnv *env, jobject obj) {
 	return (*env)->GetObjectClass(env, obj);
 }
 
+static jobject jni_NewObject(JNIEnv *env, jclass clazz, jmethodID methodID) {
+	return (*env)->NewObject(env, clazz, methodID);
+}
+
 static jmethodID jni_GetMethodID(JNIEnv *env, jclass clazz, const char *name, const char *sig) {
 	return (*env)->GetMethodID(env, clazz, name, sig);
 }
@@ -231,9 +235,11 @@ var android struct {
 	// gioCls is the class of the Gio class.
 	gioCls C.jclass
 
-	mwriteClipboard   C.jmethodID
-	mreadClipboard    C.jmethodID
-	mwakeupMainThread C.jmethodID
+	mwriteClipboard        C.jmethodID
+	mreadClipboard         C.jmethodID
+	mwakeupMainThread      C.jmethodID
+	startForegroundService C.jmethodID
+	stopService            C.jmethodID
 
 	// android.view.accessibility.AccessibilityNodeInfo class.
 	accessibilityNodeInfo struct {
@@ -433,6 +439,10 @@ func initJVM(env *C.JNIEnv, gio C.jclass, ctx C.jobject) {
 	android.mwriteClipboard = getStaticMethodID(env, gio, "writeClipboard", "(Landroid/content/Context;Ljava/lang/String;)V")
 	android.mreadClipboard = getStaticMethodID(env, gio, "readClipboard", "(Landroid/content/Context;)Ljava/lang/String;")
 	android.mwakeupMainThread = getStaticMethodID(env, gio, "wakeupMainThread", "()V")
+
+	cls = getObjectClass(env, android.appCtx)
+	android.stopService = getMethodID(env, cls, "stopService", "(Landroid/content/Intent;)Z")
+	android.startForegroundService = getStaticMethodID(env, gio, "startForegroundService", "(Landroid/content/Context;Ljava/lang/String;Ljava/lang/String;)Landroid/content/Intent;")
 
 	intern := func(s string) C.jstring {
 		ref := C.jni_NewGlobalRef(env, C.jobject(javaString(env, s)))
@@ -1495,3 +1505,56 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
 
 func (AndroidViewEvent) implementsViewEvent() {}
 func (AndroidViewEvent) ImplementsEvent()     {}
+
+var foregroundService struct {
+	intent C.jobject
+	mu     sync.Mutex
+	stop   map[*int]bool
+}
+
+// startForeground starts the foreground service on android
+// it returns a method that stops the foreground service, or an error
+func startForeground(title, text string) (func(), error) {
+	foregroundService.mu.Lock()
+	defer foregroundService.mu.Unlock()
+
+	var intent C.jobject
+	var err error
+	if len(foregroundService.stop) == 0 {
+		runInJVM(javaVM(), func(env *C.JNIEnv) {
+			intent, err = callStaticObjectMethod(env, android.gioCls,
+				android.startForegroundService,
+				jvalue(android.appCtx),
+				jvalue(javaString(env, title)),
+				jvalue(javaString(env, text)),
+			)
+			if err == nil {
+				// get a reference across JNI sessions to the returned intent
+				foregroundService.intent = C.jni_NewGlobalRef(env, intent)
+			}
+		})
+	}
+	if err != nil {
+		return nil, err
+	}
+	ref := new(int)
+	foregroundService.stop[ref] = true
+	return func() {
+		foregroundService.mu.Lock()
+		defer foregroundService.mu.Unlock()
+		delete(foregroundService.stop, ref)
+		// Each call to Start returns a stop() method; once the last stop() is called, stop the service.
+		if len(foregroundService.stop) == 0 {
+			runInJVM(javaVM(), func(env *C.JNIEnv) {
+				callVoidMethod(env, android.appCtx, android.stopService, jvalue(foregroundService.intent))
+				C.jni_DeleteGlobalRef(env, foregroundService.intent)
+				foregroundService.intent = 0
+			})
+		}
+	}, nil
+
+}
+
+func init() {
+	foregroundService.stop = make(map[*int]bool)
+}
--- a/app/os_nandroid.go
+++ b/app/os_nandroid.go
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+//go:build !android
+// +build !android
+
+package app
+
+// app.Start is a no-op on platforms other than android
+func startForeground(title, text string) (stop func(), err error) {
+	return func() {}, nil
+}
--- a/app/permission/foreground/main.go
+++ b/app/permission/foreground/main.go
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+/*
+Package foreground implements permissions to run a foreground service.
+See https://developer.android.com/guide/components/foreground-services.
+
+The following entries will be added to AndroidManifest.xml:
+
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+
+*/
+
+package foreground
--- a/app/app.go
+++ b/app/app.go
@@ -120,6 +120,15 @@ func DataDir() (string, error) {
 	return dataDir()
 }
 
+// Start starts the foreground service on android, notifies the system that the
+// program will perform background work and that it shouldn't be killed. The
+// foreground service is stopped using the cancel function returned by Start().
+// If multiple calls to Start are made, the foreground service will not be
+// stopped until the final cancel function has been called.
+func Start(title, text string) (stop func(), err error) {
+	return startForeground(title, text)
+}
+
 // Main must be called last from the program main function.
 // On most platforms Main blocks forever, for Android and
 // iOS it returns immediately to give control of the main


mixmasala:
I've tested this with my application; but feel that there could be better integration with the notification library gioui.org/x/notify, so that the foreground service can share the same notification channels, in particular so that the text of the background notification can be updated. I'm not sure that this change should depend on that package though, but I think that the Service class needs to be declared in the application manifest. Any help resolving/improving the above details is very much appreciated.

app/GioForegroundService.java
```diff
@@ -0,0 +1,69 @@
...
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.os.IBinder;
+
+public class GioForegroundService extends Service {
+	public static final String CHANNEL_ID = "GioForegroundService";
+	public static final String serviceName = "GioForeground";
+	public static final int notificationID = 42;
```
eliasnaur, line 18: This is likely to clash with other notification IDs. I suggest a larger (1000+) number, and maybe adding a notification ID base constant in Gio.java to make retain some control of future ID allocations.

app/GioForegroundService.java
```diff
@@ -0,0 +1,69 @@
...
+
+	@Override public int onStartCommand (Intent intent, int flags, int startId) {
+		return START_NOT_STICKY;
+	}
```
eliasnaur, line 22: Add blank lines after methods, here and elsewhere.

app/GioForegroundService.java
```diff
@@ -0,0 +1,69 @@
...
+import android.app.Notification;
+import android.app.Notification.Builder;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.os.IBinder;
+
+public class GioForegroundService extends Service {
```
eliasnaur, line 15: A short description of purpose and a linke to package foreground would be nice.

app/GioForegroundService.java
```diff
@@ -0,0 +1,69 @@
...
+	public void startForeground() {
+		// create the Intent that will bring GioActivity to foreground
+		Context ctx = getApplicationContext();
+		Intent resultIntent = new Intent(ctx, GioActivity.class);
+		PendingIntent pending = PendingIntent.getActivity(ctx, notificationID, resultIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
+		// create the Notification that Android will display on the tray for this service
+		Notification.Builder builder = new Notification.Builder(ctx, CHANNEL_ID)
+			.setContentTitle(CHANNEL_ID)
+			.setSmallIcon(getResources().getIdentifier("@mipmap/ic_launcher", "drawable", getPackageName()))
+			.setContentText(serviceName)
```
eliasnaur, line 40: Title and text are not very informative. Perhaps foreground.StartForeground should include them? It would be fine to do in a follow-up except that it's an API change.

app/GioForegroundService.java
```diff
@@ -0,0 +1,69 @@
...
+	@Override public void onDestroy() {
+		this.deleteNotificationChannel();
+	}
+	private void deleteNotificationChannel() {
+		NotificationManager notificationManager = getSystemService(NotificationManager.class);
+		notificationManager.deleteNotificationChannel(CHANNEL_ID);
+	}
+	private void createNotificationChannel() {
+		// https://developer.android.com/training/notify-user/build-notification#java
+		NotificationChannel channel = new NotificationChannel(CHANNEL_ID, serviceName, NotificationManager.IMPORTANCE_DEFAULT);
```
eliasnaur, line 64: Why not a lower importance? The notification is only here because Android forces us to.

app/GioForegroundService.java
```diff
@@ -0,0 +1,69 @@
...
+		Context ctx = getApplicationContext();
+		Intent resultIntent = new Intent(ctx, GioActivity.class);
+		PendingIntent pending = PendingIntent.getActivity(ctx, notificationID, resultIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
+		// create the Notification that Android will display on the tray for this service
+		Notification.Builder builder = new Notification.Builder(ctx, CHANNEL_ID)
+			.setContentTitle(CHANNEL_ID)
+			.setSmallIcon(getResources().getIdentifier("@mipmap/ic_launcher", "drawable", getPackageName()))
+			.setContentText(serviceName)
+			.setContentIntent(pending)
+			.setPriority(Notification.PRIORITY_DEFAULT);
```
eliasnaur, line 42: Lower priority? See also comment below about importance.

app/permission/foreground/android.go
```diff
@@ -0,0 +1,75 @@
...
+	foregroundService = "org/gioui/GioForegroundService"
+)
+
+// StartForeround notifies the system that the program will perform
+// background work and that it shouldn't be killed. It returns a channel
+// that should be closed when the background work is complete.
+
+// StartForeground is a no-op on Linux, Windows, macOS; Android will
+// display a notification during background work; iOS isn't supported.
+func StartForeground() chan struct{} {
```
eliasnaur, line 32: Given the package name, I suggest renaming to "Start" to avoid stutter.

app/permission/foreground/main.go
```diff
@@ -0,0 +1,25 @@
...
+
+package foreground
+
+// StartBackground notifies the system that the program will perform
+// background work and that it shouldn't be killed. It returns a channel
+// that should be closed when the background work is complete.
+
+// StartBackground is a no-op on Linux, Windows, macOS; Android will
+// display a notification during background work; iOS isn't supported.
+func StartForeground() chan struct{} {
```
eliasnaur, line 23: Repeating the documentation (and function signature) can be avoided by introducing an unexported startForeground function which is platform specific, and remove build constraints for Start and its documentation.

app/permission/foreground/android.go
```diff
@@ -0,0 +1,75 @@
...
+
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+
+*/
+
+package foreground
+
+import (
+	"gioui.org/app"
+	"git.wow.st/gmp/jni"
```
eliasnaur, line 19: We already have a JNI implementation in package app. I think it should be re-used instead of adding a new dependency.

app/permission/foreground/android.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// get a handle on the startForeground, stopForeground methods of GioForegroundService
+	// run everything in a goroutine so that start/stop calls are from the same thread
+	go func() {
+		jni.Do(jni.JVMFor(app.JavaVM()), func(env jni.Env) error {
+			appCtx := jni.Object(app.AppContext())
+			classLoader := jni.ClassLoaderFor(env, appCtx)
+			foregroundServiceClass, err := jni.LoadClass(env, classLoader, foregroundService)
+			if err != nil {
+				return err
```
eliasnaur, line 43: Expose errors to the caller and don't return from Start until the foregrounding has successfully begun.

eliasnaur:
Nice work, thank you.

> I've tested this with my application; but feel that there could be better integration with the notification library gioui.org/x/notify, so that the foreground service can share the same notification channels, in particular so that the text of the background notification can be updated. I'm not sure that this change should depend on that package though, but I think that the Service class needs to be declared in the application manifest. Any help resolving/improving the above details is very much appreciated.

Why share notifications with niotify? This change is for foreground work, niotify is for notifications. However, see my review comments for a suggestion to customize foreground notification content text.

eliasnaur:
Oh, and remember to sign-off your change and squash commits.

app/permission/foreground/android.go
mixmasala: ok, I didn't notice that - I copied gioui.org/x/notify's implementation - now I see this is available to package app so I will refactor here.

mixmasala:
> Nice work, thank you.
> 
> > I've tested this with my application; but feel that there could be better integration with the notification library gioui.org/x/notify, so that the foreground service can share the same notification channels, in particular so that the text of the background notification can be updated. I'm not sure that this change should depend on that package though, but I think that the Service class needs to be declared in the application manifest. Any help resolving/improving the above details is very much appreciated.
> 
> Why share notifications with niotify? This change is for foreground work, niotify is for notifications. However, see my review comments for a suggestion to customize foreground notification content text.

well, the foreground service requires a notification - the code here just sets a static string which isn't very appealing. Android applications using foreground services tend to have some kind of content displayed in the service notification and it would be useful to have a clean way to interact with the notification - possibly the StartForeground method should return or accept this as an argument - and this is where it starts to overlap with the functionality provided by niotify.

app/permission/foreground/android.go
mixmasala: initially I wrote it this way but realized that when I started a goroutine to wait on the channel before stopping the service, that the applicaiton would crash - I think because the caller is from another thread? I didn't see anything in the logcat output to hint exactly what was happening - but possibly this can be refactored by getting new handles before calling the stopService method.

app/GioForegroundService.java
mixmasala: I'm open to changing this but I think it would be more useful if these sort of choices were available to the application developer - although unfortunately this is rather platform specific. I don't think android wants you to be able to hide these notifications from the user in any case, so I'm not really sure what adjusting the priority does - but I will try to find out.

app/GioForegroundService.java
mixmasala: I'm happy to iterate here until the API looks good - rather than break it later.

app/permission/foreground/android.go
mixmasala: this is a bit trickier to do than I thought at first because the package foreground wont see the unexported methods from app.

app/GioForegroundService.java
mixmasala: ha you're telling me :P

> Oh, and remember to sign-off your change and squash commits.

Ok, I can do this and open a new PR once everything looks good.

app/permission/foreground/android.go
mixmasala: I think perhaps it makes sense to move the startForeground() into the app package - in os_android.go

mixmasala:
I've removed the dependency on the external JNI but I now get a crash with the following:
```
thread.cc:2365] No pending exception expected: java.lang.ClassNotFoundException: org.gioui.GioForegroundService
thread.cc:2365] (Throwable with empty stack trace)
thread.cc:2365]
```

app/GioForegroundService.java
eliasnaur: You can re-use this PR by force-pushing. For example:

$ git rebase -i main
$ git push --force-with-lease

app/GioForegroundService.java
eliasnaur: Not hide, so much as to give the notification a lower visibility. For example, I believe lower priority notifications do not appear when you drag down the system bar to reveal notification details.

app/permission/foreground/android.go
eliasnaur: You may have to refactor package app's JNI implementation into app/internal/jni.

app/permission/foreground/android.go
eliasnaur: You don't need to switch threads, just have an additional channel for errors that Start can wait on before returning closeChan.

```
errChan := make(chan error)
go func() {
// Process and send to errChan if failed.
...
errChan <- nil
}()

if err := <-errChan; err != nil {
...
}
``

**Comment by mixmasala** [comment]
I've refactored this PR to address issues raised above and changed the API to include setting the notification ID, title, and text.

File: app/os_android.go [hunk 11]
```diff
@@ -131,6 +139,10 @@ import (
 	"gioui.org/unit"
 )
 
+const (
+	foregroundService = "org/gioui/GioForegroundService"
+)
+
```
eliasnaur, line 145: Nit: omit parentheses.

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
 }
 
 func (_ ViewEvent) ImplementsEvent() {}
+
+// start a foreground service
+func Java_org_gioui_StartForeground(id int, title, text string) chan struct{} {
+	closeChan := make(chan struct{})
```
eliasnaur, line 1022: Add

errChan := make(chan error)

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+
+	// get a handle on the startForeground, stopForeground methods of GioForegroundService
+	// run everything in a goroutine so that start/stop calls are from the same thread
+	runInJVM(javaVM(), func(env *C.JNIEnv) {
+		// create an intent and set it to the foregroundService
+		intent, err := getIntent(env, android.appCtx, foregroundService)
+		if err != nil {
+			panic(err)
```
eliasnaur, line 1030: Replace with

errChan <- err
return

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
 
 func (_ ViewEvent) ImplementsEvent() {}
+
+// start a foreground service
+func Java_org_gioui_StartForeground(id int, title, text string) chan struct{} {
+	closeChan := make(chan struct{})
+
+	// get a handle on the startForeground, stopForeground methods of GioForegroundService
+	// run everything in a goroutine so that start/stop calls are from the same thread
+	runInJVM(javaVM(), func(env *C.JNIEnv) {
```
eliasnaur, line 1026: You can merge this function with the runInJVM below.

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
 }
 
 func (_ ViewEvent) ImplementsEvent() {}
+
+// start a foreground service
+func Java_org_gioui_StartForeground(id int, title, text string) chan struct{} {
```
eliasnaur, line 1021: Replace with

func Java_org_gioui_StartForeground(id int, title, text string) (chan struct{}, error) {

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+
+	go func() {
+		// wait for the channel to close, and then halt the foreground service
+		<-closeChan
+		runInJVM(javaVM(), func(env *C.JNIEnv) {
+			// create an intent and set it to the foregroundService
+			intent, err := getIntent(env, android.appCtx, foregroundService)
+			if err != nil {
+				panic(err)
+			}
```
eliasnaur, line 1051: Delete. This is not needed if you merge the runInJVM closures.

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+			cls := getObjectClass(env, android.appCtx)
+			stopService := getMethodID(env, cls, "stopService", "(Landroid/content/Intent;)Z")
+			callVoidMethod(env, android.appCtx, stopService, jvalue(intent))
+		})
+	}()
+
+	return closeChan
```
eliasnaur, line 1058: Replace with

return closeChan, <-errChan

app/os_android.go
```diff
@@ -951,3 +1016,44 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+
+		// locate the methods to start our service
+		cls := getObjectClass(env, android.appCtx)
+		startForegroundServiceMethod := getMethodID(env, cls, "startForegroundService", "(Landroid/content/Intent;)Landroid/content/ComponentName;")
+		// Call startForegroundService with the above intent
+		callObjectMethod(env, android.appCtx, startForegroundServiceMethod, jvalue(intent))
+	})
+
+	go func() {
+		// wait for the channel to close, and then halt the foreground service
```
eliasnaur, line 1044: Add

errChan <- nil

app/permission/foreground/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+package foreground
+
+// Start notifies the system that the program will perform
+// background work and that it shouldn't be killed. It returns a channel
+// that should be closed when the background work is complete.
+
+// Start is a no-op on Linux, Windows, macOS; Android will
+// display a notification during background work; iOS isn't supported.
+func Start(id int, title, text string) chan struct{} {
```
eliasnaur, line 21: Why let the app specify the channel ID, which is Android specific? I think hard-coding it is OK.

cmd/gogio/androidbuild.go
```diff
@@ -467,6 +469,11 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
 				<category android:name="android.intent.category.LAUNCHER" />
 			</intent-filter>
 		</activity>
+		{{if .HasService}}
+		<service android:name="org.gioui.GioForegroundService"
```
eliasnaur, line 473: If you want to let the user customize the channel ID etc, you could add a [meta-data](https://developer.android.com/guide/topics/manifest/meta-data-element) configuration element to GioForegroundService. Then Gio apps with special needs can supply their own AndroidManifest.xml with a tweaked channel ID.

cmd/gogio/androidbuild.go
```diff
@@ -867,6 +874,17 @@ func getPermissions(ps []string) ([]string, []string) {
 	return permissions, features
 }
 
+func hasForegroundPermission(permissions []string) bool {
+	if fgPerms, ok := AndroidPermissions["foreground"]; ok {
```
eliasnaur, line 878: Isn't ok always true?

cmd/gogio/androidbuild.go
```diff
@@ -867,6 +874,17 @@ func getPermissions(ps []string) ([]string, []string) {
...
+		for _, p := range permissions {
+			if p == fgPerms[0] { // []string, len 1
```
eliasnaur, line 880: Isn't fgPerms[0] a constant string? If so, use a Go constant here and for AndroidPermissions below.

app/permission/foreground/main.go
mixmasala: the id gives you a reference to the notification in case of wanting to modify/update it late. yes this is android specific.

cmd/gogio/androidbuild.go
mixmasala: I thought it should only be true when the package includes that import line?

mixmasala:
edit: stopped using startForegroundService because I saw some crash with resuming the application from background with same error as: https://github.com/transistorsoft/nativescript-background-geolocation-lt/issues/121#issuecomment-527318440 . Instead startService is used, which fixes this crash.

app/GioForegroundService.java
```diff
@@ -0,0 +1,82 @@
...
+		channelName = intent.getStringExtra("channelName");
+		channelDesc = intent.getStringExtra("channelDesc");
+		notificationID = intent.getIntExtra("notificationID", notificationID);
+		this.createNotificationChannel();
+
+		// create the Intent that will bring GioActivity to foreground
+		Context ctx = getApplicationContext();
+		Intent resultIntent = new Intent(ctx, GioActivity.class);
+		PendingIntent pending = PendingIntent.getActivity(ctx, notificationID, resultIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
+		// create the Notification that Android will display on the tray for this service
```
eliasnaur, line 39: Nit: I suggest dropping this comment. Otherwise make it a full sentence with upper case initial letter and ending with a period.

app/GioForegroundService.java
```diff
@@ -0,0 +1,82 @@
...
+
+	@Override public int onStartCommand (Intent intent, int flags, int startId) {
+		// get the channel parameters from intent extras
+		channelID   = intent.getStringExtra("channelID");
+		channelName = intent.getStringExtra("channelName");
+		channelDesc = intent.getStringExtra("channelDesc");
+		notificationID = intent.getIntExtra("notificationID", notificationID);
+		this.createNotificationChannel();
+
+		// create the Intent that will bring GioActivity to foreground
```
eliasnaur, line 35: Nit: full sentence: starting letter upper case, end with period.

app/GioForegroundService.java
```diff
@@ -0,0 +1,82 @@
...
+		Context ctx = getApplicationContext();
+		Intent resultIntent = new Intent(ctx, GioActivity.class);
```
eliasnaur, line 37: GioActivity may not be the actual class used when Gio is embedded ala Tailscale. I suggest bundling resultIntent along with channelID, channelName, channelDesc.

app/os_android.go
```diff
@@ -131,6 +139,8 @@ import (
 	"gioui.org/unit"
 )
 
+const foregroundService = "org/gioui/GioForegroundService"
```
eliasnaur, line 142: This constant is only used once. I suggest folding the constant into the code using it.

app/os_android.go
```diff
@@ -951,3 +1014,45 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
 }
 
 func (_ ViewEvent) ImplementsEvent() {}
+
+// start a foreground service
+func Java_org_gioui_StartForeground(id int, title, text string) (chan struct{}, error) {
```
eliasnaur, line 1019: Can you rename this function? It looks like a JNI function, but it isn't.

app/os_android.go
```diff
@@ -951,3 +1014,45 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+			cls := getObjectClass(env, android.appCtx)
+			startServiceMethod := getMethodID(env, cls, "startService", "(Landroid/content/Intent;)Landroid/content/ComponentName;")
+			stopServiceMethod := getMethodID(env, cls, "stopService", "(Landroid/content/Intent;)Z")
+
+			// Call startService with the above intent
+			_, err = callObjectMethod(env, android.appCtx, startServiceMethod, jvalue(intent))
+			if err != nil {
+				errChan <- err
+				return
+			}
```
eliasnaur, line 1047: This code, along with the helpers getIntent and setforegroundIntentExtras seem pointless to implement in Go and JNI. Why not add a, say, `Intent startForegroundService(activityClassName, title, text)` helper method to Gio.java and call that?

app/os_android.go
```diff
@@ -951,3 +1014,45 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+
+			// wait for the channel to close, and then halt the foreground service
+			// create an intent and set it to the foregroundService
+			errChan <- nil
+			<-closeChan
+			callVoidMethod(env, android.appCtx, stopServiceMethod, jvalue(intent))
```
eliasnaur, line 1053: This is subtly incorrect. If there are more than one pending StartForeground, the first closeChan closed would stop the service even though others are still doing foreground work.

I suggest starting a singleton foreground goroutine with once.Do, and make it accept startForeground, stopForeground commands. It will then only stop GioForegroundService if the number of outstanding foreground commands is 0.

app/permission/foreground/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+package foreground
+
+// Start notifies the system that the program will perform
+// background work and that it shouldn't be killed. It returns a channel
+// that should be closed when the background work is complete.
+
+// Start is a no-op on Linux, Windows, macOS; Android will
+// display a notification during background work; iOS isn't supported.
+func Start(id int, title, text string) (chan struct{}, error) {
```
eliasnaur, line 21: I still don't think `id` should be a parameter to Start. It should be hard-coded to some (documented) value. If we need overriding, we should do that through AndroidManifest.xml meta-data.

cmd/gogio/permission.go
```diff
@@ -19,6 +19,9 @@ var AndroidPermissions = map[string][]string{
 		"android.permission.READ_EXTERNAL_STORAGE",
 		"android.permission.WRITE_EXTERNAL_STORAGE",
 	},
+	"foreground": {
+		"android.permission.FOREGROUND_SERVICE",
```
eliasnaur, line 23: Re-use the constant `foregroundPermissions`.

cmd/gogio/androidbuild.go
```diff
@@ -867,6 +875,15 @@ func getPermissions(ps []string) ([]string, []string) {
 	return permissions, features
 }
 
+func hasForegroundPermission(permissions []string) bool {
```
eliasnaur, line 878: Generalize slightly to

func stringContains(haystack []string, needle string) bool

app/GioForegroundService.java
```diff
@@ -0,0 +1,94 @@
...
+// more details. To add this permission to your application, import
+// gioui.org/app/permission/foreground and use the Start method from that
+// package to control this service.
+public class GioForegroundService extends Service {
+	private String channelID;
+	private String channelName;
+	private String channelDesc;
+	private int notificationID;
+
+	@Override public int onStartCommand (Intent intent, int flags, int startId) {
```
eliasnaur, line 30: Nit: drop extra space before (.

app/GioForegroundService.java
```diff
@@ -0,0 +1,94 @@
...
+		// get the channel parameters from intent extras
```
eliasnaur, line 31: Nit: Full sentence comments.

cmd/gogio/androidbuild.go
```diff
@@ -467,6 +470,13 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
 				<category android:name="android.intent.category.LAUNCHER" />
 			</intent-filter>
 		</activity>
+		{{if .HasService}}
+		<service android:name="org.gioui.GioForegroundService"
+			android:stopWithTask="true">
+			<meta-data android:name="org.gioui.GioForegroundChannelID"
+			android:value="GioForegroundServiceChannel" />
```
eliasnaur, line 477: Nit: Drop "Channel" from the value, it seems redundant.

Also, is "org.gioui.ForegroundService" perhaps a better name?

cmd/gogio/androidbuild.go
```diff
@@ -467,6 +470,13 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
 				<category android:name="android.intent.category.LAUNCHER" />
 			</intent-filter>
 		</activity>
+		{{if .HasService}}
+		<service android:name="org.gioui.GioForegroundService"
+			android:stopWithTask="true">
+			<meta-data android:name="org.gioui.GioForegroundChannelID"
```
eliasnaur, line 476: Rename parameter to just "org.gioui.ForegroundChannelID", reducing stutter.

app/os_android.go
```diff
@@ -951,3 +1012,40 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+	// run everything in a goroutine so that start/stop calls are from the same thread
+	go func() {
+		runInJVM(javaVM(), func(env *C.JNIEnv) {
+			startForegroundService := getStaticMethodID(env, android.gioCls,
+				"startForegroundService", "(Landroid/content/Context;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)Landroid/content/Intent;")
+
+			intent, err := callStaticObjectMethod(env, android.gioCls,
+				startForegroundService,
+				jvalue(android.appCtx),
+				jvalue(javaString(env, "org/gioui/GioForegroundService")),
```
eliasnaur, line 1031: You provide the name of the foreground service class, but that's a constant. I had expected "org/gioui/GioActivity", optionally fetched from meta-data.

app/GioForegroundService.java
```diff
@@ -0,0 +1,94 @@
...
+// permission on Android, in order to run an application in the background.
+// See https://developer.android.com/guide/components/foreground-services for
+// more details. To add this permission to your application, import
+// gioui.org/app/permission/foreground and use the Start method from that
+// package to control this service.
+public class GioForegroundService extends Service {
+	private String channelID;
+	private String channelName;
+	private String channelDesc;
+	private int notificationID;
```
eliasnaur, line 28: Can some (or all) of these be local variables and method parameters?

app/GioForegroundService.java
```diff
@@ -0,0 +1,94 @@
...
+			if (metadata != null) {
+				channelID = metadata.getString("org.gioui.GioForegroundChannelID");
+				notificationID = metadata.getInt("org.gioui.GioForegroundNotificationID", 0x42424242);
+			}
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+		channelName = extras.getString("channelName");
+		channelDesc = extras.getString("channelDesc");
+		notificationID = extras.getInt("notificationID", notificationID);
```
eliasnaur, line 46: You overwrite notificationID from meta-data here.

app/GioForegroundService.java
```diff
@@ -0,0 +1,94 @@
...
+	private int notificationID;
+
+	@Override public int onStartCommand (Intent intent, int flags, int startId) {
+		// get the channel parameters from intent extras
+		Bundle extras = intent.getExtras();
+		Context ctx = getApplicationContext();
+		try {
+			ComponentName svc = new ComponentName(this, this.getClass());
+			Bundle metadata = getPackageManager().getServiceInfo(svc, PackageManager.GET_META_DATA).metaData;
+			if (metadata != null) {
```
eliasnaur, line 37: I suggest requiring all parameters set in meta-data, returning early with a log warning if they're not present. That way, we don't have to hardcode any of the values.

app/Gio.java
```diff
@@ -65,4 +66,17 @@ static void wakeupMainThread() {
 	}
 
 	static private native void scheduleMainFuncs();
+
+	static Intent startForegroundService(Context ctx, String activityClassName, String title, String text) {
+		Intent intent = new Intent();
+		try {
+			intent.setClass(ctx, ctx.getClassLoader().loadClass(activityClassName));
```
eliasnaur, line 73: You named the parameter activityClassName (as I'd expect), but use it as the service class name. See also comment below about GioActivity.

eliasnaur (commented):
Leaving a few review comments for your follow-ups.

cmd/gogio/androidbuild.go
mixmasala: GioForegroundService just follows the same naming convention as GioActivity, GioApp, etc.

app/os_android.go
mixmasala: hm yeah. not sure what happened here.

cmd/gogio/androidbuild.go
eliasnaur: GioForegroundService is ok as service class name; I was referring to the meta-data value. To be clear, I suggest:

```
<meta-data android:name="org.gioui.ForegroundChannelID" android:value="GioForegroundService"/>
```

eliasnaur: Or maybe even

<meta-data android:name="org.gioui.ForegroundChannelID" android:value="Foreground service"/>

in case the value ends up visible to the user (does it?) in the app settings.

mixmasala:
I'm not sure why the test failed above:
```
$ go test -race ./...
?   	gioui.org/app	[no test files]
?   	gioui.org/app/internal/log	[no test files]
?   	gioui.org/app/internal/xkb	[no test files]
?   	gioui.org/app/permission	[no test files]
?   	gioui.org/app/permission/bluetooth	[no test files]
?   	gioui.org/app/permission/camera	[no test files]
?   	gioui.org/app/permission/foreground	[no test files]
?   	gioui.org/app/permission/networkstate	[no test files]
?   	gioui.org/app/permission/storage	[no test files]
ok  	gioui.org/f32	0.025s
?   	gioui.org/font/gofont	[no test files]
ok  	gioui.org/font/opentype	0.106s
ok  	gioui.org/gesture	0.044s
?   	gioui.org/gpu	[no test files]
ok  	gioui.org/gpu/headless	1.972s
?   	gioui.org/gpu/internal/d3d11	[no test files]
?   	gioui.org/gpu/internal/driver	[no test files]
?   	gioui.org/gpu/internal/metal	[no test files]
?   	gioui.org/gpu/internal/opengl	[no test files]
ok  	gioui.org/gpu/internal/rendertest	13.211s
?   	gioui.org/gpu/internal/vulkan	[no test files]
?   	gioui.org/internal/byteslice	[no test files]
?   	gioui.org/internal/egl	[no test files]
ok  	gioui.org/internal/f32color	0.187s
ok  	gioui.org/internal/fling	0.020s
?   	gioui.org/internal/gl	[no test files]
?   	gioui.org/internal/ops	[no test files]
?   	gioui.org/internal/scene	[no test files]
?   	gioui.org/internal/stroke	[no test files]
?   	gioui.org/internal/vk	[no test files]
?   	gioui.org/io/clipboard	[no test files]
?   	gioui.org/io/event	[no test files]
?   	gioui.org/io/key	[no test files]
ok  	gioui.org/io/pointer	0.039s
?   	gioui.org/io/profile	[no test files]
ok  	gioui.org/io/router	0.056s
?   	gioui.org/io/system	[no test files]
ok  	gioui.org/layout	0.043s
ok  	gioui.org/op	0.036s
ok  	gioui.org/op/clip	0.031s
?   	gioui.org/op/paint	[no test files]
ok  	gioui.org/text	0.034s
?   	gioui.org/unit	[no test files]
ok  	gioui.org/widget	0.380s
ok  	gioui.org/widget/material	0.032s
```

app/GioForegroundService.java
mixmasala:
```
diff --git a/app/Gio.java b/app/Gio.java
index d6db219c..30a13f30 100644
--- a/app/Gio.java
+++ b/app/Gio.java
@@ -2,10 +2,13 @@
 
 package org.gioui;
 
+import android.app.Activity;
 import android.content.ClipboardManager;
 import android.content.ClipData;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 
@@ -67,16 +70,25 @@ public final class Gio {
 
        static private native void scheduleMainFuncs();
 
-       static Intent startForegroundService(Context ctx, String serviceClassName, String activityClassName, String title, String text) {
+       static Intent startForegroundService(Context ctx, String title, String text) {
                Intent intent = new Intent();
                try {
-                       intent.setClass(ctx, ctx.getClassLoader().loadClass(serviceClassName));
+                       Activity activity = (Activity) ctx;
+                       Bundle metadata = ctx.getPackageManager().getActivityInfo(activity.getComponentName(), PackageManager.GET_META_DATA).metaData;
+                       if (metadata != null) {
+                               String serviceClassName = metadata.getString("org.gioui.ForegroundServiceClass");
+                               intent.setClass(ctx, ctx.getClassLoader().loadClass(serviceClassName));
+                       } else {
+                               throw new RuntimeException("No ForegroundService MetaData found");
+                       }
                } catch (ClassNotFoundException e) {
                        throw new RuntimeException(e);
+               } catch (PackageManager.NameNotFoundException e) {
+                       throw new RuntimeException(e);
                }
+
                intent.putExtra("title", title);
                intent.putExtra("text", text);
-               intent.putExtra("activityClass", activityClassName);
                ctx.startService(intent);
                return intent;
        }
diff --git a/app/GioForegroundService.java b/app/GioForegroundService.java
index 1bef2e2c..4a587baa 100644
--- a/app/GioForegroundService.java
+++ b/app/GioForegroundService.java
@@ -37,9 +37,10 @@ public class GioForegroundService extends Service {
                                channelName = metadata.getString("org.gioui.ForegroundChannelName");
                                String channelDesc = metadata.getString("org.gioui.ForegroundChannelDesc");
                                String channelID = metadata.getString("org.gioui.ForegroundChannelID");
+                               String resultActivity = metadata.getString("org.gioui.ForegroundResultActvity");
                                int notificationID = metadata.getInt("org.gioui.ForegroundNotificationID", 0x42424242);
                                this.createNotificationChannel(channelDesc, channelID, channelName);
-                               Intent resultIntent = new Intent(ctx, ctx.getClassLoader().loadClass(extras.getString("activityClass")));
+                               Intent resultIntent = new Intent(ctx, ctx.getClassLoader().loadClass(resultActivity));
                                PendingIntent pending = PendingIntent.getActivity(ctx, notificationID, resultIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
                                Notification.Builder builder = new Notification.Builder(ctx, channelID)
                                        .setContentTitle(title)
diff --git a/app/os_android.go b/app/os_android.go
index 006b763c..af27e60c 100644
--- a/app/os_android.go
+++ b/app/os_android.go
@@ -969,22 +969,17 @@ var foregroundService struct {
 
 // StartForeground starts the foreground service
 func StartForeground(title, text string) (err error) {
-       const serviceClass = "org/gioui/GioForegroundService"
-       const activityClass = "org/gioui/GioActivity"
-
        foregroundService.mu.Lock()
        defer foregroundService.mu.Unlock()
        foregroundService.once.Do(func() {
                runInJVM(javaVM(), func(env *C.JNIEnv) {
                        startForegroundService := getStaticMethodID(env, android.gioCls,
                                "startForegroundService",
-                               "(Landroid/content/Context;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)Landroid/content/Intent;",
+                               "(Landroid/content/Context;Ljava/lang/String;Ljava/lang/String;)Landroid/content/Intent;",
                        )
                        foregroundService.intent, err = callStaticObjectMethod(env, android.gioCls,
                                startForegroundService,
                                jvalue(android.appCtx),
-                               jvalue(javaString(env, serviceClass)),
-                               jvalue(javaString(env, activityClass)),
                                jvalue(javaString(env, title)),
                                jvalue(javaString(env, text)),
                        )
diff --git a/cmd/gogio/androidbuild.go b/cmd/gogio/androidbuild.go
index bb216191..559bb9e3 100644
--- a/cmd/gogio/androidbuild.go
+++ b/cmd/gogio/androidbuild.go
@@ -469,6 +469,9 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
                                <action android:name="android.intent.action.MAIN" />
                                <category android:name="android.intent.category.LAUNCHER" />
                        </intent-filter>
+                       <meta-data android:name="org.gioui.ForegroundServiceClass"
+                       android:value="org/gioui/GioForegroundService" />
+
                </activity>
                {{if .HasService}}
                <service android:name="org.gioui.GioForegroundService"
@@ -481,6 +484,8 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
                        android:value="ForegroundChannelDesc" />
                        <meta-data android:name="org.gioui.ForegroundNotificationID"
                        android:value="0x42424242" />
+                       <meta-data android:name="org.gioui.ForegroundResultActivity"
+                       android:value="org/gioui/GioActivity" />
                </service>
                {{end}}
        </application>
```
The above diff builds but doesn't work and as far as I can see doesn't drop any error in logcat either. So, I'm a bit confused how to debug this.

mixmasala: Above diff applied:
https://github.com/mixmasala/gio/pull/new/wip_foreground_class_metadata

app/os_android.go
```diff
@@ -951,3 +959,61 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+	if foregroundService.count == 0 {
+		return errors.New("No foreground service running")
+	}
+	if foregroundService.count == 1 {
+		runInJVM(javaVM(), func(env *C.JNIEnv) {
+			cls := getObjectClass(env, android.appCtx)
+			stopServiceMethod := getMethodID(env, cls, "stopService", "(Landroid/content/Intent;)Z")
+			err = callVoidMethod(env, android.appCtx, stopServiceMethod, jvalue(foregroundService.intent))
+			// dispose of the global reference to intent
+			C.jni_DeleteGlobalRef(env, foregroundService.intent)
```
eliasnaur, line 1013: Defer the call to DeleteGlobalRef in the beginning of the closure.

app/GioForegroundService.java
eliasnaur: What do you mean by doesn't work? Some notes:

- Any exceptions thrown from Java is reflected in the returned error value. Make sure you check them.
- You don't need a meta-data element for ForegroundResultActivity. You can [query the current app's launcher activity](https://stackoverflow.com/questions/14129274/get-instance-of-current-android-intent-category-launcher-activity).
- Why not hard-code ForegroundServiceClass name? I don't see the need to have a meta-data parameter for it.

mixmasala: ah nice. thanks!

app/os_android.go
```diff
@@ -951,3 +959,59 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+	intent C.jobject
+	mu     sync.Mutex
+	count  int
+}
+
+// StartForeground starts the foreground service
+func StartForeground(title, text string) (err error) {
+	foregroundService.mu.Lock()
+	defer foregroundService.mu.Unlock()
+	foregroundService.once.Do(func() {
```
eliasnaur, line 974: You don't need a sync.Once when using a mutex. Just check that foregroundService.count is 0.

app/os_android.go
```diff
@@ -951,3 +959,59 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+		})
+	})
+	if err == nil {
+		foregroundService.count++
+	}
+	return
+}
+
+// StopForeground stops the foreground service
+func StopForeground() (err error) {
```
eliasnaur, line 999: A bare StopForeground makes it easy to stop a different foreground intent than intended. I suggest getting rid of foreground.Stop and change Start to

func Start(title, text string) (stop <-chan struct{}, err error)

where `close(stop)` will stop the foreground notification.

To implement this, I suggest starting a goroutine in foreground.Start that waits for stop to close. I believe you will need to replace foregroundService.count with a map to keep track of the active foregrounds.

app/Gio.java
```diff
@@ -65,4 +66,19 @@ static void wakeupMainThread() {
 	}
 
 	static private native void scheduleMainFuncs();
+
+	static Intent startForegroundService(Context ctx, String title, String text) {
+		Intent intent = new Intent();
+		intent.setClass(ctx, GioForegroundService.class);
+		try {
+			intent.setClass(ctx, ctx.getClassLoader().loadClass("org/gioui/GioForegroundService"));
+		} catch (ClassNotFoundException e) {
```
eliasnaur, line 75: You don't need to coerce ClassNotFoundException to RuntimeException. It's simpler to just declare the method can throw it.

app/GioForegroundService.java
```diff
@@ -0,0 +1,92 @@
...
+				PendingIntent pending = PendingIntent.getActivity(ctx, notificationID, launchIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
+				Notification.Builder builder = new Notification.Builder(ctx, channelID)
+					.setContentTitle(title)
+					.setContentText(text)
+					.setSmallIcon(getResources().getIdentifier("@mipmap/ic_launcher_adaptive", "drawable", getPackageName()))
+					.setContentIntent(pending)
+					.setPriority(Notification.PRIORITY_MIN);
+				startForeground(notificationID, builder.build());
+			} else {
+				throw new RuntimeException("No ForegroundService MetaData found");
```
eliasnaur, line 52: Throw RuntimeException immediately, Go style:

```
if (metadata == null) {
    throw ...
}
// query meta-data and startForeground.
```

app/GioForegroundService.java
```diff
@@ -0,0 +1,92 @@
...
+			Bundle metadata = getPackageManager().getServiceInfo(svc, PackageManager.GET_META_DATA).metaData;
+			if (metadata != null) {
+				channelName = metadata.getString("org.gioui.ForegroundChannelName");
+				String channelDesc = metadata.getString("org.gioui.ForegroundChannelDesc");
+				String channelID = metadata.getString("org.gioui.ForegroundChannelID");
+				int notificationID = metadata.getInt("org.gioui.ForegroundNotificationID", 0x42424242);
+				this.createNotificationChannel(channelDesc, channelID, channelName);
+				Intent launchIntent = getPackageManager().getLaunchIntentForPackage(ctx.getPackageName());
+				PendingIntent pending = PendingIntent.getActivity(ctx, notificationID, launchIntent, Intent.FLAG_ACTIVITY_CLEAR_TASK);
+				Notification.Builder builder = new Notification.Builder(ctx, channelID)
```
eliasnaur, line 44: The channel constructor is only available in API 26 and later.

app/os_android.go
```diff
@@ -951,3 +959,59 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
+
+var foregroundService struct {
+	once   sync.Once
+	intent C.jobject
+	mu     sync.Mutex
+	count  int
+}
+
+// StartForeground starts the foreground service
+func StartForeground(title, text string) (err error) {
```
eliasnaur, line 971: It's not great to have `StartForeground` exported. Can you move the implementation to package foreground?

cmd/gogio/androidbuild.go
```diff
@@ -467,6 +470,19 @@ func exeAndroid(tmpDir string, tools *androidTools, bi *buildInfo, extraJars, pe
...
 		</activity>
+		{{if .HasService}}
+		<service android:name="org.gioui.GioForegroundService"
+			android:stopWithTask="true">
+			<meta-data android:name="org.gioui.ForegroundChannelID"
+			android:value="ForegroundChannelID" />
+			<meta-data android:name="org.gioui.ForegroundChannelName"
+			android:value="ForegroundChannelName" />
+			<meta-data android:name="org.gioui.ForegroundChannelDesc"
+			android:value="ForegroundChannelDesc" />
```
eliasnaur, line 481: Can the description be optional if there's nothing interesting to put there? Otherwise add a short description.

app/GioForegroundService.java
mixmasala: https://support.google.com/googleplay/android-developer/answer/9859152?visit_id=637402646062706504-1797308153&rd=1#targetsdk&zippy=%2Ctarget-api-level-requirements-for-play-console

wontfix.

app/os_android.go
mixmasala: I cannot move StartForeground out of package app without exporting methods used from app.

app/GioForegroundService.java
```diff
@@ -0,0 +1,92 @@
...
+		notificationManager.deleteNotificationChannel(channelName);
+	}
+
+	private void createNotificationChannel(String channelDesc, String channelID, String channelName) {
+		// https://developer.android.com/training/notify-user/build-notification#java
+		NotificationChannel channel = new NotificationChannel(channelID, channelName, NotificationManager.IMPORTANCE_LOW);
+		channel.setDescription(channelDesc);
+		NotificationManager notificationManager = getSystemService(NotificationManager.class);
+		notificationManager.createNotificationChannel(channel);
+	}
```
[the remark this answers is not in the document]
mixmasala: I dont't think it's possible to build with gogio before android API 31 now.

app/GioForegroundService.java
eliasnaur: Note the difference between `target` SDK and `minimum` SDK. It's true that our target must be at least 30, and that `gogio` now requires SDK 31 or later to build.

However, our `minimum` SDK is [16](https://git.sr.ht/~eliasnaur/gio/tree/main/item/cmd/gogio/androidbuild.go#L362) which means that Android code should be prepared to deal with missing API.

app/GioForegroundService.java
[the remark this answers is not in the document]
eliasnaur: See comment above.

app/os_android.go
eliasnaur: To match API such as `context`, it's probably better to return a closure:

func Start(title, text string) (stop func(), err error)

instead.

mixmasala: I don't understand why you think there are multiple foreground processes - GioForegroundService is a singleton and just needs to be running in order for the application to not be killed when the application no longer has foreground focus. The caller is responsible for deciding when the application needs GioForegroundService running, so even if the API should permit the caller to be lazy and call StartForeground() whenever it feels like, there is only the single intent for startForegroundService (hard coded) that is passed to stopService.

I'll update the api again, but I really want to get this PR merged upstream because the reason I am using gio is to avoid having to muck about with android platform internals...

eliasnaur: > I don't understand why you think there are multiple foreground processes - GioForegroundService is a singleton

I didn't mean to imply multiple foreground services would exist. I meant that multiple background computations may happen concurrently. For example, let's say a component in your app is compressing a video while another is receiving new messages from a server. Those two processes are not related, but how would they coordinate their calls to Start and Stop?

I apologize for not being clearer.

app/os_android.go
```diff
@@ -1444,3 +1448,51 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
 }
 
 func (_ ViewEvent) ImplementsEvent() {}
+
+var foregroundService struct {
+	intent C.jobject
+	mu     sync.Mutex
+	count  int
```
eliasnaur, line 1455: This is much better than before, but users can still call their stop functions more than once. I suggest a `map[*int]bool` where each call to `app.Start` registers a `new(int)` token in the map. The stop function then returns eearly if the map no longer contains the token.

eliasnaur (commented):
Looks pretty good! I have a handful of comments, but it seems we can soon merge this.

app/Gio.java
```diff
@@ -65,4 +66,13 @@ static void wakeupMainThread() {
 	}
 
 	static private native void scheduleMainFuncs();
+
+	static Intent startForegroundService(Context ctx, String title, String text) throws ClassNotFoundException {
+		Intent intent = new Intent();
+		intent.setClass(ctx, ctx.getClassLoader().loadClass("org/gioui/GioForegroundService"));
```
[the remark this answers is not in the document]
mixmasala: I think that I tried doing that but the class doesn't exist unless it was first loaded.

app/os_android.go
mixmasala: I'm afraid I don't see how users can call their stop functions more than once as the returned function checks count while ohlding the lock in order to prevent stopService from being called more than once.

There should be only one instance of the GioForegroundService, so multiple calls to Start() only indicate that some part of the application is doing something that needs to continue running, and when all of the associated stop func() have been called should the service be actually stopped.

I don't see why a map is useful here, where a counter suffices.

mixmasala:
For some reason github wont let me comment on the change requests directly any more.

I moved code out of gioui.org/app/permissions, but I was not able to find a solution to check that gioui.org/app/permissions/foreground was imported when built. For example, I thought that BuildInfo might have package-level resolution, but it does not:
```
+func hasPermission(permission string) bool {
+       permissionPath := filepath.Join("gioui.org/app/permission/", permission)
+       if info, ok := debug.ReadBuildInfo(); ok {
+               for _, module := range info.Deps {
+                       fmt.Println("Imports: ", module.Path)
+                       // XXX: we do not have a way to get the list of included go packages at runtime
+                       if module.Path == permissionPath {
+                               return true
+                       }
+               }
+       }
+       return false
+}

File: app/permission/foreground/main.go [hunk 53]
```diff
@@ -0,0 +1,23 @@
...
+
+package foreground
+
+// Start notifies the system that the program will perform background work and
+// that it shouldn't be killed. It returns a CancelFunc that should be called
+// when the background work is complete.
+
+// Start is a no-op on Linux, Windows, macOS; Android will
+// display a notification during background work; iOS isn't supported.
+func Start(title, text string) (func(), error) {
```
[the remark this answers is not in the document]
mixmasala: Ok, I moved code out of the app/permission path and added os_nandroid.go to add a no-op Start() method for platforms other than android so that calls to Start() and its returned stop method have no effect.

app/Gio.java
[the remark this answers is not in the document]
eliasnaur: Please post the error message you get. It doesn't make sense to me that `org.gioui.Gio` can be loaded but `org.gioui.GioForegroundService` can't.

app/os_android.go
eliasnaur: A program may have two unrelated background jobs, say (a) uploading an image and (b) checking for new messages. If they run concurrently, the job count is incremented to 2. A programming mistake causes (a) to calls its stop function twice, decrementing the count to 0, even though (b) is still in progress. The resulting missing notification is very hard to notice.

A map allows us to detect and ignore multiple calls to any particular stop function. In other words, it is no longer a (subtle) mistake to call your stop function twice.

eliasnaur:
> For some reason github wont let me comment on the change requests directly any more.
> 
> I moved code out of gioui.org/app/permissions, but I was not able to find a solution to check that gioui.org/app/permissions/foreground was imported when built. For example, I thought that BuildInfo might have package-level resolution, but it does not:
> 
> ```
> +func hasPermission(permission string) bool {
> +       permissionPath := filepath.Join("gioui.org/app/permission/", permission)
> +       if info, ok := debug.ReadBuildInfo(); ok {
> +               for _, module := range info.Deps {
> +                       fmt.Println("Imports: ", module.Path)
> +                       // XXX: we do not have a way to get the list of included go packages at runtime
> +                       if module.Path == permissionPath {
> +                               return true
> +                       }
> +               }
> +       }
> +       return false
> +}
> ```

I don't think you need to check for the import. It should be enough to detect the Java exception/error when trying to open a foreground notification.

app/os_android.go
mixmasala: ok, this makes sense to me. Thank you for the explanation.

mixmasala:
> I don't think you need to check for the import. It should be enough to detect the Java exception/error when trying to open a foreground notification.

Good Idea. I tried to catch java.lang.SecurityException, but the exception is not raised to the caller of android.startForegroundService, as it is thrown in the method onStartCommand which is called by the android environment after the service is started. Do you have any ideas how to handle this?

app/Gio.java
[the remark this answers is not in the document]
mixmasala: ActivityManager
Unable to start service Intent
{ cmp=org.mixnetworks.katzen/
org.gioui.GioForegroundServie (has extras) } U=0:
not found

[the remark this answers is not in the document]
mixmasala: https://github.com/mixmasala/gio/blob/4c7028ff5fe3e9a562dbd20f5bb4a24aab4d2ff1/app/Gio.java#L73

eliasnaur:
> > I don't think you need to check for the import. It should be enough to detect the Java exception/error when trying to open a foreground notification.
> 
> Good Idea. I tried to catch java.lang.SecurityException, but the exception is not raised to the caller of android.startForegroundService, as it is thrown in the method onStartCommand which is called by the android environment after the service is started. Do you have any ideas how to handle this?

One idea is to make `onStartCommand` report back to Gio whether the foreground notification succeeded or not. I think you'll need to wrap a `chan error` or similar with `cgo.NewHandle` to pass it to the `Intent` argument to `onStartCommand`. Then, in `app.StartForeground` wait for the channel to return an `error` (and free the handle).

app/os_android.go
mixmasala: (changes made)

eliasnaur (commented):
I somehow failed to send this review out after writing it. I apologize.

app/Gio.java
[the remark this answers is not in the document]
mixmasala: do you know how to get android to produce a stack trace here?

app/os_nandroid.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+//go:build !android
+// +build !android
+
+package app
+
+// Start is a no-op on platforms other than android
+func Start(title, text string) (stop func(), err error) {
```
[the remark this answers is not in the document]
mixmasala: done

app/os_android.go
```diff
@@ -1444,3 +1454,53 @@ func Java_org_gioui_Gio_scheduleMainFuncs(env *C.JNIEnv, cls C.jclass) {
...
 func (_ ViewEvent) ImplementsEvent() {}
+
+var foregroundService struct {
+	intent C.jobject
+	mu     sync.Mutex
+	stop   map[*int]bool
+}
+
+// Start starts the foreground service
+func Start(title, text string) (stop func(), err error) {
```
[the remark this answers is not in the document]
mixmasala: done

ddkwork:
looks seem good

dosgo:
Is there any progress on this? Write Android applications in pure Go by adding services and BroadcastReceiver support.

PR #70, opened 2021-12-07
app: [iOS] add ViewController on ViewEvent

Inkeliz:
New "ViewController" field on ViewEvent, which contain UIViewController for
iOS. That is necessary to open custom UIViewController on top of Gio.

### DIFF
--- a/app/os_ios.go
+++ b/app/os_ios.go
@@ -87,8 +87,8 @@ import (
 )
 
 type ViewEvent struct {
-	// View is a CFTypeRef for the UIView backing a Window.
-	View uintptr
+	// ViewController is a CFTypeRef for the UIViewController backing a Window.
+	ViewController uintptr
 }
 
 type window struct {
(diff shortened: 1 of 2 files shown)


eliasnaur:
Can the ViewController be inferred from the View (e.g. through a method on UIVIew)?

Conversely, do we need View if we add ViewController?

Inkeliz:
I don't find any way to get the Controller from the View. But, it's seens possible to get the UIView from the UIViewController (https://developer.apple.com/documentation/uikit/uiviewcontroller/1621460-view?language=objc).

whereswaldon:
Can we then simplify to only providing the controller in the `ViewEvent` instead of both?

Inkeliz:
I force-push to remove the `View`. But, should keep the name `View` or change to `ViewController` (since it's `UIViewController now)?

whereswaldon:
Let's use the most descriptive name we can (`ViewController`), and let's include a link to "how to get the view from the viewcontroller" in the documentation of the field.

eliasnaur:
I don't think a link for fetching the UIView is necessary; users of ViewEvent can be assumed to have a good understanding of the UIKit, and accessing the UIView may not even be the most common use of ViewEvent.

Merged with a commit message tweak that makes it clear that ViewController is a replacement for View.

PR #72, opened 2021-12-16
change Shape() to return a PathSpec

chmike:
Tests have been run with success. 

What I'm not fully sure is if a `clip.PathSpec{}` is a valid empty PathSpec.


eliasnaur:
Thanks! I squashed your commits before submitting to keep the tests building at every commit.

PR #73, opened 2021-12-27
cmd/gogio: add -minsdk to iOS

Inkeliz:
Before that patch the minimum iOS version was hardcoded. That patch makes possible
to change the minimum iOS version using `-minsdk`.


eliasnaur:
The build failures seem real.

PR #74, opened 2021-12-27
app,io/router: fix semantics index out of range

Inkeliz:
Fix gio#328


eliasnaur:
Alternative fix: https://gioui.org/commit/cce0a121.

PR #75, opened 2021-12-27
app: [android] fix vulkan crash when activity/surface is recreated

Inkeliz:
On Android it's possible that the Activity/View must be restore. But,
before that patch, an new VkSurfaceKHR is created but never used and
that may lead to crash, in an attempt to destroy/use one invalid
surface.

Fixes gio#327

### DIFF
--- a/app/vulkan_android.go
+++ b/app/vulkan_android.go
@@ -84,5 +84,6 @@ func (c *wlVkContext) Refresh() error {
 	if err != nil {
 		return err
 	}
-	return c.ctx.refresh(surf, w, h)
+	c.surf = surf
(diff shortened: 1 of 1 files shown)


eliasnaur:
> On Android it's possible that the Activity/View must be restore. But, before that patch, an new VkSurfaceKHR is created but never used and that may lead to crash, in an attempt to destroy/use one invalid surface.
> 

Thank you. Applied with a change that resets `surf` in case vk.CreateAndroidSurface fails.

> Fixes gio#327
> 

Changed to use the new [automatic format](https://man.sr.ht/git.sr.ht/#fixes).

PR #76, opened 2022-01-31
cmd/gogio: use JAVA_HOME first

Inkeliz:
Fixes: https://todo.sr.ht/~eliasnaur/gio/349


Inkeliz:
Maybe it's better to only use `JAVA_HOME` and drop the `exec.LookPath("javac")`.

eliasnaur:
Thanks, merged.

PR #78, opened 2022-02-14
internal/gl: [js] replace Call to Bind/Invoke

Inkeliz:
Before that patch, the Call function was used to call each
JS function related to WebGL. The Call function contains
strings, which is slow on most browsers. Now, it uses
Bind on the initialization and re-use the same function,
avoiding the usage of strings.

### DIFF
--- a/internal/gl/gl_js.js
+++ b/internal/gl/gl_js.js
@@ -1,0 +1,18 @@
+(() => {
+
+    window._fastReadPixels = function (fn, x, y, width, height, format, ty, dataPtr, dataLen) {
+        fn(x, y, width, height, format, ty, new Uint8Array(go._inst.exports.mem.buffer, dataPtr, dataLen))
+    }
+
+    window._fastBufferData = function (fn, target, dataPtr, dataLen, usage) {
+        fn(target, new Uint8Array(go._inst.exports.mem.buffer, dataPtr, dataLen), usage)
+    }
+
+    window._fastBufferSubData = function (fn, target, offset, dataPtr, dataLen) {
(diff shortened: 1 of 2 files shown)


Inkeliz:
That is based on https://github.com/gioui/gio/pull/57, but removes the `map` and lookup-table. Instead, create `js.Value` on the `Context` struct. 

Measuring the frame-time, using Kitchen, that patch improves from ~6.9ms to ~5.5ms per frame. _That still slower than #4 which drops it to ~3.4ms._

I'll do another commit under that same branch (since one needs another), the next one maybe make things faster.

eliasnaur:
I took your first commit, thank you. I'm not sure about the second commit, since it seems to depend on Go runtime private details.

internal/gl/gl_js.js
```diff
@@ -0,0 +1,18 @@
+(() => {
+
+    window._fastReadPixels = function (fn, x, y, width, height, format, ty, dataPtr, dataLen) {
+        fn(x, y, width, height, format, ty, new Uint8Array(go._inst.exports.mem.buffer, dataPtr, dataLen))
```
eliasnaur, line 4: `go._inst.exports.mem.buffer` is internal to the Go runtime, no?

Inkeliz: That is "public", on JS perspective, since it's exported. But, I think it's not worth. The main issue is the `makeArgs` from `syscall/js` and the `js.Value` which causes alot of allocations. Removing the `CopyBytesToJS` and `New` will safe some allocs, but not significant.

I'm closing it. I think that is the best that we can do with `syscall/js`.

PR #79, opened 2022-02-16
app: [Android] handle DPAD_CENTER keycode

tmm1:
see https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_DPAD_CENTER

cc https://github.com/tailscale/tailscale/issues/1611#issuecomment-997622044


eliasnaur:
Thanks! Please [sign off](https://gioui.org/doc/contribute) your patch to allow me to merge it.

tmm1:
Done

PR #80, opened 2022-02-22
io/pointer: [wasm] add missing cursors

Inkeliz:

### DIFF
--- a/io/pointer/pointer.go
+++ b/io/pointer/pointer.go
@@ -111,21 +111,21 @@ const (
 	// CursorNone hides the cursor. To show it again, use any other cursor.
 	CursorNone CursorName = "none"
 	// CursorTopLeftResize is the cursor for top-left corner resizing.
-	CursorTopLeftResize CursorName = "top-left-resize"
+	CursorTopLeftResize CursorName = "nw-resize"
 	// CursorTopRightResize is the cursor for top-right corner resizing.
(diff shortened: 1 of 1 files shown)


Inkeliz:
I rename the cursors because it's used directly on JS:

https://github.com/gioui/gio/blob/bed5902476c408308230c4d4dfb53e81d4ca1a6e/app/os_js.go#L548-L551

Maybe should it change to `switch` or add some comments in the commit (or in the code)?

eliasnaur:
> I rename the cursors because it's used directly on JS:
> 
> https://github.com/gioui/gio/blob/bed5902476c408308230c4d4dfb53e81d4ca1a6e/app/os_js.go#L548-L551
> 
> Maybe should it change to `switch` or add some comments in the commit (or in the code)?

The names shouldn't be dictated by a platform, but I like that the cursor names match layout.Direction, and there's a slight advantage in using names commonly known from JS. WDYT, @pierrec?

eliasnaur:
Either way, the patch needs rebasing.

pierrec:
@eliasnaur @Inkeliz I also like the names matching Direction.

eliasnaur:
Thanks, Pierre. @Inkeliz please rebase your change and I'll merge it.

eliasnaur:
Thanks, merged.

PR #81, opened 2022-03-06
app: [windows] fix maximized window size

inkeliz:
Previously, the window size is equal to the screen size. That doesn't
consider the size of the taskbar, causing the height be bigger than the
real height. Now, the maximized have the same behavior of windowed,
since both of them must include decorations and taskbar.


inkeliz:
The test fail due to Android SDK. I think need to bump the SDK version to 31, but that is unrelated to the current change.

PR #82, opened 2022-03-06
widget: add ImageSequences and GIF support

inkeliz:
That patch adds a easy way to use GIF (or any other kind of images
sequences). That patch adds a new function NewGIF, which is similar
to NewIcon, and adds a ImageSequences similar to the existing Image.

Fixes: https://todo.sr.ht/~eliasnaur/gio/368

### DIFF
--- a/widget/image.go
+++ b/widget/image.go
@@ -4,6 +4,10 @@ package widget
 
 import (
 	"image"
+	"image/draw"
+	"image/gif"
+	"io"
+	"time"
 
 	"gioui.org/f32"
 	"gioui.org/layout"
@@ -29,7 +33,10 @@ type Image struct {
 	Scale float32
 }
 
-const defaultScale = float32(160.0 / 72.0)
+const (
+	defaultScale = float32(160.0 / 72.0)
+	defaultDelay = 16 * time.Millisecond
+)
 
 func (im Image) Layout(gtx layout.Context) layout.Dimensions {
 	scale := im.Scale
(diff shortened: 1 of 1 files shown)


widget/image.go
```diff
@@ -53,3 +60,143 @@ func (im Image) Layout(gtx layout.Context) layout.Dimensions {
...
+			im.FramesEnds[i] = delay
+			if i > 0 {
+				im.FramesEnds[i] += im.FramesEnds[i-1]
+			}
+		}
+	}
+
+	// Find next frame (maybe skip if necessary)
+	for ; frame < len(im.Frames) && im.elapsed > im.FramesEnds[frame]; frame++ {
+	}
```
inkeliz, line 177: Maybe too hacky? 

Maybe change to:

```
for i := 0 ; i < len(im.Frames); i++ {
if im.FramesEnds[frame] >  im.elapsed {
break
}
frame++
}
```

widget/image.go
```diff
@@ -53,3 +60,143 @@ func (im Image) Layout(gtx layout.Context) layout.Dimensions {
...
+	// from the previous one and only one will be drawed per frame
+	Frames []paint.ImageOp
+	// FramesEnds control when each frame ends. The FramesEnds must
+	// have the lenght of Frames.
+	FramesEnds []time.Duration
+	// LoopCount controls the number of times an animation will be
+	// restarted during display.
+	// A LoopCount of 0 means to loop forever.
+	// A LoopCount of -1 means to show each frame only once.
+	// Otherwise, the animation is looped LoopCount+1 times.
```
inkeliz, line 81: That `-1`, `0` and them `+1` is confusing. That comes from `image/gif`, to I keep the same.

widget/image.go
```diff
@@ -53,3 +60,143 @@ func (im Image) Layout(gtx layout.Context) layout.Dimensions {
...
+	Fit Fit
+	// Position specifies where to position the image within
+	// the constraints.
+	Position layout.Direction
+	// Frames is the sequence images, each image must be independent
+	// from the previous one and only one will be drawed per frame
+	Frames []paint.ImageOp
+	// FramesEnds control when each frame ends. The FramesEnds must
+	// have the lenght of Frames.
+	FramesEnds []time.Duration
```
inkeliz, line 76: The `image/gif` uses `Delay` for each frame. I choose to create some kind of `FramesEnds` to make easier to skip frames, in case Gio can't create frames as fast as the delay.

However, I think it could have a "fast path" for constant delays. Consider something like `ConstantDelay`. That will remove the slice and each "End" could be calculated by  `frame * delay`. Most GIFs uses the same "Delay" for all frame, in that case the slice seems overkill.

eliasnaur:
Cool feature, but I'm not sure it belongs in core. @whereswaldon, thoughts?

whereswaldon:
I suspect that many applications will want this feature (the ability to display animated images). It's common across media applications, games, social applications, and more. I can see the argument that we should put it in `x` to try to get the API right, but I don't strongly feel that we need to. I think it probably does eventually belong in core. Maybe it's more a question of "when".

whereswaldon:
I suspect that many applications will want this feature (the ability to display animated images). It's common across media applications, games, social applications, and more. I can see the argument that we should put it in `x` to try to get the API right, but I don't strongly feel that we need to. I think it probably does eventually belong in core. Maybe it's more a question of "when".

eliasnaur:
This may end up in core, but it also feels more "util"-like than the fundamental widgets in package `widget`: it is built entirely on top of widget.Image and it takes an `io.Reader` (by necessity) instead of an image.Image.

Regardless of where it ends up, I suggest a `video` or `media` package that can accept anything through its New constructor:

```
package media

type Image struct {
   // Maybe Fit and Position and Scale.
}

// New constructs a new Image from its contents.
// Supported image formats: JPEG, GIF, PNG, ... (maybe short .avi/.vp9/... clips in future)
func New(io.Reader) (Image, error)

func (img *Image) Layout(gtx C) D 
```

PR #83, opened 2022-03-15
gpu: prevent texture to be larger than MaxTextureSize

inkeliz:
Before that change, the size of the texture could be higher than
the maximum supported by Vulkan.

Fixes: https://todo.sr.ht/~eliasnaur/gio/380


eliasnaur:
Thanks, merged.

PR #85, opened 2022-04-22
app: [Android] improve keyboard hints

inkeliz:
This patch adds support for the following KeyboardHint: Text, Email,
Telephone and URL. Also, fixes crashes related to Samsung keyboards.

Fixes: https://todo.sr.ht/~eliasnaur/gio/404

### DIFF
--- a/app/os_android.go
+++ b/app/os_android.go
@@ -1128,20 +1128,34 @@ func (w *window) ShowTextInput(show bool) {
 func (w *window) SetInputHint(mode key.InputHint) {
 	// Constants defined at https://developer.android.com/reference/android/text/InputType.
 	const (
-		TYPE_NULL                            = 0
-		TYPE_CLASS_TEXT                      = 1
-		TYPE_CLASS_NUMBER                    = 2
-		TYPE_NUMBER_FLAG_DECIMAL             = 8192
-		TYPE_NUMBER_FLAG_SIGNED              = 4096
-		TYPE_TEXT_FLAG_NO_SUGGESTIONS        = 524288
-		TYPE_TEXT_VARIATION_VISIBLE_PASSWORD = 144
+		TYPE_NULL = 0
+
+		TYPE_CLASS_TEXT                   = 1
+		TYPE_TEXT_VARIATION_EMAIL_ADDRESS = 32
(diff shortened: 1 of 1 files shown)


app/GioView.java
```diff
@@ -554,8 +554,11 @@ private class GioInputConnection implements InputConnection {
 
 		@Override public int getCursorCapsMode(int reqModes) {
 			Snippet snip = getSnippet();
-			int selStart = imeSelectionStart(nhandle);
-			return TextUtils.getCapsMode(snip.snippet, imeToUTF16(nhandle, selStart), reqModes);
+			int selStart = imeToUTF16(nhandle, imeSelectionStart(nhandle));
+			if (selStart > snip.snippet.length()) {
+			    selStart = 0;
```
eliasnaur, line 559: What if `snip.snippet` is empty? I suggest just `return 0` here, because the caps mode is undefined at positions outside the snippet.

elias-orijtech: Thinking a bit more, we control `selStart,` so why is it ever greater than `snip.snippet.length`? Something must be updating either the selection or the snippet without adjusting the other.

eliasnaur: Thinking a bit more, we control selStart, so why is it ever greater than snip.snippet.length? Something must be updating either the selection or the snippet without adjusting the other.

app/os_android.go
```diff
@@ -1128,20 +1128,34 @@ func (w *window) ShowTextInput(show bool) {
 func (w *window) SetInputHint(mode key.InputHint) {
 	// Constants defined at https://developer.android.com/reference/android/text/InputType.
 	const (
-		TYPE_NULL                            = 0
```
eliasnaur, line 1131: Nit: the new hints should be a separate commit.

eliasnaur:
> Fixes: https://todo.sr.ht/~eliasnaur/gio/404
> 

Delete newline after the "Fixes:" line, otherwise sr.ht won't recognize it.

app/GioView.java
elias-orijtech: Fixed in tip.

inkeliz:
I remove the crash stuff, since it was already fixed.

eliasnaur:
Thanks, merged.

PR #86, opened 2022-05-29
app: [windows] fix keyboard focus

inkeliz:
On Windows, it's required to SetFocus to receive keyboard events.
Previously, Gio doesn't set focus on the current window. In some
edge cases, it prevents Gio from getting keyboard events.

That is more noticeable when spawning children's windows or parent
windows. In such a case, the child window can steal keyboard focus,
and Gio will never recover it. Now, it will retrieve the focus when
using key.FocusOp.

### DIFF
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -53,6 +53,7 @@ type window struct {
 	placement *windows.WindowPlacement
 
 	animating bool
+	focused   bool
 
 	deltas     winDeltas
 	borderSize image.Point
@@ -277,8 +278,10 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 			Type: pointer.Cancel,
 		})
 	case windows.WM_SETFOCUS:
+		w.focused = true
 		w.w.Event(key.FocusEvent{Focus: true})
 	case windows.WM_KILLFOCUS:
+		w.focused = false
 		w.w.Event(key.FocusEvent{Focus: false})
 	case windows.WM_NCHITTEST:
 		if w.config.Decorated {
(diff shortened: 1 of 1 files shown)


gedw99:
i got hit by this too. thanks for working it out @inkeliz

app/os_windows.go
```diff
@@ -749,7 +749,13 @@ func loadCursor(cursor pointer.Cursor) (syscall.Handle, error) {
 	}
 }
 
-func (w *window) ShowTextInput(show bool) {}
+func (w *window) ShowTextInput(show bool) {
+	if show {
+		windows.SetFocus(w.hwnd)
+	} else {
+		windows.SetFocus(0)
```
eliasnaur, line 756: Why force focus away from the window when `show` is false? On the other platforms dismissing software text input doesn't imply a focus loss, does it?

eliasnaur (commented):
You call SetFocus as a result of a SoftKeyboardOp, not a FocusOp, right? That's not what your change message describes.

Besides, I'm surprised clicking on the window won't automatically bring focus to it. Perhaps SetFocus should be called whenever we're clicked and don't have focus (on Windows)?

inkeliz:
Yes, it makes more sense to focus when click.

eliasnaur:
Just to be sure: this is still an issue even after https://git.sr.ht/~eliasnaur/gio/commit/cbbb5865e5c0402219bb74fedb46cc620811f61a ?

inkeliz:
That issue also affects MacOS. So, the issue might affect all OSes, since Gio was not initially design to work with others Views/Windows as children. 

I manage to fix MacOS with similar code:

```
static void setFocus(CFTypeRef windowRef, CFTypeRef viewRef) {
    NSWindow *window = (__bridge NSWindow *)windowRef;
    NSView *view = (__bridge NSView *)viewRef;
    [window makeFirstResponder:view];
}
``` 

And then calling it when:

	if typ == pointer.Press {
		C.setFocus(w.window, w.view)
	}

What happens on MacOS is different (but similar): the focus is recovered once I click somewhere else on the WebView (if I click on the "background", it the HTML Input loses focus). So MacOS is smart enough to only require focus once needed. However, if you click anywhere on Gio (including text fields), the focus is not removed. The code above fixes that issue, similar to how it fixes the issue on Windows.

----------

cc @hanswered since he seems to be working with multiple window too. :)

eliasnaur:
Merged, thanks!

PR #87, opened 2022-06-27
fix build ios ,prev declared but not used

never-afk:


never-afk:
`gogio -o kitchen.app -target ios    gioui.org/example/kitchen` 
show error `app/os_ios.go:281:2: prev declared but not used`

eliasnaur:
Thanks, merged.

PR #88, opened 2022-07-03
app: [ios] add ignoreSafeAreaInsets config (iOS >= 11.*)

afansv:
By default status bar on iPhones with notch is white (bgColor).
I think we need to add an option to allow applications control unsafe areas themself.

<table>
  <tr>
    <th>Before</th>
    <th>After</th> 
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### DIFF
--- a/app/os.go
+++ b/app/os.go
@@ -43,6 +43,8 @@ type Config struct {
 	CustomRenderer bool
 	// Decorated reports whether window decorations are provided automatically.
 	Decorated bool
+	// Ignore safe area insets for iOS >= 11.
+	IgnoreSafeAreaInsets bool
 }
 
 // ConfigEvent is sent whenever the configuration of a Window changes.
--- a/app/os_ios.go
+++ b/app/os_ios.go
@@ -46,7 +46,7 @@ static void hideTextInput(CFTypeRef viewRef) {
 	[view resignFirstResponder];
 }
 
-static struct drawParams viewDrawParams(CFTypeRef viewRef) {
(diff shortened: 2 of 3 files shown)


inkeliz:
Im not sure if that is required. You can overwrite the insets on system.FrameEvents, before give it to layout.NewContext. Something similar to:
```

case system.FrameEvent:
e.Insets = system.Insets{} // Remove Insets

gtx := layout.NewContext(ops, e)
```

Does it fixes the issue?

afansv:
> Im not sure if that is required. You can overwrite the insets on system.FrameEvents, before give it to layout.NewContext. Something similar to:
> 
> ```
> switch e := evt.(type) {
> case system.FrameEvent:
>    e.Insets = system.Insets{} // Remove Insets
> 
>    gtx := layout.NewContext(ops, e) // It will not use margins, because the event doesn't have it
> }
> ```
> 
> Does it fixes the issue?

Yes, it's better this way. Anyway, I am trying to understand one thing i came across while working on this PR:

```
UIEdgeInsets insets = v.layoutMargins;
```
Even when run on iPhone without notch, this code just gives me 0. So why we need to this obj-c call in this place? Just to use safeAreaInsets by default (when it's necessary)

inkeliz:
I'm not sure. But maybe that was useful for older versions of iOS, which doesn't support safeAreaInsets.

However, I think the minimum iOS version supported by Golang is now iOS 12. If that is right, we can remove that "fallback" and only use the safeAreaInsets.

PR #89, opened 2022-07-19
app: [android] use GioView  inside FrameLayout

inkeliz:
Before that change, on Android, was impossible to overlay GioView with
a custom view. This change adds FrameLayout and renders GioView into
that, allowing to use `addView` from Android API.

This change also adds `Layer` on ViewEvent, which represents the
FrameLayout.

Fixes: https://todo.sr.ht/~eliasnaur/gio/427

### DIFF
--- a/app/GioActivity.java
+++ b/app/GioActivity.java
@@ -5,16 +5,30 @@ package org.gioui;
 import android.app.Activity;
 import android.os.Bundle;
 import android.content.res.Configuration;
+import android.view.ViewGroup;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.FrameLayout;
 
 public final class GioActivity extends Activity {
 	private GioView view;
+	public FrameLayout layer;
 
 	@Override public void onCreate(Bundle state) {
-		super.onCreate(state);
+            super.onCreate(state);
 
-		this.view = new GioView(this);
+            layer = new FrameLayout(this);
+            view = new GioView(this);
 
-		setContentView(view);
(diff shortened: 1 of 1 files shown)


inkeliz:
I'm using the name `Layer` to match against MacOS:

https://github.com/gioui/gio/blob/d8766f6d2ddf670a9fb8fa9828aa55c656966a28/app/os_macos.go#L233-L234

However, that may sound confusing, since Android isn't "Layer", but "FrameLayout". However, saying `Frame` or `Layout` sounds weird to. However, that is just the name. 😎

app/os_android.go
```diff
@@ -203,6 +203,12 @@ type ViewEvent struct {
 	// the next ViewEvent is received.
 	// A zero View means that there is currently no view attached.
 	View uintptr
+
+	// Layer is a JNI global reference to the android.widget.FrameLayout
+	// instance which holds View and others views. The reference is
+	// valid until the next ViewEvent is received.
+	// A zero Layer means that there is currently no layer attached.
+	Layer uintptr
```
eliasnaur, line 211: Why is https://developer.android.com/reference/android/view/View#getParent() not enpough for clients?

inkeliz: I add it to make easier. If you need to provide one View to an third-party library, you can use `Layer` directly.  Otherwise, you need to write your own Java code to call `getParent()` and then provide it to the another library.

eliasnaur: How can you pass a View to the third-party without writing Java code?

eliasnaur:
Thanks, merged.

PR #91, opened 2022-07-23
layout: Add convenience method LayoutW to wrap Layout

emad-elsaid:
This was discussed on slack
https://gophers.slack.com/archives/CM87SNCGM/p1658574692395529

The idea is to make the Layout method more interop with other widgets that needs
other widgets as children.

# Slack original message: 

while using Gio I found that using layouts as children for other layouts or passing a layout as widget to another Layout function needs me to wrap it in a closure. as follows

> Flex{}.Layout(ctx, Rigid(func(c Context) Dimensions { return Flex{}.Layout(ctx, children...) }))

While if the Flex.Layout method has this signature func (f Flex) Layout(children ...FlexChild) func(gtx Context) Dimensions) then the usage can be simpler

> Flex{}.Layout(ctx, Rigid(Flex{}.Layout(children...)))

and that goes to any method that takes Context the more params. if it's changed to a function that takes the more params and return a function of `layout.Widget` it would make it easier to use with anything that needs a widget.

in my code I'm doing these wrappers (it's easy to do) but I thought that can make it easier for everyone to change it in future releases of Gio. 

# Changes
- This introduced another method `LayoutW` W for Widget.
- The new method takes all parameters of `Layout` minus `Context` and returns a Widget function.
- Tests for Stack layout and Flex layout are replicated to use `LayoutW` to make sure there are no allocations for the new approach.

### DIFF
--- a/layout/alloc_test.go
+++ b/layout/alloc_test.go
@@ -30,6 +30,23 @@ func TestStackAllocs(t *testing.T) {
 	}
 }
 
+func TestStackLayoutWAllocs(t *testing.T) {
+	var ops op.Ops
+	allocs := testing.AllocsPerRun(1, func() {
+		ops.Reset()
+		gtx := Context{
+			Ops: &ops,
+		}
+		Stack{}.LayoutW(Stacked(func(gtx Context) Dimensions {
+			return Dimensions{Size: image.Point{X: 50, Y: 50}}
+		}),
+		)(gtx)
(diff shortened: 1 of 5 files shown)


eliasnaur:
I'm considering a larger redesign of Gio layout, so I'm putting this on hold for a bit. Otherwise, looks like a neat addition.

PR #92, opened 2022-07-25
app,io/system: [API] add StageInactive when window is not in focus

inkeliz:
Now, Gio will notify when window focus changes.

### DIFF
--- a/app/os_android.go
+++ b/app/os_android.go
@@ -552,7 +552,7 @@ func Java_org_gioui_GioView_onLowMemory(env *C.JNIEnv, class C.jclass) {
 func Java_org_gioui_GioView_onConfigurationChanged(env *C.JNIEnv, class C.jclass, view C.jlong) {
 	w := cgo.Handle(view).Value().(*window)
 	w.loadConfig(env, class)
-	if w.stage >= system.StageRunning {
+	if w.stage >= system.StageInactive {
 		w.draw(env, true)
 	}
 }
@@ -563,7 +563,7 @@ func Java_org_gioui_GioView_onFrameCallback(env *C.JNIEnv, class C.jclass, view 
 	if !exist {
 		return
 	}
-	if w.stage < system.StageRunning {
+	if w.stage < system.StageInactive {
 		return
 	}
 	if w.animating {
@@ -596,7 +596,7 @@ func Java_org_gioui_GioView_onWindowInsets(env *C.JNIEnv, class C.jclass, view C
 		left:   int(left),
 		right:  int(right),
 	}
-	if w.stage >= system.StageRunning {
+	if w.stage >= system.StageInactive {
 		w.draw(env, true)
 	}
 }
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -283,6 +283,14 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 	case windows.WM_KILLFOCUS:
 		w.focused = false
 		w.w.Event(key.FocusEvent{Focus: false})
+	case windows.WM_NCACTIVATE:
+		if w.stage >= system.StageInactive {
+			if wParam == windows.TRUE {
+				w.setStage(system.StageRunning)
+			} else {
+				w.setStage(system.StageInactive)
+			}
+		}
 	case windows.WM_NCHITTEST:
 		if w.config.Decorated {
 			// Let the system handle it.
--- a/app/window.go
+++ b/app/window.go
@@ -426,7 +426,7 @@ func (w *Window) driverDefer(f func(d driver)) {
 
 func (w *Window) updateAnimation(d driver) {
 	animate := false
-	if w.stage >= system.StageRunning && w.hasNextFrame {
+	if w.stage >= system.StageInactive && w.hasNextFrame {
 		if dt := time.Until(w.nextFrame); dt <= 0 {
 			animate = true
 		} else {
@@ -826,7 +826,7 @@ func (w *Window) processEvent(d driver, e event.Event) bool {
 	}
 	switch e2 := e.(type) {
 	case system.StageEvent:
-		if e2.Stage < system.StageRunning {
+		if e2.Stage < system.StageInactive {
 			if w.gpu != nil {
 				w.ctx.Lock()
 				w.gpu.Release()
@@ -842,7 +842,7 @@ func (w *Window) processEvent(d driver, e event.Event) bool {
 		if e2.Size == (image.Point{}) {
 			panic(errors.New("internal error: zero-sized Draw"))
 		}
-		if w.stage < system.StageRunning {
+		if w.stage < system.StageInactive {
 			// No drawing if not visible.
 			break
 		}
--- a/io/system/system.go
+++ b/io/system/system.go
@@ -59,17 +59,24 @@ type StageEvent struct {
 type Stage uint8
 
 const (
-	// StagePaused is the Stage for inactive Windows.
-	// Inactive Windows don't receive FrameEvents.
+	// StagePaused is the stage for windows that have no on-screen representation.
+	// Paused windows don't receive FrameEvent.
 	StagePaused Stage = iota
-	// StateRunning is for active Windows.
+	// StageInactive is the stage for windows that are visible, but not active.
+	// Inactive windows receive FrameEvent.
+	StageInactive
+	// StageRunning is for active and visible Windows.
+	// Running windows receive FrameEvent.
 	StageRunning
 )
 
+// String implements fmt.Stringer.
 func (l Stage) String() string {
 	switch l {
 	case StagePaused:
 		return "StagePaused"
+	case StageInactive:
+		return "StageInactive"
 	case StageRunning:
 		return "StageRunning"
 	default:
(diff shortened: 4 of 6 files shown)


inkeliz:
Currently, the behavior of "focus" is defined by the OS, which doesn't match very well. On Windows, the "focus" is lost when some children is focused, on MacOS it's not, but there's another event for that case. However, the FocusEvent can be improved once Gio have better support for children windows.

app/os_x11.go
```diff
@@ -619,8 +619,10 @@ func (h *x11EventHandler) handleEvents() bool {
 			redraw = (*C.XExposeEvent)(unsafe.Pointer(xev)).count == 0
 		case C.FocusIn:
 			w.w.Event(key.FocusEvent{Focus: true})
+			w.w.Event(system.FocusEvent{Focus: true})
```
eliasnaur, line 622: Everywhere you send a system.FocusEvent when a key.FocusEvent is also sent. Why is key.FocusEvent not enough?

whereswaldon: `key.FocusEvent` is going to be consumed by the foremost key handler, which may be deeply nested within application layout code. Wanting to change a window behavior based on the window focus state is a much higher-level concern. One user asked for this focus detection feature in order to build a dmenu-esque application launcher. If the window loses focus, they want to close the window. Doing that by interrogating the state of every widget to see if any of them are focused doesn't seem like the right design.

app/os_windows.go
```diff
@@ -280,6 +281,8 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 		w.w.Event(key.FocusEvent{Focus: true})
 	case windows.WM_KILLFOCUS:
 		w.w.Event(key.FocusEvent{Focus: false})
+	case windows.WM_NCACTIVATE:
+		w.w.Event(system.FocusEvent{Focus: wParam == windows.TRUE})
```
inkeliz, line 285: On Windows, the `KILLFOCUS` happens when another window have keyboard focus, which might also happens when some child window is in focus. The `NCACTIVATE` sounds better, since it report when the window decoration should update.

app/os_x11.go
inkeliz: I don't use key.FocusEvent because of the reason mentioned by @whereswaldon. Also, the FocusEvent can work differently from key.FocusEvent. 

I force-push to make the behavior differently. Now, the FocusEvent is only triggered when the window loses focus, instead of the keyboard focus. The word "focus" could be changed to "active", or something else.

eliasnaur:
> Currently, the behavior of "focus" is defined by the OS, which doesn't match very well. On Windows, the "focus" is lost when some children is focused, on MacOS it's not, but there's another event for that case. However, the FocusEvent can be improved once Gio have better support for children windows.

This difference between macOS and WIndows is fixed by https://github.com/gioui/gio/pull/93, right?



> I don't use key.FocusEvent because of the reason mentioned by @whereswaldon. Also, the FocusEvent can work differently from key.FocusEvent.

> I force-push to make the behavior differently. Now, the FocusEvent is only triggered when the window loses focus, instead of the keyboard focus. The word "focus" could be changed to "active", or something else.

Thanks for explaining. I agree that it would be useful to know the active state of an app.Window, separate from the focus. However, I'm not sure a separate event is necessary. Why not expand `io/system.Stage` with a `StageInactive` and rename `StageRunning` to `StageActive`?

inkeliz:
I initially thought about adding a new `Stage`, but on https://lists.sr.ht/~eliasnaur/gio/%3C96373f04-28fe-6848-25f4-2d9936aa0522%40gmail.com%3E#%3CCAMAFT9VY22-baCO+OyzM+cGH7ktVU08RvgKQOTrbAtAO91beWg@mail.gmail.com%3E, you mention to addit as a new `StageEvent`.

However, I'm not sure how the new Stage will work, because we have some checks: `if stage < system.StageRunning`:

https://github.com/gioui/gio/blob/b67bef3e0d961e1c04eb4f9f449903a625664d5b/app/window.go#L828-L835

--------

Currently, the `Stage` is either "Running" or "Inactive":

https://github.com/gioui/gio/blob/b5f12c5f2627dc5d557bb0f8ffc85a1a69258cfa/io/system/system.go#L61-L67

But, in the case of "not in focus" it's both of them, at same time. It's inactive at the perspective of the user, but it's active because it can generate new frames (it still displayed on the screen).

eliasnaur:
> I initially thought about adding a new `Stage`, but on https://lists.sr.ht/~eliasnaur/gio/%3C96373f04-28fe-6848-25f4-2d9936aa0522%40gmail.com%3E#%3CCAMAFT9VY22-baCO+OyzM+cGH7ktVU08RvgKQOTrbAtAO91beWg@mail.gmail.com%3E, you mention to add that as a new `FocusEventt`.
> 

FocusEvent could work if we can't cram the focused/active state into Stage.

> However, I'm not sure how the new Stage will work, because we have some checks: `if stage < system.StageRunning`:
> 
> https://github.com/gioui/gio/blob/b67bef3e0d961e1c04eb4f9f449903a625664d5b/app/window.go#L828-L835
>

This check (and similar) should change to `e2.Stage == system.StagePaused`.
 
> Currently, the `Stage` is either "Running" or "Inactive":
> 
> https://github.com/gioui/gio/blob/b5f12c5f2627dc5d557bb0f8ffc85a1a69258cfa/io/system/system.go#L61-L67
> 
> But, in the case of "not in focus" it's both of them, at same time. It's inactive at the perspective of the user, but it's active because it can generate new frames (it still displayed on the screen).
> 
> I don't think that it makes sense but maybe we can use `1 << iota +1` and then use `StageRunning | StageInactive`? So, it's running (gio is generating frames), but it's inactive (the window is not in focus).

Stage was designed to be an integer of increasing "liveness" of the window. So for active, inactive:

```
 const ( 
 	// StagePaused is the stage for windows that have no on-screen representation. 
 	// Paused Windows don't receive FrameEvents. 
 	StagePaused Stage = iota
        // StageInactive is the stage for windows that are visible, but not active.
        StageInactive
        // Inactive windows receive FrameEvents.
 	// StateRunning is for active and visible Windows.
        // Active windows receive FrameEvents. 
 	StageRunning 
 ) 
```

io/system/system.go
```diff
@@ -59,24 +59,37 @@ type StageEvent struct {
...
 		panic("unexpected Stage value")
 	}
 }
 
+// IsReceivingFrames returns true if the window is receiving
+// FrameEvent.
+func (l Stage) IsReceivingFrames() bool {
+	return l >= StageInactive
+}
+
```
eliasnaur, line 92: What's the use case(s) for exporting IsReceivingFrames? It seems to me it shouldn't be exported.

io/system/system.go
```diff
@@ -59,24 +59,37 @@ type StageEvent struct {
...
 
 const (
-	// StagePaused is the Stage for inactive Windows.
-	// Inactive Windows don't receive FrameEvents.
+	// StagePaused is the stage for windows that have no on-screen representation.
+	// Paused windows don't receive FrameEvent.
 	StagePaused Stage = iota
-	// StateRunning is for active Windows.
+	// StageInactive is the stage for windows that are visible, but not active.
+	// Inactive windows receive FrameEvent.
```
eliasnaur, line 66: Ditto.

io/system/system.go
inkeliz: I think the use-case is to replace any `if stage < system.StageRunning`. If new Stage is added in the future, the `IsReceivingFrames()` can be updated internally, without  need to change every `if stage < system.StageRunning`:

https://github.com/gioui/gio/blob/a1c783812504da29db3df7489e453d7c8bb62e7d/app/window.go#L426-L431

It can't be private, but that lives on `io/system` and it's used by `app`. I can remove it and replace everything to `if stage != system.StagePaused` or  `stage >= system.StageInactive`, I just thought that it will be harder to understand and maintain.

eliasnaur: Adding new stages is a major addition and code that uses the stage should be inspected (for possible updating). In other words, we haven't added new stages for a long time (years?) and I don't expect we will for a long time (if ever).

app/window.go
```diff
@@ -842,7 +842,7 @@ func (w *Window) processEvent(d driver, e event.Event) bool {
 		if e2.Size == (image.Point{}) {
 			panic(errors.New("internal error: zero-sized Draw"))
 		}
-		if w.stage < system.StageRunning {
+		if w.stage == system.StagePaused {
```
eliasnaur, line 845: Ditto.

app/os_android.go
```diff
@@ -552,7 +552,7 @@ func Java_org_gioui_GioView_onLowMemory(env *C.JNIEnv, class C.jclass) {
 func Java_org_gioui_GioView_onConfigurationChanged(env *C.JNIEnv, class C.jclass, view C.jlong) {
 	w := cgo.Handle(view).Value().(*window)
 	w.loadConfig(env, class)
-	if w.stage >= system.StageRunning {
+	if w.stage > system.StagePaused {
```
eliasnaur, line 555: Nit: it seems a bit more clear to replace with `w.stage >= system.StageInactive`. Here and elsewhere.

inkeliz:
Sorry for a lot of force pushes, but there's no change between them. That fixed comparison, as you suggest.

eliasnaur:
Merged, thanks!

PR #93, opened 2022-07-25
app: [macos] set focus on click

inkeliz:


eliasnaur:
Can you describe what this PR fixes, preferably in the commit message?

app/os_macos.go
```diff
@@ -539,6 +546,11 @@ func gio_onMouse(view, evt C.CFTypeRef, cdir C.int, cbtns C.NSUInteger, x, y, dx
 	if cbtns&(1<<2) != 0 {
 		btns |= pointer.ButtonTertiary
 	}
+
+	if typ == pointer.Press {
+		C.setFocus(w.window, w.view)
```
eliasnaur, line 551: Can you track the focused state and only call setFocus if not focused? Similar to your Windows PR #86.

inkeliz: It's a little bit different than Windows, I implemented the `becomeFirstResponder` (https://developer.apple.com/documentation/uikit/uiresponder/1621113-becomefirstresponder).

Since it was introduced, I also include `key.FocusEvent` here (the system.FocusEvent should not be triggered in that case), that may happens when another window (inside the Gio window) gets the "Focus", the `windowDidBecomeKey` isn't triggered in those cases.

app/os_macos.go
```diff
@@ -563,6 +574,16 @@ func gio_onFocus(view C.CFTypeRef, focus C.int) {
...
 }
 
+//export gio_onResponder
+func gio_onResponder(view C.CFTypeRef, first C.int) {
+	// That function is called when some child view becomes first responder.
+	w := mustView(view)
+	w.focused = first == 1
+	if w.w.d != nil {
+		w.w.Event(key.FocusEvent{Focus: w.focused})
+	}
```
inkeliz, line 584: Maybe unrelated, add it as new PR?

app/os_macos.go
```diff
@@ -563,6 +574,16 @@ func gio_onFocus(view C.CFTypeRef, focus C.int) {
 	w.SetCursor(w.cursor)
 }
 
+//export gio_onResponder
+func gio_onResponder(view C.CFTypeRef, first C.int) {
```
eliasnaur, line 578: What about `gio_onFocus`, `windowDidBecomeKey`, `windowDidResignKey`? It seems to me they should be removed by this change. If not, why not?

app/os_macos.go
```diff
@@ -563,6 +574,16 @@ func gio_onFocus(view C.CFTypeRef, focus C.int) {
...
+	// That function is called when some child view becomes first responder.
+	w := mustView(view)
+	w.focused = first == 1
+	if w.w.d != nil {
```
eliasnaur, line 582: We don't check `w.w.d` for nil elsewhere, why here? If it's really necessary, it smells like a different problem, perhaps that `SetDriver` is called too late.

inkeliz: Without that check it crashes. I'm not sure exactly why. I also tried to wrapper it into `w.run()` (the main thread stuff), and still crashing.

app/os_macos.go
inkeliz: I think each one fits one purpose. The Responder seems to be related to the current window, and Key is global. In the end:

- If you minimize the window: the ResignKey is triggered, but resignFirstResponder is not.
- If you click in another child window: the ResignKey is not triggered, but the resignFirstResponder will be triggered.

Seems that in some conditions, both will notify.

app/os_macos.go
eliasnaur: What's the crash dump? Maybe we can figure out why.

inkeliz: I need to check again, but that is something like: "sending events before drivers is ready".

app/os_macos.go
eliasnaur: > I think each one fits one purpose. The Responder seems to be related to the current window, and Key is global. In the end:
> 
>     * If you minimize the window: the ResignKey is triggered, but resignFirstResponder is not.
> 
>     * If you click in another child window: the ResignKey is not triggered, but the resignFirstResponder will be triggered.
> 
> 
> Seems that in some conditions, both will notify.

So in total, Key is what we want for StageInactive, and Responder is what we want to use for focus, right?

app/os_macos.go
eliasnaur: I thought so, which is why I commented "perhaps the SetDriver is called too late". You can try moving that call earlier, or you may have to construct a window in a two-step process: first create the window, then make it visible/key/whatever.

eliasnaur:
Is this relevant again? If so, what's the connection to https://github.com/gioui/gio/pull/138, if any?

PR #94, opened 2022-07-29
app: [android] fix insets

inkeliz:
Previously, the Inset could be report wrongly when the bottom inset
is smaller than the top. Now, it uses `image.Rectangle` directly,
to bypass `image.Rect` checks.

### DIFF
--- a/app/os_android.go
+++ b/app/os_android.go
@@ -154,7 +154,7 @@ type window struct {
 
 	dpi       int
 	fontScale float32
-	insets    image.Rectangle
+	insets    pixelInsets
 
 	stage     system.Stage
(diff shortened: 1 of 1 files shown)


inkeliz:
(sounds like the tests got stuck)

eliasnaur:
> I force-push to changing to `system.Inset`. However, that might be confusing, since it's not truly `unit.Dp`. I tried to move the dpi calculation, but doesn't work. :S

You're right, of course. In that case, a better solution is a custom type, say `pixelInsets`. Sorry for the back-and-forth.

eliasnaur:
Thanks, merged!

PR #96, opened 2022-08-11
Export Undo/Redo functions

FabienJansem:
Export Undo/Redo functions
Add History and SetHistory functions

Fixes: https://todo.sr.ht/~eliasnaur/gio/438
Signed-off-by: Fabien Jansem <fabien@jansem.eu.org>


whereswaldon (approved):
I think this looks acceptable. It exports History as an opaque value, allowing us to delay committing to any particular set of fields and values. This will prevent applications from saving and restoring editor history across restarts though, as they can't serialize this value. Of course, the final decision belongs to @eliasnaur .

FabienJansem:
> It exports History as an opaque value, allowing us to delay committing to any particular set of fields and values

That was the idea
> This will prevent applications from saving and restoring editor history across restarts

Right, but I haven't found a use case for it so far.

eliasnaur (commented):
It's almost certainly a good idea to expose the change history for Editor, however the opaque `History` type and 4 methods seems unfortunate. What would it take to get rid of History.nextIdx and replace `History` with `[]Modification`? I haven't looked at the code, but it seems to me `Modification` could be made self-contained.

FabienJansem:
> It's almost certainly a good idea to expose the change history for Editor, however the opaque `History` type and 4 methods seems unfortunate. What would it take to get rid of History.nextIdx and replace `History` with `[]Modification`? I haven't looked at the code, but it seems to me `Modification` could be made self-contained.

NextIdx can be moved into Modification. This would give
```
type Modification struct {
	data []struct {
		StartRune int
		ApplyContent string
		ReverseContent string
	}
	nextIdx int
}
```
If we delete `History()` and `SetHistory()`, we have to make all the content visible and this would leave the deep copy to the caller. But we can also move them into `Modification` and call them like this:
```
type Editor struct {
	...
	History Modification
}

backup := ed.History.Get()
ed.History.Set(backup)
```
What do you think?

FabienJansem:
Will propose another way

PR #97, opened 2022-09-04
Export Undo/Redo functions

FabienJansem:
History is now self-contained


FabienJansem:
Will provide another way

PR #100, opened 2022-09-04
Export Undo/Redo functions

FabienJansem:
Export Undo/Redo functions
Editor.History is available to save/restore modification history

Fixes: https://todo.sr.ht/~eliasnaur/gio/438
Signed-off-by: Fabien Jansem fabien@jansem.eu.org


widget/editor.go
```diff
@@ -1223,32 +1222,42 @@ type modification struct {
...
-func (e *Editor) undo() {
-	if len(e.history) < 1 || e.nextHistoryIdx == 0 {
+// History of modifications may be retrived/restored
+type History struct {
+	// data contains undo/redo history.
+	data []modification
+	// nextIdx is the index within the history data of the next modification.
+	// This is only not len(data) immediately after undo operations occur.
+	// It is framed as the "next" value to make the zero value consistent.
+	nextIdx int
```
eliasnaur, line 1232: We're still stuck with `nextIdx` in your proposed change. Why can't `modification` be self-contained? I imagine calling it `Edit` and defined something like:

```
type Edit struct {
    // rng is the range to be replaced with text.
    rng key.Range
    text string
}
```

eliasnaur:
Please keep one PR per change. It's hard to follow the review history across multiple PRs. You can either [squash and force-push](https://stackoverflow.com/questions/14534397/squash-all-my-commits-into-one-for-github-pull-request) or add fixup commits on top of older commits and squash when we're ready to submit your change to main.

widget/editor.go
FabienJansem: I'm sorry, but I don't understand your request.    
nextIdx is the nextHistoryIdx that existed before my PR. I simply moved it to the History structure, which allows it to be saved along with the modifications (I changed the name because History was becoming redundant).    
I don't see how it can be deleted without losing the boundary between undo and redo when undo was the last action performed by the user in the editor. If you don't save it, then when you restore modifications editor will consider that the redo's are undo's, which doesn't correspond to what's in the text. (that's my understanding from the existing comment associated with nextHistoryIdx which I also moved into the history).

Maybe I'm missing something.

whereswaldon: @eliasnaur we need a way to track the boundary between changes that have been applied and changes that can be applied. If you undo three changes, the final three modifications are what should happen when you "redo" three times. That's what the index is for. We could eliminate it by creating two slices (one for undo and one for redo), but we'd be copying elements between them pretty often. Other than that, I don't know how to efficiently eliminate the index.

eliasnaur: Hm, we may have different views of what makes History useful. I've posted on the issue tracker to clarify my understanding of your proposal.

ddkwork:
any information for this?

PR #101, opened 2022-09-15
app: remove gofont.Collection by default

inkeliz:
This change removes `gofont.Collection()`, which imports multiples fonts and increase the binary size.

Fix: https://todo.sr.ht/~eliasnaur/gio/371

### DIFF
--- a/app/window.go
+++ b/app/window.go
@@ -14,7 +14,7 @@ import (
 	"unicode/utf8"
 
 	"gioui.org/f32"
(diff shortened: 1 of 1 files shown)


inkeliz:
I think this change mitigates the issue, but doesn't fixes it. Also, far I notice it's not possible to change the fonts, so sounds like it's locked into `gofont`.

eliasnaur:
Thank you. Merged.

PR #102, opened 2022-09-19
app,io/key: Add key code for key.Event

dennwc:
Add non-portable key code to the `key.Event` type. This allows users to handle platform-specific key codes in the application code. Backends will now always generate input events, even for keys with no well-known names. Previously backends only generated events for keys they recognized.

Signed-off-by: Denys Smirnov <dennwc@pm.me>

### DIFF
--- a/app/internal/xkb/xkb_unix.go
+++ b/app/internal/xkb/xkb_unix.go
@@ -160,19 +160,18 @@ func (x *Context) DispatchKey(keyCode uint32, state key.State) (events []event.E
 		x.utf8Buf = make([]byte, 1)
 	}
 	sym := C.xkb_state_key_get_one_sym(x.state, kc)
-	if name, ok := convertKeysym(sym); ok {
-		cmd := key.Event{
-			Name:      name,
-			Modifiers: x.Modifiers(),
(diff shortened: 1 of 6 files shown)


eliasnaur:
AFAICT, this PR implements `key.Event`s with empty `Name`s for unknown/non-portable key presses. How do you propose a Gio program process these? Ideally, demonstrate with a test in the [`io/router`](`https://git.sr.ht/~eliasnaur/gio/tree/main/item/io/router/key_test.go`) package.

inkeliz:
_I think that change will also help me with https://github.com/gioui/gio-x/pull/8, since gamepad inputs on Android seems to be sent as normal keypresses (far I remember)._ 🥳

asciimoth:
Hi! I'm also interested in this feature, especially to handle hotkeys like Ctrl-C/Ctrl-V same way on different keyboard layouts (en/ru/etc).
May I know why this PR was closed? @dennwc 
Also I'm ready to continue work on this feature myself if it lacks a maintainer.

whereswaldon:
I believe this PR was closed out of lack of interest by the original author, but I don't know for sure.

PR #103, opened 2022-09-20
Add ArrowsNavigation option

dennwc:
Directional navigation with arrow keys/dpad was only enabled on mobile targets (Android, iOS) with no way to enable it for desktop.

This change adds a new `Option` for it, which is useful for game UIs and other apps that allow faster keyboard navigation.

Signed-off-by: Denys Smirnov <dennwc@pm.me>

### DIFF
--- a/app/os.go
+++ b/app/os.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"image"
 	"image/color"
+	"runtime"
 
 	"gioui.org/io/key"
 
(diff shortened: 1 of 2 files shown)


eliasnaur:
This is surely convenient but also very specific to a narrow set of use-cases. It seems better to keep non-platform policy out of package `app`, yet allow the user to control focus. How about a `key.MoveFocusOp` operation taking a direction, similar to `key.FocusOp` for specific tags? Then the user can catch the relevant keys at the top level and move focus itself.

dennwc:
It's a reasonable change, since it only exposes the functionality already available in Gio, just for other platforms. Any embedded use case may benefit from this (since it uses Linux "desktop" with potentially limited navigation).

I think directional navigation is hard enough (there are more bugs apart from [gio/448](https://todo.sr.ht/~eliasnaur/gio/448), will report later), so implementing it on the app side might be an overkill. It's intuitive how arrows should behave, it shouldn't be too different regarding of the app. In case custom behavior is needed, this change still allows setting the option to `false` (even on mobile, it works like an override) and implementing navigation manually.

eliasnaur:
> It's a reasonable change, since it only exposes the functionality already available in Gio, just for other platforms.

As I said, it's tempting to add `ArrowsNavigation`, but I'd rather keep package `app` as non-magic as possible. And `key.MoveFocusOp` doesn't sound too hard to implement, neither on the Gio side or the client side.

> Any embedded use case may benefit from this (since it uses Linux "desktop" with potentially limited navigation).
> 

Gio should auto-detect platforms that need arrow navigation and enable it without user intervention.

> I think directional navigation is hard enough (there are more bugs apart from [gio/448](https://todo.sr.ht/~eliasnaur/gio/448), will report later), so implementing it on the app side might be an overkill. It's intuitive how arrows should behave, it shouldn't be too different regarding of the app. In case custom behavior is needed, this change still allows setting the option to `false` (even on mobile, it works like an override) and implementing navigation manually.

Please file bugs for other issues, especially if you think they matter in deciding this PR.

PR #104, opened 2022-12-06
app: add configurable appid for wayland

mkungla:
Currently building Gio apps on Wayland will show status bar **Unknown** window as active.

Screenshot from 2022-12-06 20-31-10

This pr sets default `appID = "org.gioui.app"` resulting Gio apps to show with that default appID

image

By creating simple .desktop entry  `~/.local/share/applications/org.gioui.app.desktop`

```
[Desktop Entry]
Version=1.0
Encoding=UTF-8
Type=Application
Name=Gio
GenericName=Gio application
Exec=gio %F
Terminal=false
```

image


mkungla:
This is from docs:

[xdg_toplevel::set_app_id](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id)

> Set application ID
> Set an application identifier for the surface.
> 
> The app ID identifies the general class of applications to which the surface belongs. The compositor can use this to group multiple surfaces together, or to determine how to launch a new application.
>
> For D-Bus activatable applications, the app ID is used as the D-Bus service name.
> 
> The compositor shell will try to group application surfaces together by their app ID. As a best practice, it is suggested  to select app ID's that match the basename of the application's .desktop file. For example, "org.freedesktop.FooViewer" where the .desktop file is "org.freedesktop.FooViewer.desktop".
>
> Like other properties, a set_app_id request can be sent after the xdg_toplevel has been mapped to update the property.
> 
> See the desktop-entry specification [0] for more details on application identifiers and how they relate to well-known D- Bus names and .desktop files.

[0] https://standards.freedesktop.org/desktop-entry-spec/

And since this matches already existing usage of  `-appid` for android builds for example

app/app.go
```diff
@@ -7,12 +7,19 @@ import (
...
+	// extraArgs contains extra arguments to append to
+	// os.Args. The arguments are separated with |.
+	// Useful for running programs on mobiles where the
+	// command line is not available.
+	// Set with the go linker flag -X.
+	extraArgs string
+	// appID specifies the package name for Android and Wayland
+	// or the bundle id for iOS and tvOS.
+	// appID can be set by using either AppID configuration option
+	// or using gio-cmd -appid flag.
```
eliasnaur, line 20: We don't actually have a `-target wayland` yet, right?

eliasnaur (commented):
Looks good to me. However, why not just let appID be a variable in os_wayland.go and let it be set by the -X linker flag? We can then avoid all the options machinery for something that should be a constant.

app/app.go
whereswaldon: Perhaps it should be `-target linux` anyway?

app/app.go
```diff
@@ -7,12 +7,19 @@ import (
...
+	appID = "org.gioui.app"
```
egonelbre, line 21: I'm not sure whether it's a good idea to add the identifier by default, as I understand it should be unique per app.

mkungla: No it's not required to be unique and after all you are allowed to change app_id on runtime. Having default would at least group all Gio application windows together. by default instead showing bunch of "Unknown" windows.

mkungla: Just found that for X11 there is something like that

https://github.com/gioui/gio/blob/dee53b364560b2339551a9082c4b4a8ddcbcb17d/app/os_x11.go#L796-L798

these values should be same, since they serve kind of same purpose

egonelbre: That's what I meant, grouping different gio projects together seems weird. Grouping by executable name seems like a better choice for a default behaviour.

mkungla: Ok changed default to be same as for X11

app/app.go
mkungla: Changed that comment

egonelbre (commented):
Looks good to me, but Elias has the last word :).

> However, why not just let appID be a variable in os_wayland.go and let it be set by the -X linker flag?

My intuition tells me that there's an eventual need to have them set during either app or window creation. But, I'm unable to figure out an exact scenario. I'm fine with either approach.

mkungla:
@eliasnaur I pushed these requested changes, however after playing around with it I have still feeling it should be available also as `app.Option`.

this feels counterintuitive now if I want to do somthing like this from app e.g.

```go
//go:build (linux && !android) || freebsd || openbsd

package myapp

import (
  "errors"
  "io/fs"
  "os"
  "path/filepath"
  
  _ "unsafe"
)

//go:linkname appID gioui.org/app.appID
var appID string

func setup() error {
  if len(appID) == 0 {
    log.Fatal("missing: -ldflags=\"-X 'gioui.org/app.appID=org.application.identifier'\"")
  }
  home, err := os.UserHomeDir()
  if err != nil {
    return err
  }
  desktopDir := filepath.Join(home, ".local/share/applications")
  _, err = os.Stat(desktopDir)
  if err != nil {
    return err
  }
  desktopfile := filepath.Join(desktopDir, appID+".desktop")
  if _, err := os.Stat(desktopfile); errors.Is(err, fs.ErrNotExist) {
    return os.WriteFile(desktopfile, []byte(
      "[Desktop Entry]\n"+
      "Type=Application\n"+
      "Name=My Gio Application\n"+
      "Exec="+os.Args[0],
    ), 0640)
  }
  return nil
}

```

eliasnaur:
> @eliasnaur I pushed these requested changes, however after playing around with it I have still feeling it should be available also as `app.Option`.
> 
> this feels counterintuitive now if I want to do somthing like this from app e.g.
> 
> ```go
> //go:build (linux && !android) || freebsd || openbsd
> 
> package myapp
> 
> import (
>   "errors"
>   "io/fs"
>   "os"
>   "path/filepath"
>   
>   _ "unsafe"
> )
> 
> //go:linkname appID gioui.org/app.appID
> var appID string
> 
> func setup() error {
>   if len(appID) == 0 {
>     log.Fatal("missing: -ldflags=\"-X 'gioui.org/app.appID=org.application.identifier'\"")
>   }
>   home, err := os.UserHomeDir()
>   if err != nil {
>     return err
>   }
>   desktopDir := filepath.Join(home, ".local/share/applications")
>   _, err = os.Stat(desktopDir)
>   if err != nil {
>     return err
>   }
>   desktopfile := filepath.Join(desktopDir, appID+".desktop")
>   if _, err := os.Stat(desktopfile); errors.Is(err, fs.ErrNotExist) {
>     return os.WriteFile(desktopfile, []byte(
>       "[Desktop Entry]\n"+
>       "Type=Application\n"+
>       "Name=My Gio Application\n"+
>       "Exec="+os.Args[0],
>     ), 0640)
>   }
>   return nil
> }
> ```

This example only needs a readable app id, right? If so, I don't mind an exported `app.ID`.

mkungla:
> This example only needs a readable app id, right? If so, I don't mind an exported `app.ID`.

yes, change applied

app/app.go
```diff
@@ -7,12 +7,17 @@ import (
 	"strings"
 )
 
-// extraArgs contains extra arguments to append to
-// os.Args. The arguments are separated with |.
-// Useful for running programs on mobiles where the
-// command line is not available.
-// Set with the go linker flag -X.
-var extraArgs string
+var (
```
eliasnaur, line 10: Keep the `var` declarations separate and save an indentation.

app/app.go
```diff
@@ -7,12 +7,17 @@ import (
...
+	// extraArgs contains extra arguments to append to
+	// os.Args. The arguments are separated with |.
+	// Useful for running programs on mobiles where the
+	// command line is not available.
+	// Set with the go linker flag -X.
+	extraArgs string
+	// ID Set with the go linker flag -X.
+	// Wayland: it sets the app_id
```
eliasnaur, line 18: This is difficult to understand for a Gio use that don't know the internals. How about

// ID is the app id exposed to the platform.
//
// On Android ID is the `package` property of AndroidManifest.xml,
// on iOS ID is the `CFBundleIdentifier` of the app Info.plist,
// on Wayland it's the ...
//
// ID is set by the `gogio` tool or manually with the -X linker flag:
//
//   EXAMPLE
//
// Note that ID is treated as a constant, and that changing it at runtime
// is not supported.

Note that the Android and IOS comments are a lie until `gogio` implements the linker flag.

app/os.go
```diff
@@ -43,6 +45,10 @@ type Config struct {
 	CustomRenderer bool
 	// Decorated reports whether window decorations are provided automatically.
 	Decorated bool
+
+	// id used for Wayland and X11 app_id.
+	id string
```
eliasnaur, line 50: Why not read `ID` directly from `os_wayland.go`, avoiding the overhead of an option?

app/app.go
```diff
@@ -14,11 +15,29 @@ import (
...
+// ID is the app id exposed to the platform.
+//
+// On Android ID is the package property of AndroidManifest.xml,
+// on iOS ID is the CFBundleIdentifier of the app Info.plist,
+// on Wayland it is the toplevel app_id,
+// on X11 it is the X11 XClassHint
+//
+// ID is set by the gogio tool or manually with the -X linker flag:
+//
+// EXAMPLE: gogio -ldflags="-X 'gioui.org/app.ID=org.gioui.example.Kitchen'" .
```
eliasnaur, line 27: I'd expected `go` in the example, such as:

```
...
ID is set by the gogio tool or manually with the -X linker flag. For example,

    go build -ldflags="-X 'gioui.org/app.ID=org.gioui.example.Kitchen'" .

```

I left out the concrete command so you could test it before we submit this PR.

eliasnaur:
Thanks, merged!

PR #106, opened 2022-12-10
app: wayland force redraw on config change

mkungla:
When applying window config on runtime, it is nessesary to do full redraw in order to changed config option to apply correctly. This fixes a bug where e.g window size change renders next frame in updated dimensions while native window is not scaled yet since it was waiting for stage event to apply resize.


eliasnaur:
Thanks, merged.

PR #109, opened 2023-02-23
various cleanups

jfcg:
- various cleanups

Ineffective assignments (that could be bugs) remain at:
- `layout/list.go:292`
- `text/gotext_test.go:253,256`
- `text/gotext.go:719`

Also there are about a dozen unchecked error returns.

### DIFF
--- a/gpu/compute.go
+++ b/gpu/compute.go
@@ -12,9 +12,9 @@ import (
 	"image/color"
 	"image/draw"
 	"image/png"
-	"io/ioutil"
 	"math"
 	"math/bits"
+	"os"
 	"runtime"
 	"sort"
 	"time"
@@ -655,7 +655,7 @@ func (g *compute) dumpAtlases() {
 		if err := png.Encode(&buf, nrgba); err != nil {
 			panic(err)
 		}
-		if err := ioutil.WriteFile(fmt.Sprintf("dump-%d.png", i), buf.Bytes(), 0600); err != nil {
+		if err := os.WriteFile(fmt.Sprintf("dump-%d.png", i), buf.Bytes(), 0600); err != nil {
 			panic(err)
 		}
 	}
--- a/text/text.go
+++ b/text/text.go
@@ -53,22 +53,23 @@ const (
 )
 
 const (
-	Thin       Weight = 100 - 400
-	Hairline   Weight = Thin
-	ExtraLight Weight = 200 - 400
-	UltraLight Weight = ExtraLight
-	Light      Weight = 300 - 400
-	Normal     Weight = 400 - 400
-	Medium     Weight = 500 - 400
-	SemiBold   Weight = 600 - 400
-	DemiBold   Weight = SemiBold
-	Bold       Weight = 700 - 400
-	ExtraBold  Weight = 800 - 400
-	UltraBold  Weight = ExtraBold
-	Black      Weight = 900 - 400
-	Heavy      Weight = Black
-	ExtraBlack Weight = 950 - 400
-	UltraBlack Weight = ExtraBlack
+	Thin       Weight = -300
+	ExtraLight Weight = -200
+	Light      Weight = -100
+	Normal     Weight = 0
+	Medium     Weight = 100
+	SemiBold   Weight = 200
+	Bold       Weight = 300
+	ExtraBold  Weight = 400
+	Black      Weight = 500
+
+	Hairline   = Thin
+	UltraLight = ExtraLight
+	DemiBold   = SemiBold
+	UltraBold  = ExtraBold
+	Heavy      = Black
+	ExtraBlack = Black + 50
+	UltraBlack = ExtraBlack
 )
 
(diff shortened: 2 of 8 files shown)


text/text.go
```diff
@@ -53,22 +53,23 @@ const (
...
+	ExtraBold
+	Black
+
+	Hairline   = Thin
+	UltraLight = ExtraLight
+	DemiBold   = SemiBold
+	UltraBold  = ExtraBold
+	Heavy      = Black
+	ExtraBlack = Black + 50
+	UltraBlack = ExtraBlack
```
theclapp, line 72: I think the use of `iota` actually makes this less clear, and invites errors if some new weight is inserted in the list.  I think I'd just prefer literals:

```go
Thin       Weight = -300
ExtraLight Weight = -200
Light      Weight = -100
Normal     Weight = 0
Medium     Weight = 100
SemiBold   Weight = 200
Bold       Weight = 300
ExtraBold  Weight = 400
Black      Weight = 500
```

I do like the consolidation of `Hairline` - `UltraBlack` at the bottom, I think that does make things clearer.

jfcg: Since the list is comprehensive from thin to black, I don't think a new entry will arrive, maybe to either end. Mid values (+50) could be added to the second half.

Also any decent ide will show a constant's value when you hover your mouse over it.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,30 @@
+name: CodeQL
```
eliasnaur, line 1: Please leave this out of the PR, I'm not convinced it's worth it.

gpu/compute.go
```diff
@@ -1898,11 +1898,11 @@ func (c *collector) hashOp(op paintOp) uint64 {
 	for _, cl := range op.clipStack {
 		k := cl.state
 		keyBytes := (*[unsafe.Sizeof(k)]byte)(unsafe.Pointer(unsafe.Pointer(&k)))
-		c.hasher.Write(keyBytes[:])
+		_, _ = c.hasher.Write(keyBytes[:])
```
eliasnaur, line 1901: Revert these changes. FWIW, I don't agree the extra ceremony `_, _` adds is worth it.

text/shaper_test.go
```diff
@@ -35,19 +35,19 @@ func TestWrappingTruncation(t *testing.T) {
...
+				break
 			}
 		}
+		min := i
+		if min > untruncatedCount {
+			min = untruncatedCount
+		}
+		if lineCount != min {
+			t.Errorf("expected %d lines, got %d", min, lineCount)
+		}
```
eliasnaur, line 50: @whereswaldon can you take a look at this change?

text/text.go
eliasnaur: I agree with @theclapp that we should not use `iota` to compute these constants. I also think we should get rid of the duplicates, and perhaps even cut down on the number of constants. In any case, ok to do in a future change.

eliasnaur (commented):
Thank you.

A few comments are inline. Please squash your commits before submitting an update.

.github/workflows/codeql.yml
jfcg: Ok

gpu/compute.go
jfcg: They silence the "error return" warnings from `golangci-lint` for those lines. The remaining warnings, about a dozen, seem to require error checks.

Also I do recommend to add this linter to your pipeline.

jfcg:
rebased and removed codeql

text/shaper_test.go
whereswaldon: This change is valid, but mostly points out that I've checked whether truncation occurred using an unexported implementation detail instead of the exported API. I think I'll rework this test case to count the lines using the glyph stream instead.

whereswaldon: I've sent [this patch](https://lists.sr.ht/~eliasnaur/gio-patches/patches/39302) to cover simplifying the test. Thanks @jfcg for pointing out that this can be improved!

I don't especially feel the need to rewrite all of the glyph iteration loops to use the style proposed here. It is simpler on some levels, but many more lines. When/if the custom iterators proposal is accepted, the multiple calls to NextGlyph will collapse into one invocation anyway.

jfcg:
- removed glyph for loop part
- removed iota from font weight constants

gpu/compute.go
whereswaldon: We appreciate your recommendation, but we choose not to use that metalinter at this time. Many of the linters that it executes are unhelpful, and some of them produce contradictory recommendations. `golangci-lint` also disables some important checks within `staticcheck`. Silencing a lint warning from a linter that we do not use is not sufficient reason to make this change.

jfcg: fair enough

whereswaldon (approved):
Thank you!

eliasnaur:
Thanks, merged.

PR #110, opened 2023-02-24
app: [WASM] add ViewEvent

inkeliz:
This change adds ViewEvent for JS/WASM, which returns the HTMLElement which Gio is been rendered, once started.

### DIFF
--- a/app/os_js.go
+++ b/app/os_js.go
@@ -22,7 +22,9 @@ import (
 	"gioui.org/unit"
 )
 
-type ViewEvent struct{}
+type ViewEvent struct {
+	Element js.Value
+}
 
 type contextStatus int
 
@@ -111,6 +113,7 @@ func newWindow(win *callbacks, options []Option) error {
 		w.w.SetDriver(w)
 		w.Configure(options)
 		w.blur()
+		w.w.Event(ViewEvent{Element: cont})
 		w.w.Event(system.StageEvent{Stage: system.StageRunning})
 		w.resize()
 		w.draw(true)


inkeliz:
I force-push to send an empty ViewEvent when the view is destroyed. That is the behavior on iOS/macOS/.....

app/os_js.go
```diff
@@ -562,7 +566,12 @@ func (w *window) Configure(options []Option) {
 	w.w.Event(ConfigEvent{Config: w.config})
 }
 
-func (w *window) Perform(system.Action) {}
+func (w *window) Perform(acts system.Action) {
+	if acts&system.ActionClose != 0 {
+		w.w.Event(ViewEvent{})
+		w.w.Event(system.DestroyEvent{})
+	}
+}
```
eliasnaur, line 574: I had expected `Perform(system.ActionClose)` to trigger a shutdown sequence that ended up in, say, the `visibilitychange` listener that takes care of sending `StageEvent` updates. As far as I can tell, everything keeps running even though you've received a `DestroyEvent`.

It would probably be fine to revert this change completely, punting the shutdown implementation.

app/os_js.go
```diff
@@ -2,6 +2,7 @@
 
 package app
 
+import "C"
```
eliasnaur, line 5: What's this for? It doesn't seem to be used anywhere.

inkeliz: Fixed

eliasnaur:
Merged, thanks.

PR #111, opened 2023-02-25
macOS: external drag and drop support

StarHack:
Adds some very basic DnD support for external files on macOS.

- I don't understand event routing yet, so the implementation is probably incorrect.
- Support for Windows and Linux missing
- It probably should be possible to add this for specific widgets and not for the whole window.

### Usage

**go.mod**
```go
module main

go 1.19

replace gioui.org => github.com/StarHack/gio v0.0.0-20230421195057-eb440387490a

require gioui.org v0.0.0-20230414223051-b6e0376ad2fe

require (
	gioui.org/cpu v0.0.0-20210817075930-8d6a761490d2 // indirect
	gioui.org/shader v1.0.6 // indirect
	github.com/benoitkugler/textlayout v0.3.0 // indirect
	github.com/go-text/typesetting v0.0.0-20221214153724-0399769901d5 // indirect
	golang.org/x/exp v0.0.0-20221012211006-4de253d81b95 // indirect
	golang.org/x/exp/shiny v0.0.0-20220827204233-334a2380cb91 // indirect
	golang.org/x/image v0.0.0-20220722155232-062f8c9fd539 // indirect
	golang.org/x/sys v0.0.0-20220825204002-c680a09ffe64 // indirect
	golang.org/x/text v0.7.0 // indirect
)
```

**main.go**
```go
package main

import (
	"fmt"
	"image"
	_ "image/gif"
	"io"
	"os"

	"gioui.org/app"
	"gioui.org/io/system"
	"gioui.org/io/transfer"
	"gioui.org/layout"
	"gioui.org/op"
	"gioui.org/op/paint"
	"gioui.org/widget"
)

func loadImage(file io.ReadCloser) (widget.Image, error) {
	img := widget.Image{}
	decImg, _, err := image.Decode(file)
	if err != nil {
		return img, err
	}
	img.Src = paint.NewImageOp(decImg)
	return img, nil
}

func main() {
	go func() {
		w := app.NewWindow()
		var ops op.Ops

		f, _ := os.Open("img/placeholder.png")
		img, _ := loadImage(f)

		tag := new(int)

		for {
			e := <-w.Events()

			switch e := e.(type) {
			case system.DestroyEvent:
				return
			case system.FrameEvent:

				transfer.TargetOp{
					Tag:  tag,
					Type: "image/png",
				}.Add(&ops)

				for _, gtxEvent := range gtx.Events(tag) {
					switch e := gtxEvent.(type) {
					case transfer.DataEvent:
						file, err := e.Open()
						newImg, err := loadImage(file)
						if err == nil {
							img = newImg
						} else {
							fmt.Println(err)
						}
					}
				}

				img.Layout(gtx)
				e.Frame(gtx.Ops)
			}
		}
	}()
	app.Main()
}
```

### Demo

https://user-images.githubusercontent.com/5388534/221357178-7e800dc7-9513-4575-9e84-1b191a6cb406.mov

### DIFF
--- a/app/os_macos.go
+++ b/app/os_macos.go
@@ -8,6 +8,10 @@ package app
 import (
 	"errors"
 	"image"
+	"io"
+	"mime"
+	"os"
+	"path/filepath"
 	"runtime"
 	"time"
 	"unicode"
@@ -18,6 +22,7 @@ import (
 	"gioui.org/io/key"
 	"gioui.org/io/pointer"
 	"gioui.org/io/system"
+	"gioui.org/io/transfer"
 	"gioui.org/unit"
 
 	_ "gioui.org/internal/cocoainit"
@@ -554,6 +559,22 @@ func gio_onMouse(view, evt C.CFTypeRef, cdir C.int, cbtn C.NSInteger, x, y, dx, 
 		Position:  pos,
 		Scroll:    f32.Point{X: dxf, Y: dyf},
 		Modifiers: convertMods(mods),
+	})
+}
+
+//export gio_onExternalDrop
+func gio_onExternalDrop(view C.CFTypeRef, path *C.char) {
+	fileUrl := C.GoString(path)
+	w := mustView(view)
+
+	fileExtension := filepath.Ext(fileUrl)
+	mime := mime.TypeByExtension(fileExtension)
+
+	w.w.Event(transfer.DataEvent{
+		Type: mime,
+		Open: func() (io.ReadCloser, error) {
+			return os.Open(fileUrl)
+		},
 	})
 }
 
--- a/app/os_macos.m
+++ b/app/os_macos.m
@@ -110,6 +110,20 @@ static void handleMouse(NSView *view, NSEvent *event, int typ, CGFloat dx, CGFlo
 - (void)mouseDragged:(NSEvent *)event {
 	handleMouse(self, event, MOUSE_MOVE, 0, 0);
 }
+-(NSDragOperation)draggingEntered:(id < NSDraggingInfo >)sender
+{
+	return NSDragOperationCopy;
+}
+- (void)draggingEnded:(id <NSDraggingInfo>)sender
+{
+	NSPasteboard* pbrd = [sender draggingPasteboard];
+	NSArray* droppedFiles = [pbrd propertyListForType:NSFilenamesPboardType];
+
+	for (NSString* filePath in droppedFiles) {
+		NSURL* url = [NSURL fileURLWithPath:filePath];
+		gio_onExternalDrop((__bridge CFTypeRef)self, (char*)[[url path] UTF8String]);
+	}
+}
 - (void)scrollWheel:(NSEvent *)event {
 	CGFloat dx = -event.scrollingDeltaX;
 	CGFloat dy = -event.scrollingDeltaY;
@@ -366,6 +380,7 @@ CFTypeRef gio_createView(void) {
 	@autoreleasepool {
 		NSRect frame = NSMakeRect(0, 0, 0, 0);
 		GioView* view = [[GioView alloc] initWithFrame:frame];
+		[view registerForDraggedTypes: [NSArray arrayWithObjects:NSTIFFPboardType, NSFilenamesPboardType, nil]];
 		view.wantsLayer = YES;
 		view.layerContentsRedrawPolicy = NSViewLayerContentsRedrawDuringViewResize;
 		return CFBridgingRetain(view);
--- a/io/router/router.go
+++ b/io/router/router.go
@@ -182,6 +182,8 @@ func (q *Router) Queue(events ...event.Event) bool {
 			}
 		case clipboard.Event:
 			q.cqueue.Push(e, &q.handlers)
+		case transfer.DataEvent:
+			q.pointer.queue.notifyPotentialTargets(&pointerHandler{sourceMimes: []string{e.Type}}, &q.handlers, e)
 		}
 	}
 	return q.handlers.HadEvents()
--- a/io/router/pointer.go
+++ b/io/router/pointer.go
@@ -801,6 +801,14 @@ func (q *pointerQueue) deliverEnterLeaveEvents(p *pointerInfo, events *handlerEv
 	p.entered = append(p.entered[:0], hits...)
 }
 
+func (q *pointerQueue) notifyPotentialTargets(src *pointerHandler, events *handlerEvents, event event.Event) {
+	for k, tgt := range q.handlers {
+		if _, ok := firstMimeMatch(src, tgt); ok {
+			events.Add(k, event)
+		}
+	}
+}
+
 func (q *pointerQueue) deliverDragEvent(p *pointerInfo, events *handlerEvents) {
 	if p.dataSource != nil {
 		return
@@ -814,11 +822,7 @@ func (q *pointerQueue) deliverDragEvent(p *pointerInfo, events *handlerEvents) {
 		// One data source handler per pointer.
 		p.dataSource = k
 		// Notify all potential targets.
-		for k, tgt := range q.handlers {
-			if _, ok := firstMimeMatch(src, tgt); ok {
-				events.Add(k, transfer.InitiateEvent{})
-			}
-		}
+		q.notifyPotentialTargets(src, events, transfer.InitiateEvent{})
 		break
 	}
 }
@@ -858,9 +862,9 @@ func (q *pointerQueue) deliverTransferDataEvent(p *pointerInfo, events *handlerE
 	transferIdx := len(q.transfers)
 	events.Add(p.dataTarget, transfer.DataEvent{
 		Type: src.offeredMime,
-		Open: func() io.ReadCloser {
+		Open: func() (io.ReadCloser, error) {
 			q.transfers[transferIdx] = nil
-			return src.data
+			return src.data, nil
 		},
 	})
 	q.transfers = append(q.transfers, src.data)
--- a/io/transfer/transfer.go
+++ b/io/transfer/transfer.go
@@ -103,7 +103,7 @@ type DataEvent struct {
 	Type string
 	// Open returns the transfer data. It is only valid to call Open in the frame
 	// the DataEvent is received. The caller must close the return value after use.
-	Open func() io.ReadCloser
+	Open func() (io.ReadCloser, error)
 }
 
 func (DataEvent) ImplementsEvent() {}


app/os_macos.m
```diff
@@ -110,6 +110,27 @@ - (void)mouseMoved:(NSEvent *)event {
...
+	NSPasteboard* pbrd = [sender draggingPasteboard];
+
+	for (NSPasteboardItem *item in [pbrd pasteboardItems]) {
+		for (NSString *type in [item types]) {
+			if ([type isEqualToString:@"public.file-url"]) {
+				NSData* data = [item dataForType:type];
+				NSString *clipboard_url = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+				NSURL *url = [[NSURL alloc] initWithString:clipboard_url];
+				NSString *real_path = url.path;
+				char* content = (char*)[real_path UTF8String];
```
eliasnaur, line 128: I suggest replacing this with passing the `NSData *data` to Go and let Go code parse it as an URL. I believe `NSData` has a [`bytes`](https://developer.apple.com/documentation/foundation/nsdata/1410616-bytes) property to get a pointer to the data, along with a `length` for its length.

app/os_macos.m
```diff
@@ -366,6 +387,7 @@ CFTypeRef gio_createView(void) {
 	@autoreleasepool {
 		NSRect frame = NSMakeRect(0, 0, 0, 0);
 		GioView* view = [[GioView alloc] initWithFrame:frame];
+		[view registerForDraggedTypes: [NSArray arrayWithObjects:NSTIFFPboardType,NSFilenamesPboardType,nil]];
```
eliasnaur, line 390: I believe the style is space after list commas.

app/os_macos.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"gioui.org/internal/f32"
 	"gioui.org/io/clipboard"
+	"gioui.org/io/externalDragDrop"
```
eliasnaur, line 18: I would expect this PR referred the `io/transfer` package. See also `widget.Draggable` and the `ExampleDraggable_Layout` example.

eliasnaur (changes requested):
Thank you for tackling this. Some comments inline, the most important being the use of the `io/transfer` package which is intended to cover drag-and-drop.

Also, please remove unrelated and debug commits from the PR.

StarHack:
@eliasnaur Updated implementation to use `io/transfer` package instead and took into account the other change requests to simplify the implementation. Also updated the usage example above.

app/os_macos.m
```diff
@@ -110,6 +110,20 @@ - (void)mouseMoved:(NSEvent *)event {
...
+    return NSDragOperationCopy; // NSDragOperationGeneric
+}
+- (void)draggingEnded:(id <NSDraggingInfo>)sender
+{
+	NSPasteboard* pbrd = [sender draggingPasteboard];
+	NSArray* droppedFiles = [pbrd propertyListForType:NSFilenamesPboardType];
+
+    for (NSString* filePath in droppedFiles) {
+        NSURL* url = [NSURL fileURLWithPath:filePath];
+		gio_onExternalDrop((__bridge CFTypeRef)self, (char*)[[url path] UTF8String]);
```
eliasnaur, line 124: This section seems to mix tabs and spaces.

io/router/router.go
```diff
@@ -182,6 +182,8 @@ func (q *Router) Queue(events ...event.Event) bool {
 			}
 		case clipboard.Event:
 			q.cqueue.Push(e, &q.handlers)
+		case transfer.DataEvent:
+			q.handlers.Add(nil, e)
```
eliasnaur, line 186: This just delivers the even to the `nil` tag. I had expected the event be delivered to the current target, similar to https://git.sr.ht/~eliasnaur/gio/tree/main/item/io/router/pointer.go#L859. See also the `widget/dnd_test.go` for an end-to-end use of the transfer protocol.

app/os_macos.go
```diff
@@ -557,6 +560,19 @@ func gio_onMouse(view, evt C.CFTypeRef, cdir C.int, cbtn C.NSInteger, x, y, dx,
 	})
 }
 
+//export gio_onExternalDrop
+func gio_onExternalDrop(view C.CFTypeRef, path *C.char) {
+	fileUrl := C.GoString(path)
+	w := mustView(view)
+
+	w.w.Event(transfer.DataEvent{
+		Type: "external-file",
```
eliasnaur, line 569: This should be a MIME type, as documented in the field. Doesn't the macOS API give us a type?

eliasnaur (commented):
Much better, thanks. Next issue is that the data events should arrive at the transfer target.

Also, please squash your commits into logically independent units (or just one commit with everything).

StarHack:
@eliasnaur Many thanks for your feedback. I have adjusted the implementation which now allows to register a TargetOp with the desired mime type and the event is now routed to the appropriate tag. Furthermore, the sample implementation above has been updated and all commits have been squashed into one.

fjl:
@StarHack thanks for working on this! I really need it in my own app.

app/os_macos.go
```diff
@@ -557,6 +562,22 @@ func gio_onMouse(view, evt C.CFTypeRef, cdir C.int, cbtn C.NSInteger, x, y, dx,
...
+
+	fileExtension := filepath.Ext(fileUrl)
+	mime := mime.TypeByExtension(fileExtension)
+
+	w.w.Event(transfer.DataEvent{
+		Type: mime,
+		Open: func() io.ReadCloser {
+			return io.NopCloser(strings.NewReader(fileUrl))
+		},
+	})
```
eliasnaur, line 578: Hm, thinking of this some more, it seems better to present the data directly, instead of through an URL. Otherwise, how would a receiver tell the difference between content of type `Type` and a file URL pointing to a file of type `Type`? If you need the filename it seems ok to add that as a field on `transfer.DataEvent`.

However, that also means you need to add an `error` to the `transfer.DataEvent.Open` field.

eliasnaur (commented):
Thank you. More comments below. I'm sorry for the bait-and-switch; you're hit by the fact that you're the first to implement platform support for drag-n-drop. Before your PR we only had internal dnd.

app/os_macos.go
```diff
@@ -557,6 +562,22 @@ func gio_onMouse(view, evt C.CFTypeRef, cdir C.int, cbtn C.NSInteger, x, y, dx,
 	})
 }
 
+//export gio_onExternalDrop
+func gio_onExternalDrop(view C.CFTypeRef, path *C.char) {
+	fileUrl := C.GoString(path)
+	w := mustView(view)
+
+	fileExtension := filepath.Ext(fileUrl)
+	mime := mime.TypeByExtension(fileExtension)
```
fjl, line 571: I think it's a bit weird to do detection by file extension here. The detection might not always work (lets say for example, the user is dropping an image file with an unknown extension, or no extension), and in order to receive 'any file' the app would need to register every possible mime type.

app/os_macos.go
fjl: Another correct solution would be setting `Type` to `"text/uri-list"`, and then returning `file://` data through the `Open` reader.

app/os_macos.go
eliasnaur: Correct, but let's support "any file" then, in the form of mime type patterns (e.g. "*" or "image/*"). Not for this PR, of course.

app/os_macos.go
eliasnaur: Less magic, sure, but also more work for every dnd target that has to implement support for "text/uri-list". Your browser example is compelling, though.

StarHack:
@eliasnaur @fjl Many thanks for your feedback! I appreciate it. I changed the implementation to return the result of `os.Open()` instead and we now only send the event to the topmost element of the hitlist.

io/router/router.go
```diff
@@ -182,6 +182,19 @@ func (q *Router) Queue(events ...event.Event) bool {
...
+					targetMimes := pq.handlers[h.tag].targetMimes
+					for _, mimeType := range targetMimes {
+						if mimeType == e.Type {
+							q.handlers.Add(h.tag, e)
+							break
+						}
+					}
+				}
+			}
 		}
```
eliasnaur, line 198: This is probably correct, but seems to duplicate existing logic in `pointerQueue`. Can this code be merged with, say, `pointerQueue.deliverTransferDataEvent` so the internal and external code dnd paths become similar?

fjl (changes requested):
I have verified the browser logic for handling external file DnD, and have to say, I was wrong about `text/url-list`.

Here is a JSFiddle where you can try out DnD: https://jsfiddle.net/6yezxnqp/1/

In browsers, file DnD has special handling. The drag event has a property `ev.dataTransfer.types`, which is an array of object types. These can be MIME types, but don't have to be. When files are dragged, the [HTML spec requires](https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-types-dev) a type `"Files"` to be included in this array.

For each dropped file, the browser *does* perform MIME type detection, so each transfer *item* will have a MIME type, and it is based on the file extension. I confirmed this by dropping a PDF file with the extension removed. The `ev.dataTransfer.items[0].type` property was `""` in that case.

So, overall there is lots of inspiration to be found in the way browsers handle it. For obvious security-related reasons, the browser does not provide access to the full file path (anymore). However, the basename of the file is there.

I think that gio should provide the full path, or at the very least the file basename, alongside the open function. One must also consider the possibility of dropping multiple files, directories, symlinks etc. Maybe a good way to handle all this would be changing  `DataEvent` like this:

```go
package transfer

import "fs"
import "io"

type DataEvent struct{
    Type string
    Open io.ReadCloser
    Files []File
}

type File interface{
    fs.FileInfo
    Open() (io.ReadCloser, error)    
}
```

i.e. we add special handling for files. This might be required anyway in order to handle dragging files *out* from a Gio application into an external location.

whereswaldon:
Just a thought:

> So, overall there is lots of inspiration to be found in the way browsers handle it. For obvious security-related reasons, the browser does not provide access to the full file path (anymore). However, the basename of the file is there.
> 
> I think that gio should provide the full path, or at the very least the file basename, alongside the open function.

Gio does run in the browser when compiled to WASM, so whatever we do needs to be implementable within the browser. I suppose we could degrade gracefully from full OS paths to whatever the browser offers though, so long as we document that you *may* get a full OS file path.

app/os_macos.go
```diff
@@ -557,6 +562,22 @@ func gio_onMouse(view, evt C.CFTypeRef, cdir C.int, cbtn C.NSInteger, x, y, dx,
...
+
+	w.w.Event(transfer.DataEvent{
+		Type: mime,
```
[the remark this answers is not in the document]
StarHack: The current implementation triggers an event for every file dropped so providing multiple `Files` at once probably would be inconsistent as the `Open` function still only provides access to one file. But I agree that including the (file) URL would make sense, or at least the filename, as Gio users might need them depending on their implementation.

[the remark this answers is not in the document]
fjl: I posted another API idea specifically adding files in the main review comment. Let's see what @eliasnaur has to say about it...

eliasnaur:
@fjl adding all that information seems a bit much to promise. Questions:

- Can a list of dropped files be implemented as separate DataEvents?
- Are you ok with just an URL field with a `file:` reference for local files?

StarHack:
@eliasnaur Do you have further feedback on the existing implementation? I'd like to finalize this pull request at one point or another and add support for Linux and Windows as soon as I've time to do so.

fjl:
I proposed `fs.FileInfo` mostly because it is the usual way of representing 'information about a file' in Go. If you feel it's too much, of course I'd also be fine with just a name/URL. If the URL is a complete path, the app can always resolve all necessary information itself. There are two situations where it will not work: (1) with the WASM/browser backend,
(2) when the app is running in a sandbox. But we can worry about those cases later.

Delivering multiple files as individual events is not great, but this could just be my personal opinion. I have no experience to back that up. The act of dropping a set of files is 'one event', logically speaking. In some contexts, dropping multiple files will make sense, and requires handling from the app. This can be dealt with by accumulating events. In other cases, e.g. 'avatar image box' in a messenger, only a single file is acceptable. If multiple files are dropped, the app will process every file (in case of image: open, parse, crop etc.) and then use the last one. It can't do anything else because it has no knowledge about multiple file events being in the queue.

fjl:
I don't want to be the guy holding up this extremely useful PR. If you feel it is the simplest, let's just go with a `URL` field on `transfer.DataEvent` and worry about the other problems later!

eliasnaur (commented):
I had this review comment, but it wasn't submitted because of my reply to @fjl . Sorry.

ddkwork:
A similar structure is also needed on Windows, we may drag and drop a folder instead of just a file, or a batch of drag-and-drop, or more likely a file and a directory to drag and drop together, and a drag-and-drop link is possible, so the original structure needs to be changed or a new event channel and data structure is opened for the drag-and-drop event on the desktop platform

ddkwork:
> I had this review comment, but it wasn't submitted because of my reply to @fjl . Sorry.

I implemented gio's window header dark theme and drag-and-drop support by copying the wails code, but don't know how to add it to the event correctly, can you make an example?

justyntemme:
> @whereswaldon My bad, I completely forgot about this one. Frankly, I don't think it makes sense to waste any more time on this.

Hey no worries at all, can you confirm to releasing your code changes to the public domain so that I may continue the work? Thanks for bringing us this far! Open source is a collaborative effort and you have done quite a lot in this effort.

PR #112, opened 2023-03-21
text: fix sorting in faceOrderer.sorted

theclapp:
faceOrderer.sorted tried to put the "primary" font first by tweaking the "less" function in sort.Slice, but it didn't work correctly.

If item i equaled the "primary" font, less() always returned true. This did not take into account if item j was the "primary" font, in which case it could easily be sorted differently.

Rather than adding another special case for that, which I couldn't convince myself was actually correct in every case, I just searched for the "primary" font and moved it to the front of the slice, and then omitted the first item of the slice from the rest of the sorting.


eliasnaur:
Merged, thanks. A test would be a nice follow-up :)

PR #113, opened 2023-04-18
app: [Android] Set high refresh rate on startup

ilidemi:
Some devices with high refresh rates limit SurfaceView apps to 60hz and need a specific API call to set it back. Same approach is used by https://github.com/ajinasokan/flutter_displaymode. The extra work is skipped on the devices that don't need it.

### DIFF
--- a/app/GioView.java
+++ b/app/GioView.java
@@ -26,6 +26,7 @@ import android.text.SpannableStringBuilder;
 import android.util.AttributeSet;
 import android.util.TypedValue;
 import android.view.Choreographer;
+import android.view.Display;
 import android.view.KeyCharacterMap;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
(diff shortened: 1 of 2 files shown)


app/os_android.go
```diff
@@ -776,6 +780,13 @@ func (w *window) setVisible(env *C.JNIEnv) {
 	if width == 0 || height == 0 {
 		return
 	}
+	displayModeCount := int(C.jni_CallIntMethod(env, w.view, gioView.countDisplayModes))
+	if displayModeCount > 1 {
+		// Only do the main thread work on devices that need it
+		runOnMainWithEnv(env, func(mainEnv *C.JNIEnv) {
+			callVoidMethod(mainEnv, w.view, gioView.setHighRefreshRate)
+		})
+	}
```
eliasnaur, line 789: I'm surprised you need this for every `setVisible`. Isn't it enough to set the display mode once, in the `GioView` Java constructor?

app/GioView.java
```diff
@@ -474,6 +475,67 @@ void updateCaret(float m00, float m01, float m02, float m10, float m11, float m1
...
+		if (minRefreshRate == maxRefreshRate) {
+			// Can't improve the refresh rate
+			return;
+		}
+
+		Window window = ((Activity) context).getWindow();
+		WindowManager.LayoutParams layoutParams = window.getAttributes();
+		layoutParams.preferredDisplayModeId = bestModeId;
+
+		// This is the call that needs to happen on the main thread
```
eliasnaur, line 535: Drop comment. It's implied by pretty much all Android GUI API.

app/os_android.go
ilidemi: Looks like it is, much more straightforward this way.

eliasnaur:
Merged, thanks!

gedw99:
Anyone know if screen refresh applies to iOS also ?

or maybe it’s automatically handled with no need for an api to … 

iPhone 14 Pro and 14 Pro Max both ship with a ProMotion display, which allows for an adaptive refresh rate of 1Hz to 120Hz. The refresh rate goes all the way down to 1Hz, in order to save battery…

I don’t have a newer iOS device do can’t test …

PR #115, opened 2023-06-16
app: Consider screen scale when performing system.ActionCenter on macOS

lotodore:
Fixing https://todo.sr.ht/~eliasnaur/gio/505 so that system.ActionCenter is working using builtin MacBook Pro screens.


eliasnaur:
Merged, thanks!

PR #116, opened 2023-07-04
app: [js] fix tabbing

inkeliz:


widget/editor.go
```diff
@@ -319,6 +319,9 @@ func (e *Editor) processKey(gtx layout.Context) {
 		switch ke := ke.(type) {
 		case key.FocusEvent:
 			e.focused = ke.Focus
+			if e.focused {
+				key.SoftKeyboardOp{Show: true}.Add(gtx.Ops)
+			}
```
eliasnaur, line 324: Can you expand the commit message to say what issue this fixes? Is there a TODO number?

inkeliz: Without that change TAB on browser will switch the focus to another input (which is expected!), but the focus is removed, in the next frame. 

Currently, on JS, Gio focus/blur one input based on `SoftKeyboardOp`, even on desktop.

Previously, SoftKeyboardOp is requested using `requestFocus`, which happens only after clicking into the input. However, when the focus shifts to another input without clicking (which is the tabbing case): it don't triggers `requestFocus`, then it doesn't use `SoftKeyboardOp`.

eliasnaur: This losing of focus after TAB sounds like a browser-only issue, correct? If so, can it be fixed in `os_js.go` so `Editor` doesn't need to know?

PR #117, opened 2023-07-08
io,app: add deeplinking support

inkeliz:
Now, it's possible to launch one Gio app using a custom URI scheme, such as `gio://some/data`.

This feature is supported on Android, iOS, macOS and Windows, issuing a new deeplink.Event,
containing the URL launched. If the program is already opened, one deeplink.Event will be
sent to the current opened app.

Limitations:
On Windows, if the program uses deeplink (compiled with `-deeplink`), then just a single
instance of the app can be open. In other words, just a single `myprogram.exe` can
be active.

Security:
Deeplinking have the same level of security of clipboard. Any other software can send such
information and read the content, without any restriction. That should not be used to transfer
sensible data, and can't be fully trusted.

Setup/Compiling:
In order to set the custom scheme, you need to use the new `-deeplink` flag in `gogio`, using
as `-deeplink gio` will listen to `gio://`.

If you are not using gogio you need to defined some values, which varies for each OS:

macOS/iOS - You need to define the following Properly List:
```
<key>CFBundleURLTypes</key>
<array>
  <dict>
	<key>CFBundleURLSchemes</key>
	<array>
	  <string>yourCustomScheme</string>
	</array>
  </dict>
</array>
```

Windows - You need to compiling using -X argument:
```
-ldflags="-X "gioui.org/app.schemesDeeplink=yourCustomScheme" -H=windowsgui"
```

Android - You need to add IntentFilter in GioActivity:
```
<intent-filter>
	<action android:name="android.intent.action.VIEW"></action>
	<category android:name="android.intent.category.DEFAULT"></category>
	<category android:name="android.intent.category.BROWSABLE"></category>
	<data android:scheme="yourCustomScheme"></data>
</intent-filter>
```

That assumes that you still using GioActivity and GioAppDelegate, otherwise more
changes are required.

### DIFF
--- a/app/GioActivity.java
+++ b/app/GioActivity.java
@@ -4,6 +4,7 @@ package org.gioui;
 
 import android.app.Activity;
 import android.os.Bundle;
+import android.content.Intent;
 import android.content.res.Configuration;
 import android.view.ViewGroup;
 import android.view.View;
@@ -29,6 +30,7 @@ public final class GioActivity extends Activity {
 
             layer.addView(view);
             setContentView(layer);
+            onNewIntent(this.getIntent());
 	}
 
 	@Override public void onDestroy() {
@@ -59,5 +61,10 @@ public final class GioActivity extends Activity {
 	@Override public void onBackPressed() {
 		if (!view.backPressed())
 			super.onBackPressed();
+	}
+
+	@Override protected void onNewIntent(Intent intent) {
+		super.onNewIntent(intent);
+        view.onIntentEvent(intent);
 	}
 }
--- a/app/GioView.java
+++ b/app/GioView.java
@@ -12,6 +12,7 @@ import android.app.Fragment;
 import android.app.FragmentManager;
 import android.app.FragmentTransaction;
 import android.content.Context;
+import android.content.Intent;
 import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Matrix;
@@ -311,6 +312,15 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 		window.setAttributes(layoutParams);
 	}
 
+	protected void onIntentEvent(Intent intent) {
+        if (intent == null) {
+            return;
+        }
+        if (intent.getData() != null) {
+            this.onDeeplink(nhandle, intent.getData().toString());
+        }
+    }
+
 	@Override protected boolean dispatchHoverEvent(MotionEvent event) {
 		if (!accessManager.isTouchExplorationEnabled()) {
 			return super.dispatchHoverEvent(event);
@@ -549,6 +559,7 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 	static private native void onExitTouchExploration(long handle);
 	static private native void onA11yFocus(long handle, int viewId);
 	static private native void onClearA11yFocus(long handle, int viewId);
+	static private native void onDeeplink(long handle, String uri);
 	static private native void imeSetSnippet(long handle, int start, int end);
 	static private native String imeSnippet(long handle);
 	static private native int imeSnippetStart(long handle);
--- a/app/internal/windows/windows.go
+++ b/app/internal/windows/windows.go
@@ -99,6 +99,8 @@ const (
 
 	CW_USEDEFAULT = -2147483648
 
+	ERROR_ALREADY_EXISTS = 183
+
 	GWL_STYLE = ^(uintptr(16) - 1) // -16
 
 	GCS_COMPSTR       = 0x0008
@@ -111,7 +113,8 @@ const (
 	CFS_POINT        = 0x0002
 	CFS_CANDIDATEPOS = 0x0040
 
-	HWND_TOPMOST = ^(uint32(1) - 1) // -1
+	HWND_TOPMOST   = ^(uint32(1) - 1) // -1
+	HWND_BROADCAST = 0xFFFF
 
 	HTCAPTION     = 2
 	HTCLIENT      = 1
@@ -303,15 +306,25 @@ const (
 	LR_MONOCHROME       = 0x00000001
 	LR_SHARED           = 0x00008000
 	LR_VGACOLOR         = 0x00000080
+
+	FILE_MAP_READ = 0x0004
+)
+
+var (
+	GIO_DEEPLINKING uint32 // Custom message for deep linking, lazy init
 )
 
 var (
 	kernel32          = syscall.NewLazySystemDLL("kernel32.dll")
+	_CreateMutex      = kernel32.NewProc("CreateMutexW")
+	_GetMutexInfo     = kernel32.NewProc("GetMutexInfo")
 	_GetModuleHandleW = kernel32.NewProc("GetModuleHandleW")
 	_GlobalAlloc      = kernel32.NewProc("GlobalAlloc")
 	_GlobalFree       = kernel32.NewProc("GlobalFree")
 	_GlobalLock       = kernel32.NewProc("GlobalLock")
 	_GlobalUnlock     = kernel32.NewProc("GlobalUnlock")
+	_OpenFileMapping  = kernel32.NewProc("OpenFileMappingA")
+	_ReleaseMutex     = kernel32.NewProc("ReleaseMutex")
 
 	user32                       = syscall.NewLazySystemDLL("user32.dll")
 	_AdjustWindowRectEx          = user32.NewProc("AdjustWindowRectEx")
@@ -347,9 +360,11 @@ var (
 	_PostQuitMessage             = user32.NewProc("PostQuitMessage")
 	_ReleaseCapture              = user32.NewProc("ReleaseCapture")
 	_RegisterClassExW            = user32.NewProc("RegisterClassExW")
+	_RegisterWindowMessage       = user32.NewProc("RegisterWindowMessageW")
 	_ReleaseDC                   = user32.NewProc("ReleaseDC")
 	_ScreenToClient              = user32.NewProc("ScreenToClient")
 	_ShowWindow                  = user32.NewProc("ShowWindow")
+	_SendMessage                 = user32.NewProc("SendMessageW")
 	_SetCapture                  = user32.NewProc("SetCapture")
 	_SetCursor                   = user32.NewProc("SetCursor")
 	_SetClipboardData            = user32.NewProc("SetClipboardData")
@@ -651,6 +666,42 @@ func GlobalUnlock(h syscall.Handle) {
 	_GlobalUnlock.Call(uintptr(h))
 }
 
+func CreateMutex(name string) (ptr syscall.Handle, alreadyExists bool, err error) {
+	r, _, err := _CreateMutex.Call(0, 0, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(name))))
+	switch err.(syscall.Errno) {
+	case 0:
+		return syscall.Handle(r), false, nil
+	case ERROR_ALREADY_EXISTS:
+		return syscall.Handle(r), true, nil
+	default:
+		return 0, false, fmt.Errorf("CreateMutex: %v", err)
+	}
+}
+
+func GetMutexInfo(h syscall.Handle) (pid, count uint32, err error) {
+	r, _, err := _GetMutexInfo.Call(uintptr(h), uintptr(unsafe.Pointer(&pid)), uintptr(unsafe.Pointer(&count)))
+	if r == 0 {
+		return 0, 0, fmt.Errorf("GetMutexInfo: %v", err)
+	}
+	return pid, count, nil
+}
+
+func OpenFileMapping(name string) (ptr syscall.Handle, err error) {
+	r, _, err := _OpenFileMapping.Call(FILE_MAP_READ, 0, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(name))))
+	if r == 0 {
+		return 0, fmt.Errorf("OpenFileMapping: %v", err)
+	}
+	return syscall.Handle(r), nil
+}
+
+func ReleaseMutex(h uintptr) error {
+	r, _, err := _ReleaseMutex.Call(h)
+	if r == 0 {
+		return fmt.Errorf("ReleaseMutex: %v", err)
+	}
+	return nil
+}
+
 func KillTimer(hwnd syscall.Handle, nIDEvent uintptr) error {
 	r, _, err := _SetTimer.Call(uintptr(hwnd), uintptr(nIDEvent), 0, 0)
 	if r == 0 {
@@ -735,8 +786,24 @@ func RegisterClassEx(cls *WndClassEx) (uint16, error) {
 	return uint16(a), nil
 }
 
+func RegisterWindowMessage(name string) (uint32, error) {
+	r, _, err := _RegisterWindowMessage.Call(uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(name))))
+	if r == 0 {
+		return 0, fmt.Errorf("RegisterWindowMessage: %v", err)
+	}
+	return uint32(r), nil
+}
+
 func ReleaseDC(hdc syscall.Handle) {
 	_ReleaseDC.Call(uintptr(hdc))
+}
+
+func SendMessage(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) error {
+	r, _, err := _SendMessage.Call(uintptr(hwnd), uintptr(msg), wParam, lParam)
+	if r == 0 {
+		return fmt.Errorf("SendMessage failed: %v", err)
+	}
+	return nil
 }
 
 func SetForegroundWindow(hwnd syscall.Handle) {
--- a/app/os_ios.m
+++ b/app/os_ios.m
@@ -99,6 +99,11 @@ CGFloat _keyboardHeight;
 	_keyboardHeight = 0.0;
 	[self.view setNeedsLayout];
 }
+
+- (BOOL)onDeeplink:(NSString *)url {
+    gio_onDeeplink((__bridge CFTypeRef)url);
+    return YES;
+}
 @end
 
 static void handleTouches(int last, UIView *view, NSSet<UITouch *> *touches, UIEvent *event) {
--- a/app/os_macos.go
+++ b/app/os_macos.go
@@ -7,7 +7,9 @@ package app
 
 import (
 	"errors"
+	"gioui.org/io/deeplink"
 	"image"
+	"net/url"
 	"runtime"
 	"time"
 	"unicode"
@@ -842,6 +844,23 @@ func gio_onFinishLaunching() {
 	close(launched)
 }
 
+var startupDeeplink *url.URL
+
+//export gio_onDeeplink
+func gio_onDeeplink(uri C.CFTypeRef) {
+	u, err := url.Parse(nsstringToString(uri))
+	if err != nil {
+		return
+	}
+	if len(viewMap) == 0 {
+		startupDeeplink = u
+		return
+	}
+	for _, w := range viewMap {
+		w.w.Event(deeplink.Event{URL: u})
+	}
+}
+
 func newWindow(win *callbacks, options []Option) error {
 	<-launched
 	errch := make(chan error)
@@ -867,6 +886,9 @@ func newWindow(win *callbacks, options []Option) error {
 		C.makeKeyAndOrderFront(window)
 		layer := C.layerForView(w.view)
 		w.w.Event(ViewEvent{View: uintptr(w.view), Layer: uintptr(layer)})
+		if startupDeeplink != nil {
+			w.w.Event(deeplink.Event{URL: startupDeeplink})
+		}
 	})
 	return <-errch
 }
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -3,9 +3,18 @@
 package app
 
 import (
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"gioui.org/io/deeplink"
+	syscall "golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
 	"image"
+	"io"
+	"net/url"
+	"os"
+	"path/filepath"
 	"runtime"
 	"sort"
 	"strings"
@@ -15,11 +24,8 @@ import (
 	"unicode/utf8"
 	"unsafe"
 
-	syscall "golang.org/x/sys/windows"
-
 	"gioui.org/app/internal/windows"
 	"gioui.org/unit"
-	gowindows "golang.org/x/sys/windows"
 
 	"gioui.org/f32"
 	"gioui.org/io/clipboard"
@@ -109,6 +115,9 @@ func newWindow(window *callbacks, options []Option) error {
 		w.w = window
 		w.w.SetDriver(w)
 		w.w.Event(ViewEvent{HWND: uintptr(w.hwnd)})
+		if startupDeeplink != "" {
+			w.processDeeplink(startupDeeplink)
+		}
 		w.Configure(options)
 		windows.SetForegroundWindow(w.hwnd)
 		windows.SetFocus(w.hwnd)
@@ -222,6 +231,10 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 	}
 
 	w := win.(*window)
+
+	if msg == windows.GIO_DEEPLINKING {
+		fmt.Println("GIO_DEEPLINK", msg, wParam, lParam)
+	}
 
 	switch msg {
 	case windows.WM_UNICHAR:
@@ -418,6 +431,29 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 	case windows.WM_IME_ENDCOMPOSITION:
 		w.w.SetComposingRegion(key.Range{Start: -1, End: -1})
 		return windows.TRUE
+	case windows.GIO_DEEPLINKING:
+		if schemesDeeplink == "" {
+			return windows.DefWindowProc(hwnd, msg, wParam, lParam)
+		}
+
+		size := wParam
+		if size >= 1<<16-1 {
+			return windows.TRUE
+		}
+
+		id := hex.EncodeToString(((*[unsafe.Sizeof(int(0))]byte)(unsafe.Pointer(&lParam)))[:])
+		f, err := os.Open(filepath.Join(os.TempDir(), "gio_deeplinking-"+string(id)))
+		if err != nil {
+			return windows.TRUE
+		}
+		defer f.Close()
+
+		b := make([]byte, size)
+		if _, err := io.ReadFull(f, b); err != nil {
+			return windows.TRUE
+		}
+
+		w.processDeeplink(string(b))
 	}
 
 	return windows.DefWindowProc(hwnd, msg, wParam, lParam)
@@ -652,7 +688,7 @@ func (w *window) readClipboard() error {
 		return err
 	}
 	defer windows.GlobalUnlock(mem)
-	content := gowindows.UTF16PtrToString((*uint16)(unsafe.Pointer(ptr)))
+	content := syscall.UTF16PtrToString((*uint16)(unsafe.Pointer(ptr)))
 	w.w.Event(clipboard.Event{Text: content})
 	return nil
 }
@@ -733,7 +769,7 @@ func (w *window) writeClipboard(s string) error {
 	if err := windows.EmptyClipboard(); err != nil {
 		return err
 	}
-	u16, err := gowindows.UTF16FromString(s)
+	u16, err := syscall.UTF16FromString(s)
 	if err != nil {
 		return err
 	}
@@ -850,6 +886,29 @@ func (w *window) raise() {
 		windows.SWP_NOMOVE|windows.SWP_NOSIZE|windows.SWP_SHOWWINDOW)
 }
 
+func (w *window) processDeeplink(uri string) {
+	fmt.Println("processDeeplink", uri)
+
+	u, err := url.Parse(uri)
+	if err != nil {
+		return
+	}
+
+	found := false
+	for _, scheme := range schemesDeeplinkList {
+		if u.Scheme == scheme {
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		return
+	}
+
+	w.w.Event(deeplink.Event{URL: u})
+}
+
 func convertKeyCode(code uintptr) (string, bool) {
 	if '0' <= code && code <= '9' || 'A' <= code && code <= 'Z' {
 		return string(rune(code)), true
@@ -952,6 +1011,134 @@ func configForDPI(dpi int) unit.Metric {
 		PxPerDp: ppdp,
 		PxPerSp: ppdp,
 	}
+}
+
+// schemesDeeplink is a list of schemes, comma separated, that most be
+// defined using -X compiler ldflag.
+var schemesDeeplink string
+var schemesDeeplinkList []string
+var startupDeeplink string
+
+func init() {
+	if schemesDeeplink == "" {
+		return
+	}
+	schemesDeeplinkList = strings.Split(schemesDeeplink, ",")
+
+	number, err := windows.RegisterWindowMessage(schemesDeeplink)
+	if err != nil {
+		return
+	}
+	windows.GIO_DEEPLINKING = number
+
+	/*
+		On Windows, launching the app using a deeplink will start a new instance of the app,
+		a new window. That behavior doesn't align with iOS/Android/macOS, where the deeplink
+		sends the event to the running app (if any).
+
+		In order to align the behavior, we use a global mutex to check if the app is already
+		running. If it is, we send the deeplink to the running app and exit the new instance.
+
+		That should happen on init to prevent the user from seeing the new window.
+	*/
+	if ok := processDeeplink(); ok {
+		os.Exit(0)
+		return
+	}
+
+	for _, scheme := range schemesDeeplinkList {
+		go registerDeeplink(scheme)
+	}
+}
+
+func registerDeeplink(scheme string) error {
+	path, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	key, _, err := registry.CreateKey(registry.CURRENT_USER, `Software\\Classes\\`+scheme, registry.ALL_ACCESS)
+	if err != nil {
+		return err
+	}
+	defer key.Close()
+
+	if err = key.SetStringValue("", "URL:"+scheme+" Protocol"); err != nil {
+		return err
+	}
+
+	if err = key.SetStringValue("URL Protocol", ""); err != nil {
+		return err
+	}
+
+	icon, _, err := registry.CreateKey(key, `DefaultIcon`, registry.ALL_ACCESS)
+	if err != nil {
+		return err
+	}
+	defer icon.Close()
+
+	if err = icon.SetStringValue("", `"`+path+`",1`); err != nil {
+		return err
+	}
+
+	cmd, _, err := registry.CreateKey(key, `shell\\open\\command`, registry.ALL_ACCESS)
+	if err != nil {
+		return err
+	}
+	defer cmd.Close()
+
+	// The app will launch with `-gio_deeplinking` as the first argument, and the deeplink as the second.
+	if err = cmd.SetStringValue("", `"`+path+`" -gio_deeplinking "%1"`); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func processDeeplink() (finishProcess bool) {
+	_, alreadyExists, err := windows.CreateMutex(schemesDeeplink)
+	if err != nil {
+		return false
+	}
+
+	if len(os.Args) < 3 || os.Args[1] != "-gio_deeplinking" {
+		// Only one instance of the app can run at a time.
+		// If the current instance doesn't have any deeplink to process, then
+		// it will exit if another instance is already running, or it will
+		// continue if it is the first instance.
+		return alreadyExists
+	}
+
+	u := os.Args[2]
+
+	if !alreadyExists {
+		// The mutex was created by this process, and we are the first instance.
+		startupDeeplink = u
+		return alreadyExists
+	}
+
+	id := make([]byte, unsafe.Sizeof(int(0)))
+	rand.Read(id)
+
+	// First, I tried to use mmap, but that plagues everything with unsafe pointers, and it's not worth it.
+	f, err := os.Create(filepath.Join(os.TempDir(), "gio_deeplinking-"+hex.EncodeToString(id)))
+	if err != nil {
+		return alreadyExists
+	}
+
+	if _, err := f.Write([]byte(u)); err != nil {
+		return alreadyExists
+	}
+
+	if err := f.Close(); err != nil {
+		return alreadyExists
+	}
+
+	if err := windows.SendMessage(windows.HWND_BROADCAST, windows.GIO_DEEPLINKING, uintptr(len(u)), *(*uintptr)(unsafe.Pointer(&id[0]))); err != nil {
+		return alreadyExists
+	}
+
+	return alreadyExists
 }
 
 func (_ ViewEvent) ImplementsEvent() {}
--- a/app/window.go
+++ b/app/window.go
@@ -5,6 +5,7 @@ package app
 import (
 	"errors"
 	"fmt"
+	"gioui.org/io/deeplink"
 	"image"
 	"image/color"
 	"runtime"
@@ -894,6 +895,8 @@ func (w *Window) processEvent(d driver, e event.Event) bool {
 	case ConfigEvent:
 		w.decorations.Config = e2.Config
 		e2.Config = w.effectiveConfig()
+		w.out <- e2
+	case deeplink.Event:
 		w.out <- e2
 	case event.Event:
 		handled := w.queue.q.Queue(e2)
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	golang.org/x/exp v0.0.0-20221012211006-4de253d81b95
 	golang.org/x/exp/shiny v0.0.0-20220827204233-334a2380cb91
 	golang.org/x/image v0.5.0
-	golang.org/x/sys v0.0.0-20220825204002-c680a09ffe64
+	golang.org/x/sys v0.10.0
 )
 
 require golang.org/x/text v0.7.0
--- a/io/deeplink/deeplink.go
+++ b/io/deeplink/deeplink.go
@@ -1,0 +1,12 @@
+package deeplink
+
+import (
+	"net/url"
+)
+
+// Event is generated when the app is opened from a deeplink, or when the app is already running and a deeplink is opened.
+type Event struct {
+	URL *url.URL
+}
+
+func (Event) ImplementsEvent() {}
--- a/app/framework_ios.h
+++ b/app/framework_ios.h
@@ -3,4 +3,5 @@
 #include <UIKit/UIKit.h>
 
 @interface GioViewController : UIViewController
+- (BOOL)onDeeplink:(NSString *)url;
 @end
--- a/app/os_android.go
+++ b/app/os_android.go
@@ -121,9 +121,11 @@ import "C"
 import (
 	"errors"
 	"fmt"
+	"gioui.org/io/deeplink"
 	"image"
 	"image/color"
 	"math"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -657,6 +659,16 @@ func Java_org_gioui_GioView_onClearA11yFocus(env *C.JNIEnv, class C.jclass, view
 	if w.semantic.focusID == w.semIDFor(virtID) {
 		w.semantic.focusID = 0
 	}
+}
+
+//export Java_org_gioui_GioView_onDeeplink
+func Java_org_gioui_GioView_onDeeplink(env *C.JNIEnv, class C.jclass, view C.jlong, uri C.jstring) {
+	w := cgo.Handle(view).Value().(*window)
+	u, err := url.Parse(goString(env, uri))
+	if err != nil {
+		return
+	}
+	w.callbacks.Event(deeplink.Event{URL: u})
 }
 
 func (w *window) initAccessibilityNodeInfo(env *C.JNIEnv, sem router.SemanticNode, off image.Point, info C.jobject) error {
--- a/app/os_ios.go
+++ b/app/os_ios.go
@@ -71,7 +71,9 @@ static struct drawParams viewDrawParams(CFTypeRef viewRef) {
 import "C"
 
 import (
+	"gioui.org/io/deeplink"
 	"image"
+	"net/url"
 	"runtime"
 	"runtime/debug"
 	"time"
@@ -131,6 +133,9 @@ func onCreate(view, controller C.CFTypeRef) {
 	w.Configure(wopts.options)
 	w.w.Event(system.StageEvent{Stage: system.StagePaused})
 	w.w.Event(ViewEvent{ViewController: uintptr(controller)})
+	if startupDeeplink != nil {
+		w.w.Event(deeplink.Event{URL: startupDeeplink})
+	}
 }
 
 //export gio_onDraw
@@ -349,6 +354,23 @@ func newWindow(win *callbacks, options []Option) error {
 }
 
 func osMain() {
+}
+
+var startupDeeplink *url.URL
+
+//export gio_onDeeplink
+func gio_onDeeplink(uri C.CFTypeRef) {
+	u, err := url.Parse(nsstringToString(uri))
+	if err != nil {
+		return
+	}
+	if len(views) == 0 {
+		startupDeeplink = u
+		return
+	}
+	for _, w := range views {
+		w.w.Event(deeplink.Event{URL: u})
+	}
 }
 
 //export gio_runMain
--- a/app/os_js.go
+++ b/app/os_js.go
@@ -373,6 +373,10 @@ func (w *window) keyEvent(e js.Value, ks key.State) {
 			State:     ks,
 		}
 		w.w.Event(cmd)
+
+		if cmd.Name == key.NameTab {
+			e.Call("preventDefault")
+		}
 	}
 }
 
--- a/app/os_macos.m
+++ b/app/os_macos.m
@@ -376,6 +376,7 @@ CFTypeRef gio_createView(void) {
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
 	[NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
 	[NSApp activateIgnoringOtherApps:YES];
+
 	gio_onFinishLaunching();
 }
 - (void)applicationDidHide:(NSNotification *)aNotification {
@@ -383,6 +384,11 @@ CFTypeRef gio_createView(void) {
 }
 - (void)applicationWillUnhide:(NSNotification *)notification {
 	gio_onAppShow();
+}
+- (void)application:(NSApplication *)application openURLs:(NSArray<NSURL *> *)urls {
+	for (NSURL *url in urls) {
+		gio_onDeeplink((__bridge CFTypeRef)url.absoluteString);
+	}
 }
 @end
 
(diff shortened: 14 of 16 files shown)


inkeliz:
_I will rebase soon and remove some debugs._

app/os_windows.go
```diff
@@ -954,4 +1007,132 @@ func configForDPI(dpi int) unit.Metric {
...
+
+	if err := f.Close(); err != nil {
+		return alreadyExists
+	}
+
+	if err := windows.SendMessage(windows.HWND_BROADCAST, windows.GIO_DEEPLINKING, uintptr(len(u)), *(*uintptr)(unsafe.Pointer(&id[0]))); err != nil {
+		return alreadyExists
+	}
+
+	return alreadyExists
```
inkeliz, line 1135: I think that doesn't feels right to add it in gio-core. However, doesn't make sense to left Windows with complete different behaviour.

inkeliz: I take a look at Flutter's implementation (https://pub.dev/packages/app_links), ironically they enforce the same limitation:

```
  // Dispatches link if any.
  // This method enables our app to be with a single instance too.
  // This is mandatory if you want to catch further links in same app.
  bool SendAppLinkToInstance(const std::wstring& title);
  ```

So, "enables our app to be with a single instance too" is required.  However, in the source-code they use WM_COPYDATA, instead of a custom event, and create a code that goes inside WM_COPYDATA. Maybe it's better? If I change it, then it will be simplified.

**Comment by inkeliz** [comment]
That panic was from gio-plugins, not from gio. I should not use gio-plugins/deeplink anymore, since everything was ported to Gio (this patch). I didn't use `panic` anywhere here (https://github.com/gioui/gio/pull/117/files). 

I park the code here, https://github.com/inkeliz/deeplink-demo. It uses the `replace` to point to my fork/pr.


```
git clone git@github.com:inkeliz/deeplink-demo.git
cd deeplink-demo 
go get gioui.org/cmd
go run gioui.org/cmd/gogio -target macos -deeplink test -o demo.app .
open demo_arm64.app
```

That will listen to `test://`. Then, you can open test://its-works to test.

File: app/GioActivity.java [hunk 2]
```diff
@@ -60,4 +62,9 @@ public final class GioActivity extends Activity {
 		if (!view.backPressed())
 			super.onBackPressed();
 	}
+
+	@Override protected void onNewIntent(Intent intent) {
+		super.onNewIntent(intent);
+        view.onIntentEvent(intent);
```
eliasnaur, line 68: Nit: the indentation seems off here.

app/GioView.java
```diff
@@ -311,6 +312,15 @@ private void setHighRefreshRate() {
 		window.setAttributes(layoutParams);
 	}
 
+	protected void onIntentEvent(Intent intent) {
```
eliasnaur, line 315: nit: Indentation seems off.

app/GioView.java
```diff
@@ -549,6 +559,7 @@ void updateCaret(float m00, float m01, float m02, float m10, float m11, float m1
 	static private native void onExitTouchExploration(long handle);
 	static private native void onA11yFocus(long handle, int viewId);
 	static private native void onClearA11yFocus(long handle, int viewId);
+	static private native void onDeeplink(long handle, String uri);
```
eliasnaur, line 562: Let's call this something like 'onOpenURI' or similar, here and elsewhere.

app/os_windows.go
```diff
@@ -954,4 +1007,132 @@ func configForDPI(dpi int) unit.Metric {
 	}
 }
 
+// schemesDeeplink is a list of schemes, comma separated, that most be
```
eliasnaur, line 1010: Nit: "must".

io/deeplink/deeplink.go
```diff
@@ -0,0 +1,12 @@
+package deeplink
+
+import (
+	"net/url"
+)
+
+// Event is generated when the app is opened from a deeplink, or when the app is already running and a deeplink is opened.
+type Event struct {
+	URL *url.URL
+}
```
eliasnaur, line 10: This seems to belong in `io/transfer`. Not sure whether it should be `transfer.DataEvent`.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	golang.org/x/exp v0.0.0-20221012211006-4de253d81b95
 	golang.org/x/exp/shiny v0.0.0-20220827204233-334a2380cb91
 	golang.org/x/image v0.5.0
-	golang.org/x/sys v0.0.0-20220825204002-c680a09ffe64
+	golang.org/x/sys v0.10.0
```
eliasnaur, line 13: Is this bump required? If so, let's do that as a separate change.

app/os_windows.go
eliasnaur: If `WM_COPYDATA` simplifies this, let's do that :) In either case, let's do the "easy" platforms first, then add Windows later.

io/deeplink/deeplink.go
inkeliz: The issue with `DataEvent` is that it uses ReadCloser. Of course, it's possible to use some `strings.NewReader(url)`, but I don't think that it's right. Maybe create a new `transfer.LinkEvent` or `transfer.URLEvent`?

inkeliz:
Just to remind me in the future:

- Domain names with accents (`ã`, `ç`, `ù` or emojis) have different behaviours. On Windows it's escapes similar to URL paths. On macOS/iOS it uses PunnyCode.
    - Document that behavior or try to match it (converting domain-accents to punnycode).
    - Maybe add tests for that? 
  
- Maximum URL size can vary for each OS. On iOS/macOS it can use upto 2GB, on Windows seems like 2048 characters, and CMD params have a limit of ~8K chars.
    - Add some documentation, assuming that anything < 2048 is safe to work.
    - Maybe, artificially enforce that limitation everywhere? Don't make sense, but makes easier to test.
 
 - Improve some documentation, on io/transfer (currently, io/deeplink). Telling that `-schemes` (currently, `-deeplink`) must be defined in gogio. Currently, Gio have one package and Event, but no clue about how to use that.

io/deeplink/deeplink.go
eliasnaur: `URLEvents` seems fine for now.

app/os_windows.go
```diff
@@ -1007,10 +1007,12 @@ func configForDPI(dpi int) unit.Metric {
...
 }
 
-// schemesDeeplink is a list of schemes, comma separated, that most be
+// schemesDeeplink is a list of schemes, comma separated, that must be
 // defined using -X compiler ldflag.
 var schemesDeeplink string
 var schemesDeeplinkList []string
+
+// startupDeeplink is the deeplink that started the app, if any.
 var startupDeeplink string
```
eliasnaur: Let's have all instances of `deeplink` replaced with `uri`, `openURI` or whatever makes sense. From the point of view of the Gio program, the feature is just about receiving URIs from outside.

app/os_windows.go
inkeliz: I tried to simplify the Windows implementation.

inkeliz:
> Hi @inkeliz, thanks a lot for implementing this feature.
> 
> Are there any chances of supporting Linux?

@mearaj, I don't think it's difficult to implement. However, personally, I don't use Linux, either my target-audience (it's < 1% that uses Linux). 

But, seems that you need to create a file, say `~/.local/share/applications`:

```
[Desktop Entry]
Name=MyApp
Exec=/path/to/your/program %U
Type=Application
StartupNotify=true
MimeType=x-scheme-handler/yourscheme
```
Then run `update-desktop-database` (https://www.unix.com/man-page/centos/1/update-desktop-database/). 

However, its works similar to Windows: that will open the app passing the URL as the first parameter. So, it will require to create some IPC to send the URL to the active window, and make it a single-instance (either by pipe or d-bus).

Currently, `gogio` don't support Linux. So, I'm not sure how will set those schemes. Since I have 0 experience with desktop-linux, I'm not going to work on that. At least, not now. :S

mearaj:
> > Hi @inkeliz, thanks a lot for implementing this feature.
> > Are there any chances of supporting Linux?
> 
> @mearaj, I don't think it's difficult to implement. However, personally, I don't use Linux, either my target-audience (it's < 1% that uses Linux).
> 
> But, seems that you need to create a file, say `~/.local/share/applications`:
> 
> ```
> [Desktop Entry]
> Name=MyApp
> Exec=/path/to/your/program %U
> Type=Application
> StartupNotify=true
> MimeType=x-scheme-handler/yourscheme
> ```
> 
> Then run `update-desktop-database` (https://www.unix.com/man-page/centos/1/update-desktop-database/).
> 
> However, its works similar to Windows: that will open the app passing the URL as the first parameter. So, it will require to create some IPC to send the URL to the active window, and make it a single-instance (either by pipe or d-bus).
> 
> Currently, `gogio` don't support Linux. So, I'm not sure how will set those schemes. Since I have 0 experience with desktop-linux, I'm not going to work on that. At least, not now. :S

No worries, I understand. Thank you so much for the reply and providing the helpful info. :)

mearaj:
> @mearaj
> 
> The groundwork laid down by the changes in gogio cmd by @inkeliz will mean its relatively easy to add linux support if you want to have a crack at it. I dont run linux and so not easy for me to do it and test it works.

HI @gedw99<br>,
Actually I am able to figure it out on my Arch Linux, the approach is the same as suggested by @inkeliz , but the issue is that it creates a separate instance of my app. I am trying to figure out the optimal solution. Once I find the best solution that works on most linux flavors(if not all), then I will most probably implement this feature in gio.
Thanks :)

gedw99:
If anyone is interested the Deeplinks, allows to then use Universal links.

I made an issue here showing how to do it. 

https://github.com/gioui-plugins/gio-plugins/issues/54

mearaj:
Hi @all,
I was able to implement this feature on my platform (Arch Linux). Need your help in how shall I proceed? It will take atleast 2-3 days for me to implement it in gogio and also I have no idea about it's implementation. Shall I wait for this commit to be merged or shall I create a separate PR against  [inkeliz:add-deeplink](https://github.com/inkeliz/gio/tree/add-deeplink) or whatever you say.

mearaj:
> Hi @ALL, I was able to implement this feature on my platform (Arch Linux) using @inkeliz suggested approach and unix socket. Need your help in how shall I proceed? It will take atleast 2-3 days for me to implement it in gogio and also this is my first time working with gogio specifically. Shall I wait for this commit to be merged or shall I create a separate PR against [inkeliz:add-deeplink](https://github.com/inkeliz/gio/tree/add-deeplink) or whatever you say.

I saw the gogio code, it doesn't supports linux as pointed out @inkeliz. I have decided to create a separate repo for this feature on linux. It will support both gio and non gio apps. Will post the link here. Thanks :)

gedw99:
@mearaj  

looking forward to see it. Maybe later merge it into gio cmd if your want.

inkeliz:
Another alternative is to create some app.Scheme("yourscheme") option. Problems: that only works with Linux/Windows, because macOS/iOS/Android requires some manifest, and can't register it on runtime. 

_Notice: My suggestion is NOT an `app.NewWindow` option. Because `app.Scheme` applies to instances, and not per-window. So, configuring it on `NewWindow` can be impossible and/or too late. It could be `app.Scheme` or `app.NewInstance(app.Scheme("yourscheme"))`._

My idea with `gogio` is to enforce the "compile-time", even when not required (Windows, and Linux). Since Gio usually tries to hide OS-specific APIs. If most OSes requires "compile-time" stuff (manifest...), then let's enforce it.

Another approach, is to read Golang's AST. You use `app.Scheme`, and only that. Then, `gogio` will search  `app.Scheme` usage (i.e `app.Scheme("abc")`) then gogio will compile with `abc` in Manifest. It's not perfect (conditions and so on), but that also opens `gogio` to alot of "hidden-magic", so it's seems great and terrible. I think @eliasnaur will not like that too.

-----

The easiest alternative:  create a private-variable and then use `ldflags` to define it. That already happens:

https://github.com/gioui/gio/blob/d62057a62e67f529cf40351f2c1a15fc2287fa2c/app/app.go#L27

In this patch, a new one:

https://github.com/gioui/gio/blob/fe42c61cd39d7d758762b35d7bcb536330881e1b/app/os_windows.go#L1016-L1018

It not requires `gogio`, but still required to set at compile-time (similar to Windows). I don't know how difficult is to port `gogio` to Linux. Must to create a single executable with some icon, and use the `ldflags` to set the variable. :S

mearaj:
> Another alternative is to create some app.Scheme("yourscheme") option. Problems: that only works with Linux/Windows, because macOS/iOS/Android requires some manifest, and can't register it on runtime. So, what happens on iOS/Android?!
> 
> _Notice: My suggestion is NOT an `app.NewWindow` option. Because `app.Scheme` applies to instances, and not per-window. So, configuring it on `NewWindow` can be impossible and/or too late. It could be `app.Scheme` or `app.NewInstance(app.Scheme("yourscheme"))`._
> 
> My idea with `gogio` is to enforce the "compile-time", even when not required (Windows, and Linux). Since Gio usually tries to hide OS-specific APIs. If most OSes requires "compile-time" stuff (manifest...), then let's enforce it.
> 
> Another approach, is to read Golang's AST. You use `app.Scheme`, and only that. Then, `gogio` will search `app.Scheme` usage (i.e `app.Scheme("abc")`) then gogio will compile with `abc` in Manifest. It's not perfect (conditions and so on), but that also opens `gogio` to alot of "hidden-magic", so it's seems great and terrible. I think @eliasnaur will not like that too.
> 
> The easiest alternative: create a private-variable and then use `ldflags` to define it. That already happens:
> 
> https://github.com/gioui/gio/blob/d62057a62e67f529cf40351f2c1a15fc2287fa2c/app/app.go#L27
> 
> In this patch, a new one:
> 
> https://github.com/gioui/gio/blob/fe42c61cd39d7d758762b35d7bcb536330881e1b/app/os_windows.go#L1016-L1018
> 
> It not requires `gogio`, but still required to set at compile-time (similar to Windows). I don't know how difficult is to port `gogio` to Linux. It just create a single executable with some icon, and use the `ldflags` to set the variable. :S

Hi @inkeliz , I mostly understood what you meant. I will try to implement it in gio itself by looking at the way you did for Windows. In case if I can't do it, I will share my implementation repo link here and will discuss with all of you guys :)

mearaj:
Hi @inkeliz,@gedw99 and everyone. [Desktop Entry Specs](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)<br> 
I could successfully do it on gio without gogio, but need your help further, also facing some issues currently.
Currently flow is as follows (Only for unix/linux)
### Before Window Creation:
* App checks if another instance of our app is running by trying to dial a connection and expects two type of errors. 
   * Error One: It says that file doesn't exists, in which case we assume that no other instance of our app is running.
     We install the app on user's platform. Start listening to socket connection and whenever any valid url data is received
     from the socket, we fire window.Event(transfer.URLEvent{URL: urlPtr})
   * Error Two: It says connection refuse, which implies that only socket file exists but another instance of our app is not 
     running  because it never refuses connection. Here we delete that file and continue with the above step.(Error One)
   * AnyOtherError: We exit the app. (This has no yet happened)
   * Err is nil: Here we are certain that the another instance of our app is running. We pass the arguments to that app
     and exit the app with error "another instance of this app is already running"
       
The issue currently I am facing is that I don't know how to proceed further. In the current state whenever app receives any uri, it fires window.Event(transfer.URLEvent{URL: urlPtr}).
Another issue is that on unix it's possible to pass multiple uri's. In the app.desktop entry file, in [Exec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables), there's %U for multiple uri's and %u for single uri. Is it possible to change the signature of type ```URLEvent struct {URL *url.URL}``` to ```URLEvent struct {URLs []*url.URL}```, this way on unix we would be able to share multiple uris in one request?

I will make a PR request at your branch.
Thanks :)

inkeliz:
> Another issue is that on unix it's possible to pass multiple uri's. In the app.desktop entry file, in [Exec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables), there's %U for multiple uri's and %u for single uri. Is it possible to change the signature of type URLEvent struct {URL *url.URL} to URLEvent struct {URLs []*url.URL}, this way on unix we would be able to share multiple uris in one request?

On macOS (or iOS) it also receives multiple URLs, that is translated to multiple `transfer.URLEvent`. Do you have any use-case when multiple URLs are triggered at same time?

mearaj:
> > Another issue is that on unix it's possible to pass multiple uri's. In the app.desktop entry file, in [Exec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables), there's %U for multiple uri's and %u for single uri. Is it possible to change the signature of type URLEvent struct {URL *url.URL} to URLEvent struct {URLs []*url.URL}, this way on unix we would be able to share multiple uris in one request?
> 
> On macOS (or iOS) it also receives multiple URLs, that is translated to multiple `transfer.URLEvent` (one URLEvent per URL). Do you have any use-case for multiple "bundled/combined" URLs?

Thanks for asking.
There are multiple reasons, 
* Android's intent like behavior on Unix/Linux apps, where we can send and receive multiple uris. 
* I am creating an firebase based frontend app where i need to send user credentials and other info as multiple uris to the same app on desktop. It will be very useful for communication between desktop and browser.
* On Unix/Linux, in the current implementation the app can receives uri directly from the socket or as arguments where each argument is a separate URI. Allowing multiple URI will give it a uniform way to tackle those.
* Most important reason is that it's more flexible as both options remains open. If app developer is interested in only single uri then he can just select the first element from the slice. If he is interested in multiple uris, then he can do that as well.

These are the reasons I can currently think of :)

inkeliz:
> I am creating a firebase based frontend app where i need to send user credentials and other info as multiple uris to the same app on desktop. It will be very useful for communication between desktop and browser.

Can you provide an example of how to open multiple URLs? Honestly, I've tried testing this using `window.Open` and `<a href>` in a browser, as well as testing with the browser/explorer address bar. In both cases, I'm able to open a single URL. I'm unsure how to open a list of URLs. Of course, I can open multiple URLs, but each one is independent from the others (multiple `transfer.URLEvent`). If you have `while(true) { window.Open("yourscheme://data")}` each call is a new `URLEvent`, since it only opens one URL per time.

On Android, `getData` returns a single URL (https://developer.android.com/reference/android/content/Intent#getData()), and the same is true for iOS and Windows.

Currently, only macOS (and Linux) provide a list of URLs, and I'm not sure how to test that.

-------

Changing to `[]*url.URL` is easy, but it can get tricky figuring out when they'll "bundle" or not. Since they will not always be "bundled", you need to treat both cases in the same way. In the end: You need to handle multiple `transfer.URLEvent` items and a single "combined" `transfer.URLEvent`.

mearaj:
Ahh yes, you are right. On linux I can convert multiple url into a single url . This way the app will have a common interface, so yeah your reason sounds very reasonable. So let's keep the way it is currently. I will make some changes in the current implementation and get back to you and will ask for your help again. Thanks a lot @inkeliz.

mearaj:
Hi @inkeliz , I have pushed changes to [PR](https://github.com/inkeliz/gio/pull/1).

app/window.go
```diff
@@ -895,6 +896,8 @@ func (w *Window) processEvent(d driver, e event.Event) bool {
 		w.decorations.Config = e2.Config
 		e2.Config = w.effectiveConfig()
 		w.out <- e2
+	case transfer.URLEvent:
+		w.out <- e2
```
eliasnaur, line 900: This is simple, but global. I believe `transfer.URLEvent`s should be routed like `transfer.DataEvent`. No?

app/os_windows.go
```diff
@@ -954,4 +987,95 @@ func configForDPI(dpi int) unit.Metric {
...
+	mmap, err := windows.OpenFileMapping(windows.FILE_MAP_WRITE, schemesURI)
+	if err != nil {
+		return
+	}
+	defer windows.CloseFileMapping(mmap)
+
+	dst, err := windows.MapViewOfFile(mmap, windows.FILE_MAP_WRITE, 1024*8)
+	if err != nil {
+		return
+	}
```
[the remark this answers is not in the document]
inkeliz: I tried to use `WM_COPYDATA` but it requires to know the specific HWND of the receiver. Also, in both cases it will need to have mutex, which `named-mmap` is also used for. 

The receiver creates the mmap using the scheme as the name of the mmap. The sender checks if the mmap already exists, if exists writes the URL into the mmap and then sends a event, that event don't contain any pointer or mmap.

app/window.go
inkeliz: I'm not sure. In the case of `DataEvent` it can target a specific area of the UI, right? Like, drag/drop at specific zone. In the case of `URLEvent` that is somewhat external. At least in my case, that will require to have `SomeOp{Tag: tag}` in the `main`. However, that already happens with the "Back Button" (which is now on `InputOp`).

I don't know if that should be routed, and if it should be a new `Op` or use `TargetOp`? The `MIME` field is strange in that case. Is `MIME` the scheme? What happens if want the listen "all schemes"?

eliasnaur: I think (most/all) events should be routed, for composability of components. Without routing, you can't just use some widget and forget about its internals. With routing, you don't need to know that the widget listens for a particular dnd or file open.

> I don't know if that should be routed, and if it should be a new Op or use TargetOp? The MIME field is strange in that case. Is MIME the scheme? What happens if want the listen "all schemes"?

This is a problem for dnd as well. I imagine the MIME type be deferred from the file extension, falling back to "application/octet-stream". I also imagine you being able to register a range of types with say "image/*" or similar.

See also https://lists.sr.ht/~eliasnaur/gio-patches/patches/42549.

app/os_windows.go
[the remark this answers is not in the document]
eliasnaur: > I tried to use WM_COPYDATA but it requires to know the specific HWND of the receiver.

If Windows opens a particular Gio binary, you should be able to locate any other running instances of that binary and only target them with WM_COPYDATA. For example, by the window class name or window name/id.

> Also, in both cases it will need to have mutex, which named-mmap is also used for.

See above, I don't think you need a mutex if you just target windows belonging to the binary being opened.

app/window.go
inkeliz: But in that case what "mime" is the "scheme"? I mean, how I listen to all schemes?`MIME: "schemes/*`? How I listen to a specific scheme? `MIME: "schemes/myCustomScheme`? 

Also, what happens when the app is launched using one URL/Scheme? Because, in that specific case, I think you don't have `TargetOp` setup yet. For each new `TargetOp` it sends the "startup-url"?

eliasnaur: > But in that case what "mime" is the "scheme"? I mean, how I listen to all schemes?`MIME: "schemes/*`? How I listen to a specific scheme? `MIME: "schemes/myCustomScheme`?
> 

Good points. A `SchemeOp` seems appropriate.

> Also, what happens when the app is launched using one URL/Scheme? Because, in that specific case, I think you don't have `TargetOp` setup yet. For each new `TargetOp` it sends the "startup-url"?

The incoming URL must be retined by the `app` package until one frame has been processed, to give the program a chance to receive the URL. After that, it can be forgotten.

inkeliz: I add a new commit with SchemeOp. I create a another `router/transfer.go`, since the current `TargetOp`seems to be integrated with `clip-area`.

app/os_windows.go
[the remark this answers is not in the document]
inkeliz: I found one issue related to UNC. I tried to search by path, then send COPYDATA to that executable. 

For instance, it's possible to have one `X:/demo.exe` and `//192.168.0.1/demo.exe`, the first one is one "map" to the second one. Here is the issue, because I can't use `os.Executable() == anotherExePath`.

Because of that edge-case, I'm comparing the `filepath.Base`. However, `Base` may cause collision (for instance your/app.exe is the same of randompath/app.exe). 

I think another alternative is to use `-appid` (?) from `gogio` and also checks the `WindowClassName` (which is currently hardcoded as "Gio"), not just the `filepath.Base`.

io/router/transfer.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (q *transferQueue) Clear() {
+	if q.schemes != nil {
+		q.schemes = q.schemes[:0]
+	}
+	if q.initialized {
+		q.firstURLEvent = transfer.URLEvent{}
+	}
+	q.initialized = true
```
inkeliz, line 44: That makes possible to hold the "StartupURL" for one frame. I'm not sure if that is the best way, but that doesn't require change the existent code

app/internal/windows/windows.go
```diff
@@ -303,15 +316,31 @@ const (
...
 	LR_VGACOLOR         = 0x00000080
+
+	FILE_MAP_WRITE = 0x0002
+	FILE_MAP_READ  = 0x0004
+
+	PAGE_READWRITE = 0x00000004
+)
+
+const (
+	GIO_OPEN_URL uintptr = iota + 1 // Custom message
```
eliasnaur, line 327: This belongs in `os_windows.go`, no? Also, I suggest naming it similar to the existing `_WM_WAKEUP`.

app/os_windows.go
```diff
@@ -954,4 +994,64 @@ func configForDPI(dpi int) unit.Metric {
...
+	/*
+		On Windows, launching the app using a URI will start a new instance of the app,
+		a new window. That behavior doesn't align with iOS/Android/macOS, where the deeplink
+		sends the event to the running app (if any).
+
+		In order to align the behavior, we use a global mutex to check if the app is already
+		running. If it is, we send the deeplink to the running app and exit the new instance.
+
+		That should happen on init to prevent the user from seeing the new window.
+	*/
```
eliasnaur, line 1030: The usual Go style is to use `//` for comment blocks.

app/internal/windows/windows.go
```diff
@@ -814,3 +922,71 @@ func (p *WindowPlacement) Set(Left, Top, Right, Bottom int) {
...
+		}
+
+		return 1
+	})
+
+	_EnumWindows.Call(fn, 0)
+	return resHWND, resHWND != 0
+}
+
+func RegisterScheme(scheme string) error {
```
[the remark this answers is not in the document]
inkeliz: Can you explain how `gogio` can improve that? I mean, how can we "bundle" two apps or set such Registry.

I don't like the `RegisterScheme` too, and the "IPC" thing. But, considering that Gio tries to map features across each OS as similar as possible, forcing Windows to have a installer doesn't match against macOS. Also, `gogio` only emits a single binary (`.exe`), and don't generate other installers (such as `.appx` or `.msix`).

On macOS a single `.app` is responsible to register itself, due to `plist`. On Windows, such registration is done by calling `RegisterScheme`, and then we need build something to send messages.

I agree that using a Gio executable to register and also send messages is not ideal. But, that seems to be the closest behaviour (vs macOS).

app/os_windows.go
```diff
@@ -954,4 +994,64 @@ func configForDPI(dpi int) unit.Metric {
 	}
 }
 
+// schemesURI is a list of schemes, comma separated, that must be
+// defined using -X compiler ldflag, that used in gogio.
+var schemesURI string
+var schemesURIList []string
+
+// startupURI is the URI that started the app, if any.
+var startupURI string
```
eliasnaur, line 1003: Do this need to be global? I don't see it used outside of the following `init`.

app/internal/windows/windows.go
```diff
@@ -814,3 +922,71 @@ func (p *WindowPlacement) Set(Left, Top, Right, Bottom int) {
...
+	path, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	key, _, err := registry.CreateKey(registry.CURRENT_USER, `Software\\Classes\\`+scheme, registry.ALL_ACCESS)
```
eliasnaur, line 961: Does this registry overwrite other programs handling the particular scheme?

app/internal/windows/windows.go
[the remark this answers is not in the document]
eliasnaur: Ok, I suppose this code can stay, because it is not triggered without `gogio` embedding one or more schemes.

Can we clean up registries no longer needed, to mimic the macOS plist behaviour? In other words, have the registry store the `appid` or whatever you need to identify windows belonging to the same process, and then delete schemes no longer included in `schemesURI`.

app/os_windows.go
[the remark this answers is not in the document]
inkeliz: I notice that `appID` is set in `internal/log`, and used only for Android. I create a new variable, in `os_windows.go`, but seems that both have the same purpose.

Maybe we should create some `app/meta/meta.go` which contains `AppID`, `Version`, `URLSchemes` (...)? That makes easier to set variables (without gogio) and also re-use it anywhere.

I will leave as duplicated, and in some next patch create such package.

app/os_windows.go
inkeliz: It's used inside newWindow, and the URLEvent is sent if `startupURI` is not empty.

app/internal/windows/windows.go
```diff
@@ -814,3 +926,124 @@ func (p *WindowPlacement) Set(Left, Top, Right, Bottom int) {
 	p.rcNormalPosition.Right = int32(Right)
 	p.rcNormalPosition.Bottom = int32(Bottom)
 }
+
+func RegisteredSchemes(appid string) []string {
```
eliasnaur, line 930: These functions doesn't seem to belong here, a package for Windows syscalls. Why not have them in package `app`?

app/os_windows.go
```diff
@@ -86,6 +90,9 @@ var resources struct {
 	cursor syscall.Handle
 }
 
+// appID is the application unique identifier, set using `-appid` in Gogio.
+var appID = "GioWindow"
```
eliasnaur, line 94: What about `app.ID`? I agree that `app/internal/log.appID` should also be merged with this, but doesn't have to be in this PR.

app/os_windows.go
```diff
@@ -954,4 +1001,66 @@ func configForDPI(dpi int) unit.Metric {
...
+		for i, o := range oldSchemes {
+			if s == o {
+				oldSchemes = append(oldSchemes[:i], oldSchemes[i+1:]...)
+				break
+			}
+		}
+	}
+
+	if len(oldSchemes) > 0 {
+		go windows.UnregisterSchemes(appID, oldSchemes)
```
eliasnaur, line 1028: Why the goroutine and not a call to windows.UnregisterSchemes?

app/os_windows.go
```diff
@@ -954,4 +1001,66 @@ func configForDPI(dpi int) unit.Metric {
...
+	}
+
+	if len(currentSchemes) == 0 {
+		return
+	}
+
+	if len(os.Args) >= 3 && os.Args[1] == "-gio_launch_url" {
```
eliasnaur, line 1035: Use equality, `len(os.Args) == 3`, so there's no doubt about Windows launching us, not the user.

app/os_windows.go
```diff
@@ -954,4 +1001,66 @@ func configForDPI(dpi int) unit.Metric {
...
+		startupURI = os.Args[2]
+	}
+
+	//	On Windows, launching the app using a URI will start a new instance of the app,
+	//	a new window. That behavior doesn't align with iOS/Android/macOS, where the deeplink
+	//	sends the event to the running app (if any).
```
eliasnaur, line 1041: Nit: too much indentation in the comment.

io/router/transfer.go
```diff
@@ -0,0 +1,45 @@
+package router
+
+import (
+	"gioui.org/io/transfer"
+	"strings"
+)
+
+type transferQueue struct {
+	initialized   bool
+	schemes       []transfer.SchemeOp
```
eliasnaur, line 10: Nit: maybe rename field to "handlers"?

io/router/transfer.go
```diff
@@ -0,0 +1,45 @@
...
+	firstURLEvent transfer.URLEvent
+}
+
+func (q *transferQueue) Push(evt transfer.URLEvent, events *handlerEvents) {
+	if !q.initialized && q.firstURLEvent.URL == nil {
```
eliasnaur, line 15: Why not drop `q.initialized` and replace every check with `q.firstURLEvent.URL == nil`?

app/internal/windows/windows.go
```diff
@@ -303,15 +315,27 @@ const (
...
+	_GetModuleHandleW          = kernel32.NewProc("GetModuleHandleW")
+	_GlobalAlloc               = kernel32.NewProc("GlobalAlloc")
+	_GlobalFree                = kernel32.NewProc("GlobalFree")
+	_GlobalLock                = kernel32.NewProc("GlobalLock")
+	_GlobalUnlock              = kernel32.NewProc("GlobalUnlock")
+	_MapViewOfFile             = kernel32.NewProc("MapViewOfFile")
+	_OpenFileMapping           = kernel32.NewProc("OpenFileMappingW")
+	_ReleaseMutex              = kernel32.NewProc("ReleaseMutex")
+	_UnmapViewOfFile           = kernel32.NewProc("UnmapViewOfFile")
+	_QueryFullProcessImageName = kernel32.NewProc("QueryFullProcessImageNameW")
```
eliasnaur, line 338: Some of these are no longer used. We should delete them to avoid bit rot.

app/os_windows.go
inkeliz: I put it just to prevent some slowdown, during startup.

eliasnaur: Ok How much slowdown? Please add a comment saying why a goroutine is desired and ok.

app/os_windows.go
eliasnaur: Sounds good to me.

app/internal/windows/windows.go
```diff
@@ -303,6 +313,11 @@ const (
 	LR_MONOCHROME       = 0x00000001
 	LR_SHARED           = 0x00008000
 	LR_VGACOLOR         = 0x00000080
+
+	FILE_MAP_WRITE = 0x0002
+	FILE_MAP_READ  = 0x0004
+
+	PAGE_READWRITE = 0x00000004
```
eliasnaur, line 320: Nit: remove these if unused.

app/os_windows.go
```diff
@@ -850,6 +873,28 @@ func (w *window) raise() {
...
+func (w *window) onOpenURI(uri string) bool {
+	u, err := url.Parse(uri)
+	if err != nil {
+		return false
+	}
+
+	found := false
+	for _, scheme := range strings.Split(schemesURI, ",") {
+		if u.Scheme == scheme {
+			found = true
```
eliasnaur, line 885: Instead of `found`, just call `w.w.Event` here and return true.

io/router/transfer.go
```diff
@@ -0,0 +1,43 @@
+package router
+
+import (
+	"gioui.org/io/transfer"
+	"strings"
+)
+
+type transferQueue struct {
+	handlers      []transfer.SchemeOp
+	firstURLEvent transfer.URLEvent
```
eliasnaur, line 10: After further thought, I think `firstURLEvent` and the delaying of it belongs in app.Window. I don't think the router package should know about the quirk that the first event is delayed.

io/transfer/transfer.go
```diff
@@ -107,3 +110,39 @@ type DataEvent struct {
 }
 
 func (DataEvent) ImplementsEvent() {}
+
+// SchemeOp registers a tag as a lister for a scheme updates.
+// Use multiple SchemeOps if a tag supports multiple schemes,
+// or use an empty scheme to listen for all schemes.
+//
+// The app must be compiled with the `-schemes <scheme>` flag,
+// in gogio, to be able to receive URLEvents.
```
eliasnaur, line 119: Nit:

// Note that only schemes registered with the platform receive URLEvents. Use a packaging tool
// such as [gioui.org/cmd/gogio] to register schemes.

eliasnaur (changes requested):

eliasnaur:
I haven't forgotten this, but I'm waiting for https://todo.sr.ht/~eliasnaur/gio/550 to be resolved. In particular, with #550 I believe you no longer need to retain the first event anywhere.

Inemsamuel:
I need a job

eliasnaur:
Ok, #550 and #555 are now merged. Sorry for the delay and merge conflicts.

inkeliz:
its-fine-dog-fine

eliasnaur:
image

gedw99:
> Ok, #550 and #555 are now merged. Sorry for the delay and merge conflicts.

I am trying to get https://github.com/gioui-plugins/gio-plugins/tree/add-signin working as well as a few other plugins at https://github.com/orgs/gioui-plugins/projects/1

, so it would be awesome if this Deep linking could be finally merged.

inkeliz:
I don't think we have any "blockers." The only "blocker" is "time." I need to upgrade everything (my app, gio-plugins, gio-cmd, and this change) to the new Gio. The gio-plugins rely on unsafe to modify Gio internals (events). Some plugins also support only Go 1.20, which needs update as well (because Go 1.22 has been released). To make matters worse, I am currently working on my final project at university, which is also why I am not very active on Gio anymore.

Additionally, the "sign-in" and "deeplink" features require changes to gio-cmd, along with a bunch of changes that are coupled together (https://github.com/gioui/gio-cmd/pull/9, https://github.com/gioui/gio-cmd/pull/13, and https://github.com/gioui/gio-cmd/pull/14). I also need to rebase those changes due to conflicts, if @eliasnaur can review some of those changes will be great.

gedw99:
> university

Thanks for the summary and the information on what needs to be done.

Definitely prioritise your University - Good luck with it ..

eliasnaur:
> Additionally, the "sign-in" and "deeplink" features require changes to gio-cmd, along with a bunch of changes that are coupled together ([gioui/gio-cmd#9](https://github.com/gioui/gio-cmd/pull/9), [gioui/gio-cmd#13](https://github.com/gioui/gio-cmd/pull/13), and [gioui/gio-cmd#14](https://github.com/gioui/gio-cmd/pull/14)). I also need to rebase those changes due to conflicts, if @eliasnaur can review some of those changes will be great.

I'm sorry to hear that you're still waiting for review for so long. If you think I missed a PR, feel free to ping me in a GH comment.

As for the 3 PRs:

- I've commented on gioui/gio-cmd#13 which I missed because GH doesn't notify on status changes.
- gioui/gio-cmd#14 is still in draft, which I take as a signal it's not ready for review.
- gioui/gio-cmd#9 looks fine in general to me, but I'd rather not merge it until this PR (#117) is ready to merge.

inkeliz:
I'm porting everything to Gio 0.6 (#136  or #134 still a major blocker).

But, far I notice the `_Something_Filter` usually contains some `Tag`. In the particular case of "deeplinking", it doesn't have any source/target, since it's received from the OS itself. This current PR assumes that it have one `Op` to watch for events, but in Gio 0.6 sounds like `Op` are used for features that are bounded by `clip.Rect` (or other ops). Using `Cmd` don't see ideal too, because it's not an immediate command, it don't perform any task.

In general, what should I do in that case? Create a new `transfer.SchemeFilter` and `transfer.SchemeEvent`? In that case `SchemeFilter` don't contains any `Tag`, only (optionally) `Scheme: string`.

gedw99:
Hey @inkeliz 

Yep I just tried it. Almost works too for me

```sh

# [github.com/gioui-plugins/gio-plugins/plugin/gioplugins](http://github.com/gioui-plugins/gio-plugins/plugin/gioplugins)
../../plugin/gioplugins/gioplugins.go:62:8: undefined:
input.SourceEventProcessor
../../plugin/gioplugins/gioplugins.go:71:8: undefined:
input.SourceExecuteProcessor
../../plugin/gioplugins/gioplugins.go:77:5: r.Execute undefined (type
*input.Router has no field or method Execute)

I am guessing you know about this ?

```

inkeliz:
@gedw99, I think you are talking about https://github.com/gioui-plugins/gio-plugins/pull/69, which is unrelated to this PR. 

In case of the gio-plugin you need to use https://github.com/gioui/gio/pull/136 version of Gio, for now.

inkeliz:
@eliasnaur I need to squash multiple commits, due to rebase. I test have tested it on macOS, Android, iOS and Windows.

I tried to avoid re-introducing the old `views` map/list, but since the URL is received from the main app, not to a specific view, I didn't find a better way to route such event. :\

app/os_macos.go
```diff
@@ -334,6 +336,17 @@ type window struct {
 // launched is closed when applicationDidFinishLaunching is called.
 var launched = make(chan struct{})
 
+// activeViews is the list of active windows.
+var activeViews = make([]*window, 0, 1)
+
+// mutexActiveViews protects activeViews.
+var mutexActiveViews = sync.Mutex{}
```
eliasnaur, line 343: Is a mutex necessary? `activeViews` seems accessed from the main thread only.

io/input/pointer.go
```diff
@@ -320,6 +329,12 @@ func (p *pointerFilter) Matches(e event.Event) bool {
 				return true
 			}
 		}
+	case transfer.URLEvent:
+		for _, t := range p.scheme {
+			if e.URL != nil && (t == "" || t == e.URL.Scheme) {
```
eliasnaur, line 334: I think we should assume `e.URL != nil`.

io/input/router.go
```diff
@@ -215,6 +215,8 @@ func (q *Router) Event(filters ...event.Filter) (event.Event, bool) {
 			t = f.Target
 		case transfer.TargetFilter:
 			t = f.Target
+		case transfer.URLFilter:
+			t = q // See comment in processEvent.
```
eliasnaur, line 219: Why not treat `transfer.URLFilter` the same as the tag-less `key.Filter` above, instead of inventing a fake tag?

io/input/router.go
```diff
@@ -464,6 +468,10 @@ func (q *Router) processEvent(e event.Event, system bool) {
 		cstate, evts := q.cqueue.Push(state.clipboardState, e)
 		state.clipboardState = cstate
 		q.changeState(e, state, evts)
+	case transfer.URLEvent:
+		var evts []taggedEvent
+		evts = append(evts, taggedEvent{tag: q, event: e})
```
eliasnaur, line 473: Like `key.Event` above, I would not expect a `transfer.URLEvent` to have an associated tag.

io/transfer/transfer.go
```diff
@@ -57,6 +58,15 @@ type TargetFilter struct {
 	Type string
 }
 
+// URLFilter filters for any [URLEvent] which the scheme matches against
+// the scheme from the event.
+type URLFilter struct {
+	// If it's blank/empty, it will handle any scheme.
+	Scheme string
+}
```
eliasnaur, line 66: I suggest

// URLFilter filters for [URLEvent].
type URLFilter struct {
	// Scheme must equal the URLEvent scheme to match. The empty Scheme matches all URLEvents.
	Scheme string
}

io/transfer/transfer.go
```diff
@@ -91,5 +101,22 @@ type DataEvent struct {
 
 func (DataEvent) ImplementsEvent() {}
 
+// URLEvent is generated when the app is invoked with a URL. This event is sent to
+// all SchemeOps that match the URL scheme.
+//
+// In order to receive URLEvents, it's necessary to use SchemeOp and the app must
+// register a scheme. The scheme can be registered using gogio, with `-schemes <scheme>`
+// flag.
```
eliasnaur, line 109: This comments needs updating. For example

// URLEvent is generated when the program is invoked with a URL. It is matched
// by [URLFilter].
//
// In order to receive the program must register a scheme. The scheme can be registered
// using gogio, with the `-schemes` flag.

app/os_windows.go
```diff
@@ -993,3 +1031,187 @@ func configForDPI(dpi int) unit.Metric {
...
+	}
+
+	if len(currentSchemes) == 0 {
+		return
+	}
+
+	// On Windows, launching the app using a URI will start a new instance of the app,
+	// a new window. That behavior, by default, doesn't align with iOS/Android/macOS, where
+	// the deeplink sends the event to the running app (if any). We are emulating it.
+	if hwnd, _ := windows.FindWindow(ID); hwnd != 0 {
```
eliasnaur, line 1067: What if multiple program instances have windows matching the ID?

app/os_windows.go
```diff
@@ -59,7 +61,10 @@ type window struct {
 	invMu sync.Mutex
 }
 
-const _WM_WAKEUP = windows.WM_USER + iota
+const (
+	_WM_WAKEUP = windows.WM_USER + iota
+	_WM_OPEN_URL
```
eliasnaur, line 66: The `_WM_OPEN_URL` name and value is misleading because it's not a window message identifier like `WM_PAINT` or our `_WM_WAKEUP`. It should probably be something like

`const copyDataURLType = 0xffffff00`

with a sufficiently large constant that minimizes the risk of collision.

eliasnaur (changes requested):
Looks great overall. A few nits and questions in the comments.

app/os_windows.go
inkeliz: Previously the executable path was used as identifier, then moved to `appid` (https://github.com/gioui/gio/pull/117#discussion_r1300420694). 

Maybe in `gogio` it must throw one error if `appid` is not provided and `schemes` are. 

Assuming that `appid` is unique only one instance of this app will run.

io/input/router.go
inkeliz: Because `q.key.scratchFilter = append(q.key.scratchFilter, f)` is a specific slice that holds key.Filter. Where the "generic one" (used by anything else) uses "taggedFilter". I don't think is right to add yet another specialised slice.

eliasnaur:
I wanted to merge this soon, and thought more about the corner-cases. I'm having cold feet with respect to exposing deeplinking support as `io/transfer.URLFilter`, for two reasons:

1. Deeplink events feels like something handled by the app, not any particular window event loop.
2. More importantly, programs on the macOS platform may not even have any windows open to handle the deeplink event. In other words, you want to open a new window when a deeplink event arrives, but there's no open windows to detect it.

I don't have better ideas than to back out the `io/transfer` changes and add a `app.Events` call:

```go
package app

// Events is an iterator that yields events that are not specific to any window,
// such as [URLEvent]. It never returns.
//
// Events must be called by the main goroutine, and replaces the
// call to [Main].
func Events(yield func(event.Event) bool)
```

and `app.Main` rewritten to:

```go
package app

func Main() {
    Events(func(event.Event){})
}
```

(`app.Main` will probably be removed in a future version).

eliasnaur:
One consequence of a global `app.Events` is that the Windows implementation probably needs to create and drive a hidden window just for the purpose of receiving deeplink events.

inkeliz:
Originally, the event was sent directly to window.Event (instead of gtx.Event).

I find it odd to create another event handler (window.Event, layout.Context.Event, and now app.Events). While I understand the reasoning, since this Event is not associated with a specific Window, I think it's just too much. Also, some OSes (Android and Switch) it's somewhat bounded to the View/Window.

eliasnaur:
How do you propose delivering the deeplink events to a running programs without open windows?

barrowkwan:
does gio community consider to merge this to main?  I am testing out golang to write mobile app and working on an Oauth ( keycloak ) authentication which will do a "callback", normally we will define custom scheme ( eg  myapp://callback ) to receive the token back.  it will be helpful to have the deeplink support.  

@inkeliz is it easy to make the deeplink as a separate package ?

eliasnaur:
> I also notice another issue, after more tests with the latest version (the iterator without channel):
> 
> ```
>  panic: runtime error: range function continued iteration after whole loop exit
> 12-13 23:20:13.695 29018     0 E Go      : go
> ```
> 

Have you dug deeper into this? I'm guessing something calls `yield` after it has returned `false` (because the loop exited with a `break` or because of an unrelated panic).

I think we should keep the `iter.Seq` signature. In the worst case, it should be possible to use a channel and goroutine to feed it.

inkeliz:
I change it again, I make two commits to make easier to review. 
I have tested on Android (physical device), Windows, iOS (emulator) and macOS.

The `w.Invalidated()` still causing issues and needs to run a new goroutine.

inkeliz:
I'm still testing...

eliasnaur (commented):
A question about the shared instance model on Windows below.

app/os_windows.go
```diff
@@ -1052,3 +1089,195 @@ func getPointerButtons(pi windows.PointerInfo) pointer.Buttons {
...
+		go unregisterSchemes(ID, oldSchemes)
+	}
+
+	if len(currentSchemes) == 0 {
+		return
+	}
+
+	// On Windows, launching the app using a URI will start a new instance of the app,
+	// a new window. That behavior, by default, doesn't align with iOS/Android/macOS, where
+	// the deeplink sends the event to the running app (if any). We are emulating it.
```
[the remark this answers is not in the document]
eliasnaur: Fair points, thanks.

app/internal/windows/windows.go
```diff
@@ -767,8 +791,11 @@ func SetWindowPos(hwnd syscall.Handle, hwndInsertAfter uint32, x, y, dx, dy int3
 }
 
 func SetWindowText(hwnd syscall.Handle, title string) {
-	wname := syscall.StringToUTF16Ptr(title)
-	_SetWindowText.Call(uintptr(hwnd), uintptr(unsafe.Pointer(wname)))
+	wname, err := syscall.UTF16FromString(title)
+	if err != nil {
+		return
```
eliasnaur: I think an explicit `panic(err)` is appropriate here. Ignoring the title change is more confusing.

app/internal/windows/windows.go
```diff
@@ -767,8 +791,11 @@ func SetWindowPos(hwnd syscall.Handle, hwndInsertAfter uint32, x, y, dx, dy int3
 }
 
 func SetWindowText(hwnd syscall.Handle, title string) {
-	wname := syscall.StringToUTF16Ptr(title)
-	_SetWindowText.Call(uintptr(hwnd), uintptr(unsafe.Pointer(wname)))
+	wname, err := syscall.UTF16FromString(title)
```
eliasnaur: Ditto.

app/internal/windows/windows.go
```diff
@@ -576,6 +588,18 @@ func EmptyClipboard() error {
 	return nil
 }
 
+func FindWindow(lpClassName string) (syscall.Handle, error) {
+	className, err := syscall.UTF16FromString(lpClassName)
```
eliasnaur: Ditto.

app/os_windows.go
```diff
@@ -132,13 +139,19 @@ func initResources() error {
 	}
 	resources.cursor = c
 	icon, _ := windows.LoadImage(hInst, iconID, windows.IMAGE_ICON, 0, 0, windows.LR_DEFAULTSIZE|windows.LR_SHARED)
+
+	appid, err := syscall.UTF16FromString(ID)
```
eliasnaur: Ditto.

app/os_windows.go
```diff
@@ -108,6 +112,9 @@ func newWindow(win *callbacks, options []Option) {
 		defer winMap.Delete(w.hwnd)
 		w.Configure(options)
 		w.ProcessEvent(Win32ViewEvent{HWND: uintptr(w.hwnd)})
+		if u := startupURI(); u != "" {
+			w.onOpenURI(u)
+		}
 		windows.SetForegroundWindow(w.hwnd)
```
eliasnaur: This doesn't seem right. It triggers for every new window, right? I think `startupURI` should be checked when `app.Events` is called.

eliasnaur (commented):
Getting there. I believe the iOS check,

```go
	if !isMainThread() {
		panic("app.Main must be run on the main goroutine")
	}
```

should be moved into the `mainModeUndefined` case. The other modes no longer run on the main thread.

eliasnaur:
@whereswaldon any comments on this and https://github.com/gioui/gio-cmd/pull/9?

app/os_ios.m
```diff
@@ -283,16 +288,20 @@ void gio_viewSetHandle(CFTypeRef viewRef, uintptr_t handle) {
...
 	self.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
-	GioViewController *controller = [[GioViewController alloc] initWithNibName:nil bundle:nil];
-	self.window.rootViewController = controller;
+	self.controller = [[GioViewController alloc] initWithNibName:nil bundle:nil];
+	self.window.rootViewController = self.controller;
 	[self.window makeKeyAndVisible];
 	return YES;
 }
+- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options {
+	return [self.controller onOpenURI:url.absoluteString];
```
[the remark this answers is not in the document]
inkeliz: I don't remember the reason, I will try to move it.

app/app.go
```diff
@@ -56,6 +58,20 @@ type FrameEvent struct {
...
 
+// URLEvent is generated when the program is invoked with a URL. This event is delivered
+// through Event iterator and not through a Window.Event.
+//
+// In order to receive the program must register a scheme. The scheme can be registered
+// using gogio, with the `-schemes` flag.
+//
+// Due to differences between OSes, URLEvent might not be generated when the URL is
+// larger than 2048 characters. Domain names using non-ASCII characters have
+// undefined behavior, it can be either encoded as Punycode, QueryEscape,
```
eliasnaur: Saying "undefined behaviour" and then listing possible encodings is not great for the user. Also, which platform(s) encodes non-ASCII domains as "QueryEscape"? Can we decode those to UTF-8, and then say

"Domain names using non-ASCII characters may be encoded in UTF-8 or punycode"

?

In other words, I'd like to expose only the minimal possible encodings to the user. *someone* has to bear the decoding burden and it might as well be Gio.

PR #119, opened 2023-08-12
Add openURL cross-platform

g45t345rt:
Android - `Intent object`
Windows - `exec command "cmd start"`
Unix - `exec command "xdg-open"`
Macos - `exec command "open"`
IOS - `UIApplication openURL`
JS - `window.open`

### DIFF
--- a/app/GioView.java
+++ b/app/GioView.java
@@ -12,10 +12,12 @@ import android.app.Fragment;
 import android.app.FragmentManager;
 import android.app.FragmentTransaction;
 import android.content.Context;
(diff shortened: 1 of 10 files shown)


eliasnaur:
I've left a few minor review comments, but my main objection is whether this change belongs in Gio core. It seems to be implementable in a separate module, possibly through `app.AppContext`. Perhaps gio-x (CC @whereswaldon).

whereswaldon:
I would be happy to accept this functionality within gio-x.

PR #120, opened 2023-08-16
widget: click button only if key pressed and released

vsariola:
This commit fixes the non-intuitive behaviour, where hitting return or space with a button focused, then tabbing to another button and releasing the key causes the second button to trigger. It feels wrong, as the "gesture" was never initiated on the second button. The fix makes widget.Clickable track which key was pressed, in a variable called pressedKey, and only considers a key release if the released key matches the pressed key. Finally, if the widget loses focus, pressedKey is cleared.

Fixes: https://todo.sr.ht/~eliasnaur/gio#525

### DIFF
--- a/widget/button_test.go
+++ b/widget/button_test.go
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: Unlicense OR MIT
+
+package widget_test
+
+import (
+	"image"
+	"testing"
+
+	"gioui.org/io/key"
+	"gioui.org/io/router"
+	"gioui.org/io/system"
+	"gioui.org/layout"
+	"gioui.org/op"
+	"gioui.org/widget"
+)
+
+func TestClickable(t *testing.T) {
+	var (
+		ops op.Ops
+		r   router.Router
+		b1  widget.Clickable
+		b2  widget.Clickable
+	)
+	gtx := layout.NewContext(&ops, system.FrameEvent{Queue: &r})
+	layout := func() {
+		b1.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
+			return layout.Dimensions{Size: image.Pt(100, 100)}
+		})
+		// buttons are on top of each other but we only use focus and keyevents, so this is fine
+		b2.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
+			return layout.Dimensions{Size: image.Pt(100, 100)}
+		})
+	}
+	frame := func() {
+		ops.Reset()
+		layout()
+		r.Frame(gtx.Ops)
+	}
+	// frame: request focus for button 1
+	b1.Focus()
+	frame()
+	// frame: gain focus for button 1
+	frame()
+	if !b1.Focused() {
+		t.Error("button 1 did not gain focus")
+	}
+	if b2.Focused() {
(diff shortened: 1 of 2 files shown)


eliasnaur:
> This commit fixes the non-intuitive behaviour, where hitting return or space with a button focused, then tabbing to another button and releasing the key causes the second button to trigger. It feels wrong, as the "gesture" was never initiated on the second button. The fix makes widget.Clickable track which key was pressed, in a variable called pressedKey, and only considers a key release if the released key matches the pressed key. Finally, if the widget loses focus, pressedKey is cleared.

Can you add a test, say in `widget/button_test.go`, that demonstrates the issue and verifies the fix? There is an existing test for `Bool` in `widget/widget_test.go`.

> 
> Fixes: https://todo.sr.ht/~eliasnaur/gio#525

I believe this should be with a slash:

Fixes: https://todo.sr.ht/~eliasnaur/gio/525

vsariola:
> Can you add a test, say in `widget/button_test.go`, that demonstrates the issue and verifies the fix? There is an existing test for `Bool` in `widget/widget_test.go`.

Done, to the best of my ability, but I have a feeling there's cleaner way to write the test.

> > Fixes: https://todo.sr.ht/~eliasnaur/gio#525
> 
> I believe this should be with a slash:
> 
> Fixes: https://todo.sr.ht/~eliasnaur/gio/525

Changed to a slash. I got the # syntax from here https://gioui.org/doc/contribute so assuming slash is correct, there's likely a typo in the instructions.

widget/button_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"gioui.org/io/key"
+	"gioui.org/io/router"
+	"gioui.org/io/system"
+	"gioui.org/layout"
+	"gioui.org/op"
+	"gioui.org/widget"
+)
+
+func TestButtonKeypresses(t *testing.T) {
```
eliasnaur, line 17: Nit: rename to TestClickable.

widget/button_test.go
```diff
@@ -0,0 +1,109 @@
...
+					return layout.Dimensions{Size: image.Pt(100, 100)}
+				})
+			}),
+			layout.Rigid(func(gtx layout.Context) layout.Dimensions {
+				return b2.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
+					return layout.Dimensions{Size: image.Pt(100, 100)}
+				})
+			}),
+		)
+	}
```
eliasnaur, line 38: If you only need to offset `b2`, `layout` can be written without Flex, by using `op.Offset` just before laying out `b2`:

```go
defer op.Offset(image.Pt(100, 100)).Push(gtx.Ops).Pop()
```

eliasnaur (commented):
LGTM, apart from nits. @whereswaldon do you have any objections to this behaviour change?

whereswaldon:
> LGTM, apart from nits. @whereswaldon do you have any objections to this behaviour change?

No, I wholeheartedly agree with it.

vsariola:
Done. I didn't even offset the buttons; just laid them out on top of each other as the test doesn't click anything, just uses Focus and keyboard presses, so it doesn't make a difference. Didn't want to add anything unnecessary to the test.

PR #122, opened 2023-09-04
[widget] don't open soft keyboard if editor is readonly

g45t345rt:
For mobile, you don't need the keyboard for editing if it's readonly.


widget/editor.go
```diff
@@ -643,7 +643,9 @@ func (e *Editor) layout(gtx layout.Context, textMaterial, selectMaterial op.Call
 	key.InputOp{Tag: &e.eventKey, Hint: e.InputHint, Keys: keys}.Add(gtx.Ops)
 	if e.requestFocus {
```
eliasnaur, line 644: Wouldn't it be better if `e.requestFocus` is not set when `e.ReadOnly`? Then read-only Editors will never reach this branch, and not ask for focus with the `key.FocusOp` below.

eliasnaur (changes requested):
Thank you. One comment about `Editor.requestFocus`. Please sign-off and prefix your commit messages according to the [constribution guidelines](https://gioui.org/doc/contribute). For example:

`widget: don't let readonly Editors ask for focus and keyboard`

eliasnaur:
I believe this PR is obsoleted by https://todo.sr.ht/~eliasnaur/gio/550.

g45t345rt:
Dope. I'll take a look at this branch, and I really the approach you're taking with the new releases.

PR #124, opened 2023-09-22
io/transfer,app: [macOS] add support for file opening events

lzambarda:
Correctly handle macOS file opening events so that an app containing a binary using gio can be used to open certain files and even be set as default app for certain file extensions.
This is done via implementing application:openFile in GioAppDelegate.

File events issued this way will be made available to a gio through the io/transfer package. This package now also contains documentation for how to register a macOS app as able to support certain file extensions.

**Note:** I was working on this and sending patches to the main developer via email. However I struggle to find time to finalise some changes and therefore I will leave this so that it can be discussed and maybe picked-up by some other developer as I believe the changes in here are quite valuable for macOS developers!

### DIFF
--- a/app/os_macos.go
+++ b/app/os_macos.go
@@ -8,6 +8,10 @@ package app
 import (
 	"errors"
 	"image"
+	"io"
+	"mime"
+	"os"
+	"path"
 	"runtime"
 	"time"
 	"unicode"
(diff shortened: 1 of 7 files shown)


eliasnaur:
Duplicate of or overlaps https://github.com/gioui/gio/pull/117?

PR #125, opened 2023-10-02
widget/material: allow changing height & radius of progressbar

apprehensions:


widget/material/progressbar.go
```diff
@@ -49,6 +49,10 @@ func (p ProgressBarStyle) Layout(gtx layout.Context) layout.Dimensions {
 			return shader(progressBarWidth, p.TrackColor)
 		}),
 		layout.Stacked(func(gtx layout.Context) layout.Dimensions {
+			if p.Progress == 0.0 {
+				return layout.Dimensions{}
+			}
```
eliasnaur, line 54: Is this a bug fix or optimization? I prefer `shader` being tolerant of a zero width.

eliasnaur (commented):
LGTM except for a single nit.

Please sign off your commits as described in https://gioui.org/doc/contribute.

widget/material/progressbar.go
apprehensions: Simply look fix, when a progress bar is rendered and it's 0, there will be tiny dots that don't look very nice.

apprehensions: > I prefer shader being tolerant of a zero width

How do I do this?

apprehensions: I would also like to mention this is a workaround around the radius, this problem dissapears if the radius is set to 0. If its not then it will still render the 'rounded' corners despite there being no progress. Another issue is if the radius is bigger than the height it will cause weird rendering issues.

eliasnaur: I see. Doesn't the progress indicator look odd as long as it's narrower than the rounded corner radius? I'd expect something like (untested)

```go
fillWidth := int(float32(progressBarWidth) * clamp1(p.Progress))
if minWidth :=  rr*2; fillWidth < minWidth {
   fillWidth = minWidth
}
```

to ensure that there is always a minimal (circle) progress indicator. Alternatively, you can keep the early return to not draw an indicator if progress is too little.

apprehensions: > to ensure that there is always a minimal (circle) progress indicator.

Is there any framework that does this? i've never seen this before and i've mostly only seen just empty progressbars.

eliasnaur: I'll leave it to you whether to return. My point was that 0.0 is probably not the right cut-off.

apprehensions: Sorry, I still don't quite understand. If there is no progress, don't display it. The current behavior as i said earlier is a bug within the radius rendering where the progress isn't drawn, correct, but the rounded corners themselves are still drawn, if this gets solved than the return wont be necessary.

eliasnaur: What happens if, say, `p.Progress == math.SmallestNonzeroFloat32`? Does the progress bar still render nicely? If so, your fix is correct and I retract my objection. If not, I argue that your test should be changed to `p.Progress < min` where `min` is greater than 0.0.

apprehensions: I hope you do realize this is a workaround to the radius problem, anything other than 0.0 causes the progress to be rendered and the rounded corners to appear again, despite there being no progress.

eliasnaur: Perhaps we're talking across each other? I agree that the should be a cut-off, but I think the 0.0 value is too low.

To illustrate, I patched gioui.org/example/kitchen with

```patch
diff --git a/kitchen/kitchen.go b/kitchen/kitchen.go
index d5254cf..bfb75fb 100644
--- a/kitchen/kitchen.go
+++ b/kitchen/kitchen.go
@@ -240,6 +238,8 @@ func (b iconAndTextButton) Layout(gtx layout.Context) layout.Dimensions {
 }

 func kitchen(gtx layout.Context, th *material.Theme) layout.Dimensions {
+       const uglyProgress = math.SmallestNonzeroFloat32 * 1e42
+       fmt.Println(uglyProgress)
        for _, e := range lineEditor.Events() {
                if e, ok := e.(widget.SubmitEvent); ok {
                        topLabel = e.Text
@@ -327,7 +327,7 @@ func kitchen(gtx layout.Context, th *material.Theme) layout.Dimensions {
                                layout.Rigid(material.ProgressCircle(th, progress).Layout),
                        )
                },
-               material.ProgressBar(th, progress).Layout,
+               material.ProgressBar(th, uglyProgress).Layout,
                func(gtx C) D {
                        return layout.Flex{Alignment: layout.Middle}.Layout(gtx,
                                layout.Rigid(
```

which printed `0.0014012984643248171` and showed the ugly progress indicator.

apprehensions: Reminder: this is what the progress looks like if 0.0.

apprehensions: I've been trying to figure out a way to cut off the height and make the radius smaller depending on the progress's radius and if it is small enough to not make a full circle on the radius, but cannot figure it out...

eliasnaur:
LGTM, but leaving for @whereswaldon to determine whether the Height/Width fields fit the style of widgets.

whereswaldon (approved):
I think these fields do belong here. This API change strictly provides more control to consumers of package material.

eliasnaur:
Merged, thank you for your patience.

PR #126, opened 2023-10-10
text: fix run offset

morlay:
xOffset will change when letter after different letter
example: "Tooltip" 
* the first  o with -1 xOffset
* the second o with  0 xOffset

the glyph boxes are overlapping, when the negative xOffset exists. like, The `T` box and `o` box.
image

Without this patch, the double `o` is so closed, without overlapping
image

### DIFF
--- a/text/shaper.go
+++ b/text/shaper.go
@@ -458,7 +458,13 @@ func (l *Shaper) NextGlyph() (_ Glyph, ok bool) {
 			l.advance += g.xAdvance
 		}
 		// runOffset computes how far into the run the dot should be positioned.
-		runOffset := l.advance
+		//
+		// xOffset will change when letter after different letter
+		// example: "Tool"
+		//  	the first  o with -1 xOffset
+		//  	the second o with  0 xOffset
+		// and the glyph boxes are overlapping, because of the xOffset
+		runOffset := l.advance + g.xOffset
 		if rtl {
 			runOffset = run.Advance - l.advance
 		}


eliasnaur:
Please sign off your change according to [the contribution guide](https://gioui.org/doc/contribute). Thank you.

whereswaldon (changes requested):
Thank you for the PR! However, after some thought, I don't think this is the proper fix. The xOffset field of a glyph is a rendering concern. It shifts the origin of the coordinate space of the glyph's path data relative to the "dot". It does not influence the logical position of the advance for that glyph.

In this diagram, the `xOffset` field is called `xBearing`, but it's the same thing.

glyph-metrics-3 [source](https://freetype.org/freetype2/docs/glyphs/glyph-metrics-3.svg)

I think the problem is elsewhere in your example. Which font did you use in testing? If I could reproduce this issue, I could comment more intelligently. I tried to identify the font in your screenshot using some online tools, but I was unable to reproduce the problem with "Nimbus Sans", which seemed like the best match. If you don't know the font, you should be able to get some insight with `GIODEBUG=text` when running your application.

morlay:
@whereswaldon 


Roboto and Noto, which used by m3

https://m3.material.io/styles/typography/overview

https://github.com/octohelm/gio-compose/blob/main/pkg/component/m3/theming/typography.go#L191

whereswaldon:
I've tried to replicate your problem with this program:

```go
// SPDX-License-Identifier: Unlicense OR MIT

package main

// A simple Gio program. See https://gioui.org for more information.

import (
	"log"
	"os"

	"gioui.org/app"
	"gioui.org/io/system"
	"gioui.org/layout"
	"gioui.org/op"
	"gioui.org/text"
	"gioui.org/widget/material"
)

func main() {
	go func() {
		w := app.NewWindow()
		if err := loop(w); err != nil {
			log.Fatal(err)
		}
		os.Exit(0)
	}()
	app.Main()
}

func loop(w *app.Window) error {
	th := material.NewTheme()
	var ops op.Ops
	for {
		e := <-w.Events()
		switch e := e.(type) {
		case system.DestroyEvent:
			return e.Err
		case system.FrameEvent:
			l := material.Body1(th, "Tooltip")
			l.Font.Typeface = "Roboto"
			l.Alignment = text.Middle
			l.Layout(gtx)
			e.Frame(gtx.Ops)
		}
	}
}
```

However, both Roboto and Noto Sans display the two "o"s as non-overlapping. Can you provide a minimal program like the above that forces it to occur? Perhaps it only happens at a particular text size or something?

morlay:
@whereswaldon 

I use your code. and got the overlap too.
image

With the patched code

```go.mod
replace gioui.org => github.com/morlay/gio v0.0.0-20231011025659-28acde2d58e3
```

It looks better
image


May os issue?

I run this on macOS 14.0 

```shell
$ uname -a

Darwin MacBook-Pro-2.local 23.0.0 Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:43 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T6000 arm64 arm Darwin
```

whereswaldon:
This suggests to me that we are using different versions of the affected fonts. Can you locate those fonts on disk and attach them here? I'll share mine when I'm back at a computer. It's possible that this is a bug in an old font version or even a corrupt font file. Being able to diff the fonts in use would be helpful.

morlay:
@whereswaldon 

I made a mistake.  I haven't install fonts both
https://fonts.google.com/noto/specimen/Noto+Sans
https://fonts.google.com/specimen/Roboto.

I think the issued font may some `sans-serif`  I don't know.

whereswaldon:
I've written [a patch](https://lists.sr.ht/~eliasnaur/gio-patches/patches/46590) to make it easier to check this in the future. With that patch applied, running your app with `GIODEBUG=text` will log each time a system font is loaded. That should help you to determine when/if you're using the fonts you expect.

In general though, this seems to be an issue with a bad system font rather than a Gio text stack problem. I think I'll close this PR for the time being.

@morlay Thank you very much for your patience working through this puzzling problem with me. Your dedication to improving our text stack is appreciated.

PR #127, opened 2023-10-13
widget: Update Selectable key filter

theclapp:
Selectable was using a key event filter copied directly from editor.go, but it didn't actually process all those keys. Update the filter to only ask for the keys that Selectable actually uses.


whereswaldon (approved):
Thank you!

whereswaldon:
Merged.

```
To git.sr.ht:~eliasnaur/gio
   63fea3d2..ae2b1f42  main -> main
```

PR #128, opened 2023-11-20
app: raiseWindow on macos should activate the app if not active

dirslashls:
Calling window.Perform(system.ActionRaise) does not show the window on the top if the app is currently not active. This can happen for example, if the app integrated with systray (https://pkg.go.dev/fyne.io/systray) where the menu item launches a window, the window is not showing at the top. It is fixed by activating the current app if necessary

### DIFF
--- a/app/os_macos.go
+++ b/app/os_macos.go
@@ -192,6 +192,10 @@ static CFTypeRef windowForView(CFTypeRef viewRef) {
 }
 
 static void raiseWindow(CFTypeRef windowRef) {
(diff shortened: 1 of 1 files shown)


whereswaldon (commented):
Thank you for reading the contribution docs and both signing-off your commit and formatting the commit message appropriately. Structurally, this PR looks great! I'll wait for @eliasnaur's thoughts on the content.

eliasnaur:
Merged, thanks!

PR #131, opened 2024-02-04
material: fix documentation of using buttons

jes:
The material `Button` and `IconButton` take the `*widget.Clickable` as an argument to the constructor, instead of an argument to `Layout()`, so I've updated the documentation accordingly.

Also `IconButton()` takes a description string, so I added one, I chose `"Click me!"` for consistency with the other example, but maybe something else is more appropriate.


eliasnaur:
Thank you! Merged.

PR #132, opened 2024-02-28
Enhance focus handling in X11 by updating stage status

olablt:
app: improve focus event handling for X11 environment

This commit enhances the application's response to focus gain and loss events under the X11 environment. Specifically, when the application window gains focus, it now triggers a StageEvent with the 'StageRunning' state. Conversely, when the window loses focus, it transitions to the 'StageInactive' state. This change ensures more intuitive application behavior in response to user focus changes.

Discussion on Slack regarding future enhancements: https://gophers.slack.com/archives/CM87SNCGM/p1709048436843239

### DIFF
--- a/app/os_ios.go
+++ b/app/os_ios.go
@@ -21,6 +21,7 @@ struct drawParams {
 };
 
 static void writeClipboard(unichar *chars, NSUInteger length) {
+#if !TARGET_OS_TV
 	@autoreleasepool {
 		NSString *s = [NSString string];
 		if (length > 0) {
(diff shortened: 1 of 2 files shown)


whereswaldon (commented):
Thank you! Change looks reasonable to me, but the commit needs to be signed off as described on this page: https://gioui.org/doc/contribute#style

Also, please update the commit message to the style described on that page.

olablt:
I have signed off on the commit and updated the commit message according to the guidelines provided.

Regarding the implementation details, could you provide some advice on how to trigger a `pointer.Leave` event when the pointer is at the edge and the window loses focus? Any insights or references would be greatly appreciated.

eliasnaur:
The merging of https://todo.sr.ht/~eliasnaur/gio/555 replaces `StageEvent` with `Config.Focused`. Sorry for the conflict.

> Regarding the implementation details, could you provide some advice on how to trigger a `pointer.Leave` event when the pointer is at the edge and the window loses focus? Any insights or references would be greatly appreciated.

I believe you'll have to change `app.Window` to tell its `io/input.Router` that focus is lost, thereby triggering a (synthetic) `pointer.Leave`.  There is already code to translate a window focus to a widget focus: https://git.sr.ht/~eliasnaur/gio/tree/main/item/app/window.go#L634.

PR #135, opened 2024-05-30
io: make Event/Execute not inlined

inkeliz:
That change makes possible to overwrite both functions, by changing the function body during the linking phase, namely using //go:linkname.

That is a alternative to #134.


inkeliz:
I'm dropping it, after further testing, it causes some issues with Go compiler. :|

PR #136, opened 2024-05-31
io: add option to overwrite Event/Execute

inkeliz:
This change makes possible to change the default
behaviour, and also expose Router.Execute.

### DIFF
--- a/io/input/router.go
+++ b/io/input/router.go
@@ -170,14 +170,6 @@ func (q *Router) Source() Source {
 	return Source{r: q}
 }
 
-// Execute a command.
-func (s Source) Execute(c Command) {
-	if !s.Enabled() {
-		return
-	}
-	s.r.execute(c)
-}
-
 // Enabled reports whether the source is enabled. Only enabled
 // Sources deliver events and respond to commands.
 func (s Source) Enabled() bool {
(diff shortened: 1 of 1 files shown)


ddkwork:
Is it possible to skip interfaces and global variables in favor of generics

inkeliz:
Sorry, but can you clarify how would Generics be implemented here?


The usage of generics will also requires a specific implementation, for instance in `app.FrameEvent`, `layout.Context`, which uses `Source`, or `input.Source` itself. If it's `layout.Context[T]` (then `layout.Context[Source]`) the issue still the same, since it's not possible to use other type. If `input.Source` becomes `input.Source[T any]` then we have the exactly same issue.

I don't see how Generics can be implemented here, but might missing something (I just awakup few minutes ago )

ddkwork:
I'm currently busy refactoring another library, and I'll take the time to read the source code and analyze the feasibility of generalization soon!



---Original---
From: ***@***.***&gt;
Date: Sat, Jun 1, 2024 15:45 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [gioui/gio] io: add option to overwrite Event/Execute (PR #136)




 
Sorry, but can you clarify how would Generics be implemented here?
 
The usage of generics will also requires a specific implementation, for instance in app.FrameEvent, layout.Context, which uses Source, or input.Source itself. If it's layout.Context[T] (then layout.Context[Source]) the issue still the same, since it's not possible to use other type. If input.Source becomes input.Source[T any] then we have the exactly same issue.
 
I don't see how Generics can be implemented here, but might missing something (I just awakup few minutes ago )

io/input/router.go
```diff
@@ -193,12 +185,34 @@ func (s Source) Focused(tag event.Tag) bool {
...
+// implement a custom event processor or testing.
+var SourceEventProcessor = (*Router).Event
+
 // Event returns the next event that matches at least one of filters.
 func (s Source) Event(filters ...event.Filter) (event.Event, bool) {
 	if !s.Enabled() {
 		return nil, false
 	}
-	return s.r.Event(filters...)
+	return SourceEventProcessor(s.r, filters...)
```
eliasnaur, line 215: I believe this indirection will cause `filters` to escape the heap, causing an allocation for every call to `Source.Event`.

ddkwork: > I believe this indirection will cause `filters` to escape the heap, causing an allocation for every call to `Source.Event`.

So a callback function method should be added?

ddkwork: I also prepared a file drag and drop implementation for gio for the Windows platform, but the issue of passing events is a bit confusing, as he said, the current gui can only receive events can not generate custom events which brings confusion to the implementation of the file drag and drop events

PR #137, opened 2024-06-02
pointer: fix documentation

inkeliz:
Previously it uses event.Op{}, but such struct don't exists anymore. Instead, it have a function with the same name.


eliasnaur:
Thanks, merged.

PR #138, opened 2024-06-08
io,widget: [android] improve focus

inkeliz:
This patch is intended to fix two issues related to focus and software-keyboard.

The focus is now reseted if it's lost due to external event, such as clicking on non-Gio window/activity/fragment.

The software-keyboard will now re-open when clicked, except if it's ReadOnly.

Fixes: https://todo.sr.ht/~eliasnaur/gio/591


app/os_macos.m
```diff
@@ -47,13 +47,17 @@ - (void)windowDidChangeScreen:(NSNotification *)notification {
...
+	}
 }
 - (void)windowDidResignKey:(NSNotification *)notification {
 	NSWindow *window = (NSWindow *)[notification object];
-  GioView *view = (GioView *)window.contentView;
-	gio_onFocus(view.handle, 0);
+	GioView *view = (GioView *)window.contentView;
+	if ([window firstResponder] == view) {
+		gio_onFocus(view.handle, 0);
+	}
```
inkeliz, line 60: This is relative to the window, not the view. If you have multiple views in the same window, it used to incorrectly report that Gio had the focus.

widget/editor.go
```diff
@@ -289,6 +289,9 @@ func (e *Editor) processPointerEvent(gtx layout.Context, ev event.Event) (Editor
 				Y: int(math.Round(float64(evt.Position.Y))),
 			})
 			gtx.Execute(key.FocusCmd{Tag: e})
+			if !e.ReadOnly {
+				gtx.Execute(key.SoftKeyboardCmd{Show: true})
+			}
```
inkeliz, line 294: Doesn't make sense to request SoftKeyboard when it's ReadOnly, maybe separated commit?

app/os_macos.go
```diff
@@ -514,7 +519,11 @@ func (w *window) EditorStateChanged(old, new editorState) {
 	}
 }
 
-func (w *window) ShowTextInput(show bool) {}
+func (w *window) ShowTextInput(show bool) {
+	if show && !w.config.Focused {
+		C.setFocus(w.view)
+	}
+}
```
inkeliz, line 526: This is macOS, which doesn't have SoftwareKeyboard. However, I didn't find any specific function to request focus, and adding a new function will require changes on multiple OSes, even as `no-op`.

io/input/router.go
```diff
@@ -567,6 +567,14 @@ func (q *Router) changeState(e event.Event, state inputState, evts []taggedEvent
 			e.event = de
 		}
 	}
+	for i := range evts {
+		e := &evts[i]
+		if fe, ok := e.event.(key.FocusEvent); ok {
+			if !fe.Focus {
+				state.keyState.focus = nil
```
eliasnaur, line 574: It seems to me this should be done in `Router.processEvent`, near the lines

```go
	case key.EditEvent, key.FocusEvent, key.SelectionEvent:
		var evts []taggedEvent
		if f := state.focus; f != nil {
			evts = append(evts, taggedEvent{tag: f, event: e})
		}
		q.changeState(e, state, evts)
```

But I'm not perfectly sure what use-case you're fixing. Can you add a test to io/input/key_test.go for my understanding and to guard against regressions?

app/os_macos.go
eliasnaur: I'm not sure why you need this here and not on other desktop platforms. Why can't `C.setFocus` be called automatically in the cases where focus was lost?

eliasnaur (commented):
I extracted the non-controversial changes, leaving only the two that confuse me. Please rebase.

io/input/router.go
inkeliz: What I'm fixing is reseting the `focus` when receiving `key.FocusEvent` from "the OS".

Consider that you have one `FocusCmd{Tag: TextArea}`. If Gio gets one `FocusEvent` from "the OS" (say: macos), it will not remove the focus from `TextArea`. That is the issue here. 

So, this change changes the `state` (state.keyState.focus`) to `nil` when the FocusEvent is `false`. Since it's messing with `state` I use the `changeState` function.

app/os_macos.go
inkeliz: It's possible to call it when the GioView is clicked, and doesn't have focus. I don't think Gio should immediate request for Focus, because it's possible to have multiple non-Gio View in the same Window.

PR #140, opened 2024-08-13
text,widget/material: Update doc for Shaper & Theme

theclapp:
Note that you should use different Themes, with different Shapers, for different top-level windows, and explain why.


theclapp:
Fixes https://todo.sr.ht/~eliasnaur/gio/607.

PR #142, opened 2024-08-24
app: [Windows] Fix abnormal sudden change in window size

sxwlmsrbi:
In previous versions, as long as `app.MinSize()` is applied (`app.Decorated` is `true`), there is a probability of a sudden change in window size after the window has been successively minimized and restored. I let the `bw` and `bh` variables be calculated only once to solve the problem.

Fixes: https://todo.sr.ht/~eliasnaur/gio/608


eliasnaur:
Thank you. I reworked your fix into https://gioui.org/commit/3230dd7.

PR #143, opened 2024-08-29
app: add support to hidden minimize and maximize buttons on macos and …

kkeybbs:
Two options `app.HiddenMinimizeButton` and `app.HiddenMaximizeButton` are added to create a window without minimize or/and maximize buttons.

## hide both minimize and maximize buttons
```
w := new(app.Window)
w.Option(app.HiddenMinimizeButton(true), app.HiddenMaximizeButton(true))
```
image

## hide minimize button, keep maximize button:
```
w := new(app.Window)
w.Option(app.HiddenMinimizeButton(true), app.HiddenMaximizeButton(false))
```
- on macOS:
image

- on Windows:
image

### DIFF
--- a/app/os.go
+++ b/app/os.go
@@ -45,6 +45,12 @@ type Config struct {
 	CustomRenderer bool
 	// Decorated reports whether window decorations are provided automatically.
 	Decorated bool
+
+	// MinimizeButtonHidden hide the window's minimize button (only works for windows and macOS)
+	MinimizeButtonHidden bool
+	// MaximizeButtonHidden hide the window's maximize button (only works for windows and macOS)
+	MaximizeButtonHidden bool
+
 	// Focused reports whether has the keyboard focus.
 	Focused bool
 	// decoHeight is the height of the fallback decoration for platforms such
--- a/app/window.go
+++ b/app/window.go
@@ -969,6 +969,24 @@ func Decorated(enabled bool) Option {
 	}
 }
 
+// MinimizeButtonHidden hides the minimize button of the window.
+//
+// MinimizeButtonHidden is supported on Windows and macOS.
+func MinimizeButtonHidden(hidden bool) Option {
+	return func(_ unit.Metric, cnf *Config) {
+		cnf.MinimizeButtonHidden = hidden
+	}
+}
+
+// MaximizeButtonHidden hides the maximize button of the window.
+//
+// MaximizeButtonHidden is supported on Windows and macOS.
+func MaximizeButtonHidden(hidden bool) Option {
+	return func(_ unit.Metric, cnf *Config) {
+		cnf.MaximizeButtonHidden = hidden
+	}
+}
+
 // flushEvent is sent to detect when the user program
 // has completed processing of all prior events. Its an
 // [io/event.Event] but only for internal use.
--- a/app/os_macos.go
+++ b/app/os_macos.go
@@ -495,8 +495,11 @@ func (w *window) Configure(options []Option) {
 	C.setWindowTitleVisibility(window, titleVis)
 	C.setWindowStyleMask(window, mask)
 	C.setWindowStandardButtonHidden(window, C.NSWindowCloseButton, barTrans)
-	C.setWindowStandardButtonHidden(window, C.NSWindowMiniaturizeButton, barTrans)
-	C.setWindowStandardButtonHidden(window, C.NSWindowZoomButton, barTrans)
+	// no decorated or hide minimize and maximize buttons
+	w.config.MinimizeButtonHidden = cnf.MinimizeButtonHidden
+	w.config.MaximizeButtonHidden = cnf.MaximizeButtonHidden
+	C.setWindowStandardButtonHidden(window, C.NSWindowMiniaturizeButton, toInt(!cnf.Decorated || cnf.MinimizeButtonHidden))
+	C.setWindowStandardButtonHidden(window, C.NSWindowZoomButton, toInt(!cnf.Decorated || cnf.MaximizeButtonHidden))
 	// When toggling the titlebar, the layer doesn't update its frame
 	// until the next resize. Force it.
 	C.resetLayerFrame(w.view)
@@ -1159,6 +1162,14 @@ func convertMods(mods C.NSUInteger) key.Modifiers {
 		kmods |= key.ModShift
 	}
 	return kmods
+}
+
+// toInt golang bool to C.int
+func toInt(v bool) C.int {
+	if v {
+		return C.int(C.YES)
+	}
+	return C.int(C.NO)
 }
 
 func (AppKitViewEvent) implementsViewEvent() {}
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -710,6 +710,8 @@ func (w *window) Configure(options []Option) {
 	w.config.Decorated = cnf.Decorated
 	w.config.MinSize = cnf.MinSize
 	w.config.MaxSize = cnf.MaxSize
+	w.config.MinimizeButtonHidden = cnf.MinimizeButtonHidden
+	w.config.MaximizeButtonHidden = cnf.MaximizeButtonHidden
 	windows.SetWindowText(w.hwnd, cnf.Title)
 
 	style := windows.GetWindowLong(w.hwnd, windows.GWL_STYLE)
@@ -771,7 +773,9 @@ func (w *window) Configure(options []Option) {
 
 	// Disable maximize button if MaxSize is set.
 	if cnf.MaxSize != (image.Point{X: 0, Y: 0}) {
-		style &^= windows.WS_MAXIMIZEBOX
+		// On Windows, clicking maximize button will maximize the window to its' max size, so we don't hide the button.
+		// style &^= windows.WS_MAXIMIZEBOX
+
 		// Disable window resizing if MinSize and MaxSize are equal.
 		if cnf.MinSize == cnf.MaxSize {
 			style &^= windows.WS_THICKFRAME
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -772,14 +772,8 @@ func (w *window) Configure(options []Option) {
 	}
 
 	// Disable maximize button if MaxSize is set.
-	if cnf.MaxSize != (image.Point{X: 0, Y: 0}) {
-		// On Windows, clicking maximize button will maximize the window to its' max size, so we don't hide the button.
-		// style &^= windows.WS_MAXIMIZEBOX
-
-		// Disable window resizing if MinSize and MaxSize are equal.
-		if cnf.MinSize == cnf.MaxSize {
-			style &^= windows.WS_THICKFRAME
-		}
+	if cnf.MaxSize != (image.Point{X: 0, Y: 0}) && cnf.MinSize == cnf.MaxSize {
+		style &^= windows.WS_THICKFRAME
 	}
 
 	// Note: these invocation all trigger the windows callback method which may process a pending system.ActionCenter
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -758,6 +758,16 @@ func (w *window) Configure(options []Option) {
 		swpStyle |= windows.SWP_NOMOVE | windows.SWP_NOSIZE
 		showMode = windows.SW_SHOWMAXIMIZED
 	}
+	if w.config.MinimizeButtonHidden {
+		style &^= windows.WS_MINIMIZEBOX
+	} else {
+		style |= windows.WS_MINIMIZEBOX
+	}
+	if w.config.MaximizeButtonHidden {
+		style &^= windows.WS_MAXIMIZEBOX
+	} else {
+		style |= windows.WS_MAXIMIZEBOX
+	}
 
 	// Disable maximize button if MaxSize is set.
 	if cnf.MaxSize != (image.Point{X: 0, Y: 0}) {


app/os.go
```diff
@@ -45,6 +45,12 @@ type Config struct {
 	CustomRenderer bool
 	// Decorated reports whether window decorations are provided automatically.
 	Decorated bool
+
+	// HiddenMinimizeButton hide the window's minimize button
+	HiddenMinimizeButton bool
+	// HiddenMaximizeButton hide the window's maximize button
+	HiddenMaximizeButton bool
```
theclapp, line 52: I think you should either note that these don't work under Linux, or make them work under Linux.

kkeybbs: I have tried for days to test on ubuntu with os_x11.go, but still not working.
- _NET_WM_ALLOWED_ACTIONS
- _MOTIF_WM_HINTS
- _NET_WM_WINDOW_TYPE: _NET_WM_WINDOW_TYPE_DIALOG

They all didn't work. I belive these properties also depend on window manager implementation, according to infomations bellow:
- https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html
- https://stackoverflow.com/questions/14442081/disable-actions-move-resize-minimize-etc-using-python-xlib
- https://superuser.com/questions/186748/how-to-hide-or-minimize-x11-window-from-console

I'll continue wokring to make them work on x11 and wayland, but this may take long time.
So I add tips "only works for windows and macOS" for the options.
Maybe the feature working only on some os is acceptable?

theclapp: > Maybe the feature working only on some os is acceptable?

Fine with me. Better than nothing! Thanks for trying.

eliasnaur:
Can you describe the use-cases you need these controls for? Can we just automatically apply them when, say, MinSize == MaxSize?

kkeybbs:
> Can you describe the use-cases you need these controls for? Can we just automatically apply them when, say, MinSize == MaxSize?

Some usefull cases in my business products:
- using both HiddenMaximizeButton and HiddenMinimizeButton
  - Making a dialog window, such as application's "About" dialog. Just need closing insteadof minimizing and maximizing.
  - Instead of minimizing to Dock, user can click the app, and makes it work in the background and an icon in the systray. Users like this easy experience, it  saves spaces in Dock.
- using HiddenMaximizeButton, to disallow a window to be a fullscreen window in new desktop on macOS.

> So maybe I can try how to make a window be dialog in x11.

Sometimes HiddenMaximizeButton can be automatically apply with "MinSize == MaxSize", but not always when size is screen size. Maximizing will make the window as a new fullscreen window in macOS. So leave them standalone will be better.

eliasnaur:
Sorry for the slow response; I'm unusually busy these days.

> > Can you describe the use-cases you need these controls for? Can we just automatically apply them when, say, MinSize == MaxSize?
> 
> Some usefull cases in my business products:
> 
>     * using both HiddenMaximizeButton and HiddenMinimizeButton
>       
>       * Making a dialog window, such as application's "About" dialog. Just need closing insteadof minimizing and maximizing.
>       * Instead of minimizing to Dock, user can click the app, and makes it work in the background and an icon in the systray. Users like this easy experience, it  saves spaces in Dock.
> 
>     * using HiddenMaximizeButton, to disallow a window to be a fullscreen window in new desktop on macOS.
> 
> 

Upon further though, perhaps the rules could be:

 - If MinSize != {0, 0}, hide/disable minimize button
 - If MaxSize != {0, 0}, hide/disable maximize button.

What do you think?

> > So maybe I can try how to make a window be dialog in x11.
> 
> Sometimes HiddenMaximizeButton can be automatically apply with "MinSize == MaxSize", but not always when size is screen size. Maximizing will make the window as a new fullscreen window in macOS. So leave them standalone will be better.

Can you elaborate? What is special about the screen size?

kkeybbs:
I'm not sure if the rules from app.Option like this `app.MinSize(0, 0)`?
So `MinSize != {0, 0}`  and `MaxSize != {0, 0}` will always be true.

```
// MinSize sets the minimum size of the window.
func MinSize(w, h unit.Dp) Option {
	if w <= 0 {
		panic("width must be larger than or equal to 0")
	}
	if h <= 0 {
		panic("height must be larger than or equal to 0")
	}
	return func(m unit.Metric, cnf *Config) {
		cnf.MinSize = image.Point{
			X: m.Dp(w),
			Y: m.Dp(h),
		}
	}
}
```

eliasnaur:
If you're saying there's no way to revert to no MinSize once a MinSize has been set, then yes. It seems to me Gio will need to allow `app.MinSize(0, 0)` to mean "remove min size cosntraint". Same with `app.MaxSize`.

Would you like to contribute that change as well?

kkeybbs:
You are right as the initial size may be `{0, 0}`, or Gio change this cosntraint.

But when using `MinSize != {0, 0}`, most users just want to make the window not too small, and still keep the minimize button. Only few users want to hide/disable it.

And changing the rules change the default window-style which the most users expected.
Explicit options to hide minimize and maximize buttons are good for the most.

app/os.go
```diff
@@ -176,6 +182,8 @@ type context interface {
 // driver is the interface for the platform implementation
 // of a window.
 type driver interface {
+	// DriverName get driver anme
+	DriverName() string
```
eliasnaur, line 186: I believe `DriverName` and its implementation are irrelevant to this change.

app/os.go
```diff
@@ -45,6 +45,12 @@ type Config struct {
 	CustomRenderer bool
 	// Decorated reports whether window decorations are provided automatically.
 	Decorated bool
+
+	// HiddenMinimizeButton hide the window's minimize button (only works for windows and macOS)
+	HiddenMinimizeButton bool
+	// HiddenMaximizeButton hide the window's maximize button (only works for windows and macOS)
+	HiddenMaximizeButton bool
```
eliasnaur, line 52: Do we still need this, and the `Option` counterpart? As per our discussion, I think the maximize button should be hidden automatically when `MaxSize` is set.

app/os.go
```diff
@@ -45,6 +45,12 @@ type Config struct {
 	CustomRenderer bool
 	// Decorated reports whether window decorations are provided automatically.
 	Decorated bool
+
+	// HiddenMinimizeButton hide the window's minimize button (only works for windows and macOS)
+	HiddenMinimizeButton bool
```
eliasnaur, line 50: Nit: name this `MinimizeButtonHidden` and move the comment about supported platforms to its `Option` counterpart.

app/window.go
```diff
@@ -970,6 +978,20 @@ func Decorated(enabled bool) Option {
 	}
 }
 
+// HiddenMinimizeButton controls whether show minimize button on macos and windows
+func HiddenMinimizeButton(hidden bool) Option {
```
eliasnaur, line 982: Nit: change name to `MinimizeButtonHidden` and separate comment about platform support.

```
// MinimizeButtonHidden hides the minimize button of the window.
//
// MinimizeButtonHidden is supported on Windows and macOS.

**Review by eliasnaur — CHANGES_REQUESTED** [review]
I agree with your point about the minimize button. I've left a review on your change.

File: app/os.go (hunk 2 above)
Comment (kkeybbs) [reply app/os.go:186]: Yes, I'll remove this, and create another pr later for this

File: app/os.go (hunk 4 above)
Comment (kkeybbs) [reply app/os.go:50]: It's okay

File: app/os.go (hunk 3 above)
Comment (kkeybbs) [reply app/os.go:52]: MaxSize  only limit the max size for layout. An example on macOS without HiddenMaximizeButton, after clicking "maximize button", the window becomes a fullscreen with black around the layout.
image

So hiding the maximize button will disallow maximize to fullscreen.
image

File: app/window.go (hunk 5 above)
Comment (kkeybbs) [reply app/window.go:982]: Yes, I's more clearly for reading.

File: app/os.go (hunk 3 above)
Comment (eliasnaur) [reply app/os.go]: Can you imagine a situation where you want to set `MaxSize` to something other than (0, 0) and set `MaximizeButtonHidden` to false? And vice versa, is there a situation where `MaxSize` is (0, 0) and you want `MaximizeButtonHidden` set to true?

If not, then `MaximizeButtonHidden = MaxSize != image.Point{}`, and doesn't need to be exported in Config nor as an Option.

**Comment by kkeybbs** [comment]
I've resolved conflicts with newest code from branch main, the pr commits is clean now.

File: app/os.go (hunk 3 above)
Comment (kkeybbs) [reply app/os.go]: When using app.MaxSize, we still need the maximize button (`app.MaximizeButtonHidden(false)`) on Windows.
- On Windows, clicking maximize button will maximize the window to its' max size (and the window moved to the top-left), not to the screen size.
It's a good and normal experience.
I've commit to rollback changes in issue https://todo.sr.ht/~eliasnaur/gio/631 to keep the maximize button when using app.MaxSize.

- On macOS, clicking maximize button will create a new screen.
But the window will be resized to app.MaxSize, with black area around the it.
image

It's a bad experience, so hidding the maximize button is a good idea when using app.MaxSize.
But I'm not sure if we shoule coupling the two option behaviours: hide maximize button automatically for app.MaxSize.
Or leave them independent, and let the user check the options behaviours.
```
- C.setWindowStandardButtonHidden(window, C.NSWindowZoomButton, toInt(!cnf.Decorated || cnf.MaximizeButtonHidden))
+ C.setWindowStandardButtonHidden(window, C.NSWindowZoomButton, toInt(!cnf.Decorated || cnf.MaximizeButtonHidden || cnf.MaxSize != (image.Point{})))
```

----

When using app.MaximizeButtonHidden(true), without app.MaxSize option, the actual max size will be screen size or a little smaller than screen.
- On Windows, we need app.MaxSize to limit the window size in most scenarios.

- On macOS, sometimes we want a largest window (exclude the menu and dock bar) by double-clicking the title bar, instead of fullscreen mode in a new screen.
Using app.MaximizeButtonHidden(true) without app.MaxSize works in this scenario.

eliasnaur: The issue with macOS fullscreen can be fixed by setting `maxFullScreenContentSize` in addition to `contentMaxSize`. See https://developer.apple.com/documentation/appkit/nswindow/maxfullscreencontentsize?language=objc . With `maxFullScreenContentSize` set, the maximize button behaves the same as on Windows.

So it seems to me we should *never* hide the maximize button. WDYT?

eliasnaur: > I've seen some apps without maximize button, based on Qt. One is designed close to tray, with fixed window size.
> image

I assume for this window `MinSize` == `MaxSize`? If so, Gio could automatically hide the maximize button.

> Another app limits the max window size and disables fullscreen mode.
> image
> 

This case is better served by setting `maxFullScreenContentSize` by Gio, as I noted above. That way, the user will still be able to make the window as large as allowed by pressing the maximize button, which the behaviour you say Windows has. Note that by setting `maxFullScreenContentSize` macOS automatically disables fullscreen mode.

eliasnaur: Let's try `maxFullScreenContentSize` and make Gio a little more automatic for everyone. We can always add `HiddenMaximizeButton` later.

kkeybbs: Changing and predefining the behaviours are hard things, as we want more compatibility and good expreiences at the same time.

Perhaps we could add options **separately**, each one has its own behaviour(s), and less changes in exists behaviours.
I prefer to keep options simple, standalone, and combine when needed.

We may solve some common scenarios, in new issues if some users have new idea for the usage?

For example, I've removed in the PR's newest commit, it's hard to say it takes good effects for before users.
They have to add `app.MinimizeButtonHidden(true)` instead of only setting the `app.MaxSize`.
These two shoule be standalone, but now it's a breaking changing sadly.
image

kkeybbs: I have create the first commit but without the line `style &^= windows.WS_MAXIMIZEBOX`.
I think `app.MaxSize` should not remove the style `WS_MAXIMIZEBOX`.

About `maxFullScreenContentSize`, my native swift demo is still not working with `maxFullScreenContentSize` like Windows OS, but `contentMinSize` takes effects.

So this PR should already provide separate options and enough behaviours about maximize and minimize buttons.

Could we talk about maxFullScreenContentSize in another new PR, maybe?
I'd like to show my swift demo, and talk about the fullscreen's experience on `maxFullScreenContentSize `.

eliasnaur: > I have create the first commit but without the line `style &^= windows.WS_MAXIMIZEBOX`. I think `app.MaxSize` should not override the option `MaximizeButtonHidden`.
> 

I disagree, I want Gio defaults to be hands-off as far as possible and Options should only be for configuration that cannot be automatically deduced.

I've pushed a commit that leaves the maximize button visible on Windows, except for fixed size windows.

> About `maxFullScreenContentSize`, my native swift demo is still not working with `maxFullScreenContentSize` like Windows OS, but `contentMinSize` takes effects.
> 

I pushed a commit that seems to work for me.

kkeybbs: I've rechecked, I also agree to make Gio hands-off. I merged the commit into this PR.

I' glad to see that commit [d50ef68](https://github.com/gioui/gio/commit/d50ef68), I'm interested in to check how it works.

eliasnaur:
Right, so I'm back to the question: when are you going to use `HiddenMaximizeButton`?

kkeybbs:
An **about dialog** like this:
- image
- thought some apps show two gray buttons like this:
- (but I prefer hidding instead of disabling buttons, it looks nice.)
- image

or a simple fixed size main dialog window like this:
- image

The app.MaxSize and app.MinSize still can be set on demand.

In fact I have already rebuilt my qt app  in business, with **gioui**.
image

kkeybbs:
Yes, I've found none.
I agree with you, and I have merged [that commit](https://github.com/gioui/gio/pull/143/files#diff-7541f144700e34176fe3ceb4422fae6b84d0b104fdbd96348a980be51c1c8706R776) from upstream already.
image

It seems we should remove this option MaximizeButtonHidden?

And should we have the same implementation on macOS instead of [these changes](https://github.com/gioui/gio/pull/143/files#diff-b162960493f774fe07a8759b7ad08ff8090b87c3c2d296640159573ac3db7bc8R502)? 
image

It's quite a lot of changes.

PR #144, opened 2024-09-09
app: [android] fix compatibility with older Android versions

inkeliz:
Previously, setHighRefreshRate requires APIs restricted to Android 30, or higher.

Tested on Android 6.0.1 (released on 2015).


inkeliz:
Ping @eliasnaur

eliasnaur:
Thanks, merged!

PR #148, opened 2025-04-23
app: [Windows] correctly center window on startup

P1-support:
When the window is created position it *before* processing the actions to perform (which may include system.ActionCenter). Note that the actions are performed in the callback windowProc().

Fixes: https://todo.sr.ht/~eliasnaur/gio/646


eliasnaur:
Thank you, merged.

PR #149, opened 2025-05-05
use iterators for events

ddkwork:
Signed-off-by: ddkwork

This change will improve the event traversal syntax

	for {
		ev, ok := gtx.Event(pointer.Filter{
			Target: r,
			Kinds:  pointer.Press | pointer.Release,
		})
		if !ok {
			break
		}
	  }

//------------------------------->

	for e := range gtx.Events(pointer.Filter{Target: r,Kinds:pointer.Press | pointer.Release}) {
             ....
	}
Using the range iteration will never traverse to the null pointer, and the return value bool will automatically stop traversing, so it is not necessary to judge bool downstream, and it will be automatically processed at the syntax level


eliasnaur:
Thank you. Some of my comments from #147 doesn't seem applied. Is this still a work-in-progress PR?

ddkwork:
> Thank you. Some of my comments from #147 doesn't seem applied. Is this still a work-in-progress PR?

Yes, this is a cleanup of the first two pr's, and I patiently withdrew almost all of the EVENTS unrelated changes, so please let me know if I missed anything. If this pr passes, I'll be happy to pr the gio-x and example repositories.

io/input/key.go
```diff
@@ -99,12 +100,14 @@ func (q *keyQueue) Reset() {
 	q.dirOrder = q.dirOrder[:0]
 }
 
-func (k *keyHandler) ResetEvent() (event.Event, bool) {
-	if k.reset {
-		return nil, false
+func (k *keyHandler) ResetEvent() iter.Seq[event.Event] {
```
eliasnaur, line 103: This is not a cleanup.

app/doc.go
```diff
@@ -8,7 +8,7 @@ See https://gioui.org for instructions to set up and run Gio programs.
 
 # Windows
 
-A Window is run by calling its Event method in a loop. The first time a
+A Window is run by calling its Events method in a loop. The first time a
```
eliasnaur, line 11: This is not cleanup. Here and below.

app/os.go
```diff
@@ -301,11 +301,7 @@ func (e *eventLoop) FlushEvents() {
 	}
 	e.delivering = true
 	defer func() { e.delivering = false }()
-	for {
-		evt, ok := e.win.nextEvent()
-		if !ok {
-			break
-		}
+	for evt := range e.win.nextEvent() {
```
eliasnaur, line 304: This is not cleanup.

app/window.go
```diff
@@ -36,7 +37,7 @@ type Option func(unit.Metric, *Config)
 // Window represents an operating system window.
 //
 // The zero-value Window is useful; the GUI window is created and shown the first
-// time the [Event] method is called. On iOS or Android, the first Window represents
+// time the [Events] method is called. On iOS or Android, the first Window represents
```
eliasnaur, line 40: Not cleanup.

app/window.go
```diff
@@ -562,46 +563,60 @@ func (c *callbacks) Invalidate() {
...
-		s.cfg = nil
-		return e, true
-	case s.frame != nil:
-		e := *s.frame
-		s.frame = nil
-		s.framePending = true
-		return e.FrameEvent, true
-	case s.wakeup:
-		return wakeupEvent{}, true
+func (c *callbacks) nextEvent() iter.Seq[event.Event] {
```
eliasnaur, line 566: Not purely cleanup.

app/window.go
```diff
@@ -726,11 +741,9 @@ func (w *Window) Event() event.Event {
 		w.init()
 	}
 	if w.driver == nil {
-		e, ok := w.nextEvent()
-		if !ok {
-			panic("window initialization failed without a DestroyEvent")
+		for e := range w.events() {
```
eliasnaur, line 744: Not cleanup.

app/window.go
```diff
@@ -974,7 +987,7 @@ func Decorated(enabled bool) Option {
 
 // flushEvent is sent to detect when the user program
 // has completed processing of all prior events. Its an
-// [io/event.Event] but only for internal use.
+// [io/event.Events] but only for internal use.
```
eliasnaur, line 990: Not cleanup.

eliasnaur (changes requested):
I see, so a preliminary cleaup PR? Sure, but then update the commit subject and message to reflect that. Say:

```all: automated code cleanup

This change includes no functional changes, just cleanup of
style and the use of modern Go constructs.
```

I've marked a handful of changes that are not cleanups, there's probably more. Please look through the PR for similar non-cleanups.

io/input/key.go
ddkwork: When it doesn't use iterators then I don't know how it's supposed to handle it upstream

eliasnaur: The point is that this PR should *only* contain clean-ups, and no functional changes. Then, when this PR is merged, you can make another PR switching to iterators for package app. Finally, yet another PR changing package `io/input` to use iterators.

Otherwise, there are simply too many changes at once, which makes future debugging harder.

ddkwork: You may have misunderstood me, what I was trying to convey at first was:This pr is a withdrawal of iterators by removing changes not related to iterators, as per your previous request. I'll do the pr for the syntax cleanup of go1.23 after the iterator pr is merged.

Now it's iterator pr before, cleanup will be done later, sorry for my bad English.

ddkwork:
I think I should pr in the following order, hopefully the translation engine can accurately translate the Chinese I described:.

1. iterator pr, this pr only does this.
2. next pr, go 1.23 syntax cleanup.
3. go 1.23 syntax cleanup and iterator for gio-x and example, both repositories also pr twice.

In addition, in order to keep the import order of packages consistent with the rest of the format after you and I make changes to the code, I propose that you and I use gofupmt to format and commit the code at the same time, so that if I pull your commits and make subsequent changes, I can make sure that the packages are imported in the same order as the rest of the format and that the differences between the changes will be small. What do you think?

ddkwork:
I've thought about this for a while, please allow me to do a syntax cleanup pr of go1.23 for all three repositories before I close this pr. This way the differences are minimal and it's easier for you to review the code. Please stay tuned for the pr, I will finish these things tonight as I have other things to get done tomorrow, wait for me for a while.

ddkwork:
see here please

https://github.com/gioui/gio/pull/150

ddkwork:
Please check the pr of the other three repositories, in addition, I will pr the other three repositories again after upgrading to the iterator for the latest tests.

ddkwork:
are you online?

PR #153, opened 2025-05-26
layout: avoid heap escapes in Stack and Flex

vsariola:
I looked into the pertinent problem with layout.Flex and layout.Stack creating garbage for every child. The problem in both cases seems to be that the "children" parameter (variadic input slice) of Flex.Layout and Stack.Layout is also used as the scratch space, specifically for op.CallOp. So, both Stack and Flex do something like `children[i].call = c` and later something like `child.call.Add(gtx.Ops)`. For some reason, this seems to confuse the golang escape analysis **and cause everything in children to escape to heap**.

Why this causes a massive number of garbage objects is that if we do something like `Flex{}.Layout(gtx, layout.Rigid(buttonStyle.Layout))`, not only does the FlexChild move to heap but **also the entire buttonStyle** object.

I was able to reduce the number of allocated objects considerably by using temporary slice for the scratch space:

```
	type scratchSpace struct {
		call op.CallOp
		dims Dimensions
	}
	var scratchArray [32]scratchSpace
	scratch := scratchArray[:0]
	scratch = append(scratch, make([]scratchSpace, len(children))...)
```

For the case the number of children <= 32, this has the entire scratch space on stack, and even in the cases children > 32, only the scratch space gets allocated from the heap, the children never move to heap.

What's the end result? For Sointu, the number of allocated objects (alloc_objects in profiler) during 1 minute **dropped from 1.35 million objects to less than 0.35 million***

* Before:
image

* After:
image

Furthermore, we can see that before the change, many of the layout functions were the worst offenders, when sorting the functions by flat, where as after the change, all the layout functions disappeared from the list of worst offenders.

* Before:
image

* After:
image


vsariola:
I had a version which used sync.Pool for the scratchSpace slices that were allocated from the heap, but that seemed a bit overkill; I can change the implementation to that if the rather arbitrary cutoff limit of 32 children on stack induces too much OCD.

vsariola:
Oh, I also realized that if this gets merged, it mostly defeats the purpose of `Background`, which was supposed to be just a more performant implementation of:

```go
layout.Stack{Alignment: layout.C}.Layout(gtx,
    layout.Expanded(background),
    layout.Stacked(widget)
)
```

By the way, I have noticed that `layout.Background` is NOT 1-1 identical to that code. When I replaced a few `Stack`s with `Background`, I noticed the layout breaking; I didn't look into exactly what was happening, but probably `Background` and `Stack` handle Constraints a bit differently. If `Background` is kept, I'd prefer it to be EXACTLY identical to the code above, to limit the number of gotchas. But this is another story.

What got me into looking into optimizing the `Stack` and `Flex` garbage was that I actually needed at least `Foreground` too. For example, to expand a tooltip area over an existing widget. But reimplementing performant variants of every possible alternative got annoying super fast, so I decided to have another look at exactly why `Stack` and `Flex` cause heap escapes. While poking around the `Stack` implementation, I quickly discovered that the variadic arguments alone were not causing the heap escape. With a little experimentation, I zeroed in on the line `child.call.Add(gtx.Ops)` as the heap escape reason: by commenting out the line, no garbage was created.

I am honestly not sure why this is though - `child` here is a local variable within the loop, and thus you'd think that Go would not "know" that `child` came from the `children` slice.

vsariola:
My guess is that in the original implementation, golang is not sure that StackChild.call.ops is not pointing to a stack variable, because it looks a bit like it would be passed as a parameter to the function (even though it's private), so it moves everything to heap instead. But this is just a hunch.

eliasnaur:
Very nice analysis and fix. Thank you. Merged.

PR #155, opened 2025-05-31
internal/debug: go 1.23.8 compat; use strings.Split not strings.SplitSeq

mil:
Currently build fails as go.mod uses go 1.23.8 which doesn't have strings.SplitSeq. 

Note: strings.SplitSeq was introduced in go 1.24.


eliasnaur:
Thanks, merged.

PR #158, opened 2025-09-25
app: [Wayland]  implement missing wayland decoration calls

HTechHQ:
On Linux with Wayland (KWin) the `customdeco` example is not working.

This implements calls to Wayland to set the decoration as configured in the Gio application.

### DIFF
--- a/app/os_wayland.go
+++ b/app/os_wayland.go
@@ -200,6 +200,7 @@ type window struct {
 		dir            f32.Point
 	}
 
+	alreadyDecorated  bool
 	configured        bool
 	lastFrameCallback *C.struct_wl_callback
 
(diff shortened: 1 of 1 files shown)


app/os_wayland.go
```diff
@@ -1061,7 +1061,13 @@ func (w *window) Configure(options []Option) {
 	cnf := w.config
 	cnf.apply(cfg, options)
 	w.config.decoHeight = cnf.decoHeight
-
+	if w.decor != nil && (prev.Decorated != cnf.Decorated || !w.configured) {
```
eliasnaur, line 1064: Why set the mode when `!w.configured`? It seems to me the initial `Decorated` should be set to whatever the display server tells us.

HTechHQ: Thank you for taking a look. 
My goal is to have custom decoration show immediately if I start the app with:
```
window.Option(
    app.Decorated(false),
)
```
I am still trying to understand the full live-cycle and how the app interacts with the server. My best guess right now:
1. `Configure` is called
2. The server always decorates the window and thus the server selected decoration is set in `gio_onToplevelDecorationConfigure`
3. `gio_onXdgSurfaceConfigure` is called and w.configured is set BUT the options are not applied any more, because they were processed in step 1

With the `!w.configured` I got it to work.

Please let me know, if you have more insights into this. It's the first time I interact with Wayland at all. 

Also: because I don't understand the full live-cycle yet, I tried to be defensive. If you know and can confirm that the `w.decor` is never nil this condition could go.

PR #162, opened 2026-02-04
app: set focused attribue on fore/background events for ios and android

placid2000:
This path set focused flag in ConfigEvent when app goes back or foreground on ios and android


eliasnaur:
In our discussion of this change, I wrote:

> I now think the Config.Focused behaviour on Android is (subtly) wrong.
> I'm thinking of `Config.Focused` as "the window has (keyboard) focus",
> that is that the window is the one the user is currently interacting
> with. In that sense, `Config.Focused` should probably change to match
> the `onPause`/`onResume` lifecycle state on Android. The Java
> `OnFocusChangedListener` should just be removed.
> (With that change, it's possible the `window.visible` and `started`
> states can be tracked in an `lifecycle` enum).

Why not use onPause/onResume and what about the focus listener?

placid2000:
Some changes has been made in following commits:
https://github.com/gioui/gio/commit/47ab4c97b2f0c309e3b83af9bfc2e47be7d8daf8
https://github.com/gioui/gio/commit/92fa23b59b544f534d6d3721d0f20ab87371a8

PR #165, opened 2026-03-07
app,gpu,op: [android,ios,macos,windows] add ExternalOp paint for external views

inkeliz:
This change adds paint.ExternalOp, which allows native views, activities, or HWNDs to appear "inside" a Gio layout.

The implementation uses a punch-through method. Gio marks a region as transparent so the layer below becomes visible. One user-defined function receives the area/position and renders there. Before this change, integrating external components such as WebView, Ads, or Camera views had two problems.

1. Gio content could not render above the external view.
2. The external view required an absolute position. That was difficult when combined with offsets or transformations.

The paint.ExternalOp reduces both issues while keeping the API simple. The operation includes a callback that runs every frame. The callback can update position, size, and z-order to keep the native view synchronized with the Gio layout. Platform behavior required small adjustments, mainly on Windows.

Windows uses an extra child window:
MainWindow → GioWindow

This structure allows additional HWNDs to attach to the MainWindow while GioWindow remains on top. For example: MainWindow → GioWindow → WebView

ExternalOp exposes a region of the GioWindow so the underlying HWND, such as a WebView, becomes visible in that area.

On macOS, iOS, and Android the rendering stack already uses additional layers. No structural change was required there, only the background was changed to Transparent.

-----

What the `ExternalOp` do:
1. Make the current clip-area transparent (it will show what is below the Gio layer, in z-index).
2. Ignore events (like clicks and touch) in that clip-area (at OS level, no Gio proxying).
3. Call a user-defined function, which is responsible to render the content in that area.

### DIFF
--- a/app/GioView.java
+++ b/app/GioView.java
@@ -16,6 +16,7 @@ import android.content.Intent;
 import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Matrix;
+import android.graphics.PixelFormat;
 import android.graphics.Rect;
 import android.os.Build;
 import android.os.Bundle;
@@ -55,6 +56,8 @@ import android.view.accessibility.AccessibilityNodeProvider;
 import android.view.accessibility.AccessibilityNodeInfo;
 import android.view.accessibility.AccessibilityEvent;
 import android.view.accessibility.AccessibilityManager;
+import android.view.ViewGroup;
+import android.widget.FrameLayout;
 
 import java.io.UnsupportedEncodingException;
 
@@ -84,9 +87,12 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 		// Late initialization of the Go runtime to wait for a valid context.
 		Gio.init(context.getApplicationContext());
 
+		setZOrderOnTop(true);
+
 		// Set background color to transparent to avoid a flickering
 		// issue on ChromeOS.
 		setBackgroundColor(Color.argb(0, 0, 0, 0));
+		getHolder().setFormat(PixelFormat.TRANSPARENT);
 
 		ViewConfiguration conf = ViewConfiguration.get(context);
 		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
@@ -151,6 +157,22 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 		// so assume it's good for us as well.
 		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
 			requestUnbufferedDispatch(event);
+		}
+
+		// Check if touch event should be handled by Gio or passed through
+		// to external views. Only check at the beginning of a trace so drags
+		// crossing over external regions are not truncated.
+		if (nhandle != 0) {
+			int action = event.getActionMasked();
+			if (action == MotionEvent.ACTION_DOWN || action == MotionEvent.ACTION_POINTER_DOWN) {
+				int idx = event.getActionIndex();
+				float x = event.getX(idx);
+				float y = event.getY(idx);
+				if (!hitTest(nhandle, x, y)) {
+					// Event is on an external view, don't consume it.
+					return false;
+				}
+			}
 		}
 
 		dispatchMotionEvent(event);
@@ -549,6 +571,62 @@ public final class GioView extends SurfaceView implements Choreographer.FrameCal
 		imm.updateCursorAnchorInfo(this, inf);
 	}
 
+	boolean setEmbedViewPosition(View view, int x, int y, int w, int h, int zOrder) {
+		ViewGroup parent = (ViewGroup) this.getParent();
+		boolean isLayoutUpdated = false;
+		
+		if (zOrder == 0) {
+			view.setVisibility(View.GONE);
+			view.requestLayout();
+			return true;
+		}
+		
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+			// On API 21+: use Z for draw ordering without mutating the view
+			// hierarchy, so GioView's SurfaceView surface is never destroyed.
+			if (view.getVisibility() != View.VISIBLE) {
+				view.setVisibility(View.VISIBLE);
+				isLayoutUpdated = true;
+			}
+			if (view.getZ() != (float) zOrder) {
+				// Suppress any outline-based drop shadow the Z value would produce.
+				view.setOutlineProvider(null);
+				view.setZ((float) zOrder);
+				isLayoutUpdated = true;
+			}
+		}
+
+		ViewGroup.LayoutParams lp = view.getLayoutParams();
+		if (lp == null || lp.width != w || lp.height != h) {
+			view.setLayoutParams(new FrameLayout.LayoutParams(w, h));
+			isLayoutUpdated = true;
+		}
+
+		int oldX = (int) view.getX();
+		int oldY = (int) view.getY();
+		if (oldX != x || oldY != y) {
+			view.setX(x);
+			view.setY(y);
+			isLayoutUpdated = true;
+		}
+
+		if (isLayoutUpdated) {
+			view.invalidate();
+			view.requestLayout();
+			this.invalidate();
+		}
+
+		return isLayoutUpdated;
+	}
+
+	public void updateLayout() {
+		ViewGroup parent = (ViewGroup) this.getParent();
+		if (parent != null) {
+			parent.bringChildToFront(this);
+			parent.forceLayout();
+		}
+	}
+
 	static private native long onCreateView(GioView view);
 	static private native void onDestroyView(long handle);
 	static private native void onStartView(long handle);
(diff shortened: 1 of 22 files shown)


ddkwork:
abs pos see this

https://github.com/ddkwork/gio/commit/3e5e5907be4754f3aec1b198e7b09d838112382d

CoyAce:
In my view, there are two core issues here: first, Gio needs to support atomic updates; second, ExternalOp should provide a native view handle to leverage Gio's atomic update API for synchronized rendering. Simply providing a handle without Gio's atomic update capability is meaningless, and therefore the relevant interface should not be exposed at this stage. On a positive note, the current discussion around atomic updates points in a clear direction for future architectural evolution.

CoyAce:
From an architectural standpoint, the handle + atomic update model is clearly the more elegant long-term direction. That said, callback-based integration still holds value—both as a compatibility fallback and as a lower-friction entry point for simpler use cases. Critically, these two approaches should not be coupled within the same API surface. Callback and handle ought to be separate implementations of ExternalOp, not merely two modes of a single implementation. This ensures cleaner abstraction, better testability, and future extensibility without accumulating technical debt.

eliasnaur:
> In my view, there are two core issues here: first, Gio needs to support atomic updates; second, ExternalOp should provide a native view handle to leverage Gio's atomic update API for synchronized rendering. Simply providing a handle without Gio's atomic update capability is meaningless, and therefore the relevant interface should not be exposed at this stage. On a positive note, the current discussion around atomic updates points in a clear direction for future architectural evolution.

Can you explain what's gained by also offering the callback API, other than it's easier to implement? Exposing the atomic-capable API now, even before we have atomic updates, is the right decision: users won't have to change their programs when we implement atomic updates.

> Or.... Another alternative ?

I imagine something like

```go
package paint // import "op/paint"

// EmbedOp embeds an external view at the origin position of
// the current transform. See package app for ways to construct
// EmbedOps.
type EmbedOp struct {
    handle uintptr // or an array of uintptrs, as needed
}

func (o *EmbedOp) Add(o *op.Ops)
```

and

```go
//go:build darwin
package app

// AppKitView represents an AppKit UIView.
type AppKitUIView struct {
   Layer uintptr
}

func (v *AppKitUIView) Op() paint.EmbedOp
```

CoyAce:
Hi Elias, thanks for the thoughtful question.

You're right that exposing an atomic-capable API now would let users adopt it early and avoid breaking changes later—that's a valid forward-looking concern.

After more thought, I realized the tradeoff is more nuanced.

### The real tradeoff: who carries the complexity?

A callback API might seem simpler at first glance, but it actually **shifts complexity from Gio to the user**:

```go
// Callback: user writes this
NewExternalOp(func(visibleArea, elementArea image.Rectangle, zIndex int){
    // User is now responsible for:
    // - Holding the native view reference
    // - Calling platform-specific APIs
    view:=viewFor(handle)
    view.MoveTo(visibleArea,zIndex) // This line hides immense complexity
}).Add(ops)
```

A handle-based API, by contrast, **keeps complexity inside Gio**:

```go
// Handle: user writes this
area:=clip.rect(rect).push()
paint.EmbedOp{handle}.Add(ops)
area.pop()
// gio internally handle like this
view :=  viewFor(handle)
view.MoveTo(rect) // Gio handles all platform details internally

// Gio internally does the hard work:
// - JNI/FFI calls
// - Error handling
```

### Why this matters

For most developers, in most scenarios, they don't want to know about platform details.

They just want their WebView to appear in the right place. **Callback forces them to care about platform details; handle-based API lets them not care.**

### What atomic updates change—and don't change

Atomic updates make handle-based APIs more powerful by adding precise timing. But they don't change the fundamental tradeoff: **callback still offloads complexity to users, handle still encapsulates it.**

### My current view

Rather than viewing callback as a "temporary fallback" and handle as the "real future," I see them as representing a fundamental design choice:

- **Callback**: Gio says "here's the position, you figure out the rest"
- **Handle**: Gio says "give me the view, I'll take care of it"

I believe the latter is more aligned with Gio's philosophy of providing a high-level, cross-platform abstraction. Users who need low-level control can always access the underlying handle directly—but for the 80% case, they shouldn't have to.

Curious to hear your thoughts.

inkeliz:
> Can you explain what's gained by also offering the callback API, other than it's easier to implement? 

I mean it's easier to implement. :P 

One example is WebView2 on Windows. It needs to call `put_bounds` (see https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2host?view=webview2-0.9.430#put_bounds). 

That call can't be done by Gio itself. Gio will only move/resize the sub-window.  The implementer of WebView2 will need to handle and proxy all `WM_SIZE` messages (and similar) to call  WebView2 functions. 

In other words:

1. Gio -> Callback -> (External Code: Calls PutBounds)
2. Gio -> Move HWND -> (External Code: Receive WM_SIZE -> Calls PutBounds)

Well, that is the same "easier" argument. But, it can be easier on the other side too (not only in-Gio).

---

I'll update try to update my code (and my WebView implementation test) to something similar to @eliasnaur suggestion:

```
type AppKitUIView struct {
   Layer uintptr
}
```
```
type WindowsView struct {
   HWND uintptr
}
```
```
type AndroidView struct {
   View uintptr
   SurfaceControl uintptr
}
```

I'll not implement atomicity. Only the resize and movement and clicks/touch coords translation.

-----

_Note: I'm using WebView as example because it's one of the most complex one (it needs to handle events, inputs, display stuff...)_

inkeliz:
I'm currently testing it on a couple of Android, on my iPhone and macOS. The Windows seems to be the most difficult one to integrate with external stuff. The iOS/macOS are almost lag-free, not sure if it's the API or the CPU/GPU performance. _On Android I'm testing with couple of old and low-end devices, so even native apps are slow. _

EDIT: The Android version is laggy even on "high-end" devices (like Motorola Razr 50 Ultra and Google Pixel 9).

inkeliz:
One way to mitigate the "laggy" feeling is setting the background to be the same of the rendered area. I'm not sure if it's worth adding it and I'm not sure how to do that.

Imagine that you have a Yellow box (100% of the screen) and in the middle you have a VideoPlayer (EmbedOp).  If the "root background" is Yellow, any "gap" (between Gio and VideoPlay) would be Yellow. Making it less noticeable. The "gap" can happen during animation.

----

Since this patch is already large enough, I'll keep it as simple as possible (without such mitigation). That can be useful for old Android (without atomic alternatives).

inkeliz:
While testing on Android I noticed one issue: Unisoc.

Sounds like multiple Unisoc devices (like SPC Discovery SE, with Unisoc Tiger T310) doesn't produce a texture transparent. In other words: the "EmbedOp"-area is black, not transparent. 

The workaround is moving away from `SurfaceView` and use `TextureView`. I don't think it's the right way of fixing it.

-----

That is the list of devices that I use for testing:

```
Motorola | Moto G 2nd Gen
Motorola | Moto E6s
Motorola | Moto E14
Motorola | Moto E15
Motorola | Moto G35 5G
Motorola | Moto G 2nd Gen
Xiaomi | Redmi 7A
Xiaomi | Redmi Note 9
Xiaomi | Redmi A3
Xiaomi | Redmi A5
Xiaomi | Redmi Note 12 5G
Xiaomi | Redmi Note 14
Xiaomi | POCO M7 Pro 5G
Samsung | Galaxy A04
Samsung | Galaxy A06
Samsung | Galaxy A13
Samsung | Galaxy A26 5G
Samsung | Galaxy A20e
Samsung | Galaxy A21s
SPC | Discovery SE
Realme | Note 70T
Google | Pixel 9
Huawei | P8 Lite (2017)
Tecno | Spark Go BG6
Vivo | Y22S
Vivo | Y16
```

CoyAce:
How about detecting the hardware manufacturer and selecting the appropriate view type based on the hardware? For example:

```java
import android.os.Build;

public static boolean isUnisocDevice() {
    // API 31+ (Android 12+) has SOC_MANUFACTURER publicly available
    if (Build.VERSION.SDK_INT >= 31) {
        String socManufacturer = Build.SOC_MANUFACTURER;
        return "UNISOC".equalsIgnoreCase(socManufacturer) ||
               "Spreadtrum".equalsIgnoreCase(socManufacturer);
    }
    
    // API < 31 fallback to hardware name detection
    String hardware = Build.HARDWARE.toLowerCase();
    return hardware.contains("ums") ||      // Unisoc chip code prefix
           hardware.contains("unisoc") ||
           hardware.contains("spreadtrum");
}
```

CoyAce:
I looked into it. The `SurfaceControl` API is the officially recommended way to manipulate layers on Android. However, it requires Android 10+, and it's still uncertain whether Unisoc's implementation is reliable. If that approach doesn't work out, we'll ultimately have to fall back to `TextureView`.

inkeliz:
I don’t think “blacklisting” is a good option. Some devices, like the Motorola E14, use Unisoc and work, while others, such as the Xiaomi A3x and SPC Discovery SE, do not. Also, these devices are much slower. On AnTuTu, the difference between the Motorola and SPC is about 30,000 points, but the Gio app is much slower on the SPC.

I will look at how Flutter works (https://github.com/flutter/flutter/blob/master/docs/platforms/android/Texture-Layer-Hybrid-Composition.md).

I will investigate which backend is being used and whether some flag or extension is not available. If there is a way to query the GPU and detect that a required feature is not supported, that would be better than blocking devices based on the GPU name.

CoyAce:
Alright, this is indeed a better practice of pursuing excellence. Looking forward to your good news.

PR #167, opened 2026-03-09
text: render SVG font as black-white outline

inkeliz:
Previously, using SVG fonts will cause Gio to render invisible "characters".

Now, some fonts (like Noto Sans Emoji) will be rendered based on "Outline". Gio don't support SVG fonts, but now it will show the outline ("black-and-white") alternative.


whereswaldon (approved):
Simple and clear. Thank you!

whereswaldon:
Merged, thank you!

To git.sr.ht:~eliasnaur/gio
   c3a6e85f..65d86895  main -> main

PR #171, opened 2026-07-08
app: improve IME composition handling

qianniancn:
Place the Windows IME candidate window under the active composition text.

Also avoid cancelling IME composition when a snippet update only expands
around matching text.

Fixes: https://todo.sr.ht/~eliasnaur/gio/697

### DIFF
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -625,7 +625,7 @@ func (w *window) EditorStateChanged(old, new editorState) {
 		return
 	}
 	defer windows.ImmReleaseContext(w.hwnd, imc)
-	if old.Selection.Range != new.Selection.Range || old.Snippet != new.Snippet {
+	if shouldCancelComposition(old, new) {
 		windows.ImmNotifyIME(imc, windows.NI_COMPOSITIONSTR, windows.CPS_CANCEL, 0)
 	}
 }
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -10,6 +10,7 @@ import (
 	"golang.org/x/sys/windows/registry"
 	"image"
 	"io"
+	"math"
 	"os"
 	"runtime"
 	"sort"
@@ -407,11 +408,7 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 			return windows.TRUE
 		}
 		defer windows.ImmReleaseContext(w.hwnd, imc)
-		sel := w.w.EditorState().Selection
-		caret := sel.Transform.Transform(sel.Caret.Pos.Add(f32.Pt(0, sel.Caret.Descent)))
-		icaret := image.Pt(int(caret.X+.5), int(caret.Y+.5))
-		windows.ImmSetCompositionWindow(imc, icaret.X, icaret.Y)
-		windows.ImmSetCandidateWindow(imc, icaret.X, icaret.Y)
+		w.updateIMEWindows(imc)
 		return windows.TRUE
 	case windows.WM_IME_COMPOSITION:
 		imc := windows.ImmGetContext(w.hwnd)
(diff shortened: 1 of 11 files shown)


app/os_windows.go
```diff
@@ -420,38 +417,56 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
...
+			// RESULTSTR is committed text. Keep it separate from COMPSTR so a
+			// preedit update never looks like a commit.
+			rng := imeRange(state)
+			result := windows.ImmGetCompositionString(imc, windows.GCS_RESULTSTR)
+			start := rng.Start
+			w.w.EditorReplace(rng, result)
+			end := start + utf8.RuneCountInString(result)
+			w.w.SetComposingRegion(key.Range{Start: -1, End: -1})
+			w.w.SetEditorSelection(key.Range{Start: end, End: end})
+			state = w.w.EditorState()
```
whereswaldon, line 434: Given that we commit the composition here and clear the composing region, do we actually need to fall through and execute the rest of this switch case, or can we early-return here?

The only interaction that I see is that maybe there are messages that set `GCS_DELTASTART` or `GCS_CURSORPOS` at the same time as `GCS_RESULTSTR`, but I'm not sure that can happen in practice (what would move the composing region when the composing region no longer exists?).

Early-returning here would help make the various phases of processing this message easier to understand, but we don't need to do it if there actually is a case where we get `GCS_RESULTSTR` and those other positioning messages together.

app/os_windows.go
```diff
@@ -420,38 +417,56 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
...
+				w.w.SetComposingRegion(comp)
+			}
+			if lParam&windows.GCS_CURSORPOS != 0 {
+				rel := windows.ImmGetCompositionValue(imc, windows.GCS_CURSORPOS)
+				pos := state.RunesIndex(state.UTF16Index(rng.Start) + rel)
+				w.w.SetEditorSelection(key.Range{Start: pos, End: pos})
+			}
 		}
-		w.w.SetEditorSelection(key.Range{Start: pos, End: pos})
+		w.updateIMEWindows(imc)
```
whereswaldon: We'd need to defer this if we did the early-return strategy I mentioned above.

widget/editor.go
```diff
@@ -713,6 +723,7 @@ func (e *Editor) layout(gtx layout.Context, textMaterial, selectMaterial op.Call
 		e.scrollCaret = false
 		e.text.ScrollToCaret()
 	}
+	e.updateIMEState(gtx)
```
whereswaldon, line 741: I worry about the overhead of doing this every frame. I know that we don't actually issue a `key.SelectionCmd` unless the selection has changed, but traversing the text regions to resolve the composition bounds every frame seems expensive. Could we instead only do this if we know the selection, contents, scroll position, or layout constraints have changed?

app/os_windows.go
```diff
@@ -420,38 +417,56 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
...
+				start := windows.ImmGetCompositionValue(imc, windows.GCS_DELTASTART)
+				comp.Start = state.RunesIndex(state.UTF16Index(comp.Start) + start)
+			}
+			w.w.SetComposingRegion(comp)
+			pos := end
+			if lParam&windows.GCS_CURSORPOS != 0 {
+				rel := windows.ImmGetCompositionValue(imc, windows.GCS_CURSORPOS)
+				pos = state.RunesIndex(state.UTF16Index(rng.Start) + rel)
+			}
+			w.w.SetEditorSelection(key.Range{Start: pos, End: pos})
```
whereswaldon, line 457: Similarly, here we handle both `GCS_COMPSTR` and any associated `GCS_DELTASTART`/`GCS_CURSORPOS`. Might be an opportunity to early-return?

whereswaldon (changes requested):
Thank you very much for working on this. I have a couple requests, but they aren't very complex:

1. The implementation of the windows IME event handling feels like it could be made easier to comprehend by a refactor focusing on returning early from each category of message. I may have misunderstood the constraints on the implementation though, so feel free to correct me if my requests there aren't possible/practical.
2. The way that the editor is now updating the IME state seems too expensive to do every frame; I'd like to more precisely do that work only when we must.

Overall, I think this is looking really promising. Thanks again.

app/os_windows.go
qianniancn: Good point. I now return after RESULTSTR unless the same message also contains GCS_COMPSTR. In that case we refresh the state and keep processing.

app/os_windows.go
qianniancn: Yep, updateIMEWindows is deferred now, so the early returns won't skip it.

widget/editor.go
qianniancn: Agreed. I added a cheap guard around updateIMEState, so the caret and composition bounds are only recalculated when the relevant editor state changes.

app/os_windows.go
qianniancn: The COMPSTR path now returns after handling DELTASTART and CURSORPOS.

widget/editor.go
```diff
@@ -633,12 +636,24 @@ func (e *Editor) Update(gtx layout.Context) (EditorEvent, bool) {
...
 		Start: start,
 		End:   end,
 	}
+	scrollOff := e.text.ScrollOff()
+	if rng == e.ime.selection.rng &&
+		e.ime.composition == e.ime.lastCompose &&
+		e.text.version == e.ime.textVersion &&
+		scrollOff == e.ime.scrollOff {
+		return
+	}
```
whereswaldon, line 650: ~I'm not certain that the selection range, composition state, version, and scroll offset are sufficient to ensure that we redo this work when we reflow the text without changing its contents.~

~Let's assume that an editor gains or loses width as a consequence of some other layout widget changing size. It will have the same selection state, the same composition state, the same text version, and the same scroll offset (if the editor is not scrolled at all). However, the "caret" information available to the IME **does** need to change if the selected words change their on-screen coordinates due to re-flowing the text. I think that my concerns would be addressed by additionally checking if  `gtx.Constraints` has changed since the last IME state update.~

whereswaldon: After writing the above, I investigated and realized that `e.text.version` increments with each invalidation of the text layout, even those caused by reflow. Thus my comment above is wrong, and this logic looks sufficient.

widget/editor_test.go
```diff
@@ -1142,6 +1142,42 @@ func TestSelectMove(t *testing.T) {
...
+		if len(e.text.regions) == 0 {
+			t.Fatalf("%s: changed IME state did not traverse text regions", name)
+		}
+	}
+	check("selection", func() { e.SetCaret(1, 1) })
+	check("content", func() { e.text.Replace(0, 1, "H") })
+	check("scroll", func() { e.text.ScrollRel(10, 0) })
+	check("constraints", func() {
+		gtx.Constraints = layout.Exact(image.Pt(40, 20))
+		e.text.Layout(gtx, shaper, font.Font{}, unit.Sp(10))
```
whereswaldon, line 1177: Why call layout only on `e.text` instead of `e`?

widget/editor_test.go
```diff
@@ -1142,6 +1142,42 @@ func TestSelectMove(t *testing.T) {
...
+		t.Fatal("unchanged IME state traversed text regions")
+	}
+
+	check := func(name string, change func()) {
+		e.text.regions = nil
+		change()
+		e.updateIMEState(gtx)
+		if len(e.text.regions) == 0 {
+			t.Fatalf("%s: changed IME state did not traverse text regions", name)
+		}
```
whereswaldon, line 1170: I dislike using `e.text.regions` to detect whether or not we've handled the IME change. It's using an implementation detail to detect a side-effect of the change we actually care about. That being said, right now we can't easily mock the `input.Source` within the `gtx` so that we could directly detect the expected `key.SelectionCmd`. @eliasnaur, This feels like an important hole in our ability to test widgets. Am I missing a strategy for checking whether a widget issued a command, or are tests essentially unable to observe that?

whereswaldon: This concern is not blocking the acceptance of this PR, for the record. I just want to discuss it with @eliasnaur, and this situation is a good motivating example.

app/os_windows.go
whereswaldon: Thank you!

app/os_windows.go
whereswaldon: Thank you!

widget/editor.go
whereswaldon: Thank you!

whereswaldon (approved):
Thank you, this is now merged.

```
To git.sr.ht:~eliasnaur/gio
   18034696..c7e3d6c1  main -> main
```

There is still an open comment I'd like to discuss with @eliasnaur , so I'm not going to close out the PR just yet.

widget/editor_test.go
whereswaldon: I pushed an additional commit changing this.

qianniancn:
Thanks for merging this. I agree that using e.text.regions in the test is a bit awkward. I couldn’t find a way to observe key.SelectionCmd directly. Is there a better pattern for this?

whereswaldon:
> Thanks for merging this. I agree that using e.text.regions in the test is a bit awkward. I couldn’t find a way to observe key.SelectionCmd directly. Is there a better pattern for this?

Not presently, but there should be.

PR #172, opened 2026-07-20
internal/stroke: fix angle wraparound

qianniancn:
atan2 wraps at +/-Pi. Subtracting angles on opposite sides of that
boundary made a small turn look like an almost complete circle, causing
round joins to use the wrong direction.

Normalize the difference to the shortest arc and add tests for both
wraparound directions.

Fixes: https://todo.sr.ht/~eliasnaur/gio/698


whereswaldon:
Thank you for working on this! See my comment [here](https://todo.sr.ht/~eliasnaur/gio/698#event-552255). I want to understand the problem fully before committing to a solution.

qianniancn:
On Windows, clip.Stroke sometimes leaves an extra-thick pixel at an RRect join when display scaling is enabled. The behavior changes with both the scale factor and the stroke width. With a 2 dp stroke, I see it at 125% and 150%, but not at 100% or 175%. Using 3 dp or 5 dp changes which scale factors reproduce it. The cause is an atan2 angle wraparound at the ±Pi boundary.

whereswaldon (approved):
Thank you very much, merged!

To git.sr.ht:~eliasnaur/gio
   a7fa0095..2253f8af  main -> main

PR #173, opened 2026-08-03
app: restore a minimized window on system.ActionRaise

dkrisman:
SetForegroundWindow does not un-minimize, so a minimized window ignored
system.ActionRaise entirely. The common consumers of ActionRaise - tray icon
click, single-instance handoff, notification activation - are exactly the
cases where the window is likely minimized.

window.Perform now restores first: if IsIconic reports the window minimized,
call ShowWindow(SW_RESTORE) before the existing SetForegroundWindow +
SetWindowPos(HWND_TOP) sequence. Non-minimized windows take the exact same
path as before. Adds the IsIconic binding and the SW_RESTORE constant.

This composes with the recent "app: [Windows] don't make raised windows
topmost" change (30dc7ff2) in the same switch arm, and mirrors the
accepted macOS fix for the same pain point (activate the app on raise).

Verified on Windows 10 (LTSC 2021): raising from a tray icon restores and
foregrounds a minimized window; raising a visible window is unchanged.

### DIFF
--- a/app/internal/windows/windows.go
+++ b/app/internal/windows/windows.go
@@ -260,6 +260,7 @@ const (
 	SW_SHOWMAXIMIZED = 3
 	SW_SHOWNORMAL    = 1
 	SW_SHOW          = 5
+	SW_RESTORE       = 9
 
 	SWP_FRAMECHANGED  = 0x0020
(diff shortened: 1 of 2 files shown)


eliasnaur:
Thanks, merged.

PR #174, opened 2026-08-03
app,internal/d3d11: present the D3D11 swapchain with the flip model

dkrisman:
The D3D11 backend still presents with the legacy blt model
(DXGI_SWAP_EFFECT_DISCARD, BufferCount 1). On recent drivers this can render
nothing at all: on 2025+ Intel Arc (Lunar Lake 140V) the window stays blank
while frames are drawn. The flip model has been Microsoft's recommended
presentation path since Windows 8.1, and DXGI_SWAP_EFFECT_FLIP_DISCARD
(Windows 10+) is the variant DXGI documentation steers new code toward.

Changes:

- `CreateSwapChain` uses DXGI_SWAP_EFFECT_FLIP_DISCARD with BufferCount 2
  (the flip model requires at least two buffers).
- The flip model rejects sRGB backbuffer formats, so the backbuffer becomes
  DXGI_FORMAT_R8G8B8A8_UNORM and the render target view is created over it
  with DXGI_FORMAT_R8G8B8A8_UNORM_SRGB via a new
  `CreateRenderTargetViewDesc` binding (flattened RENDER_TARGET_VIEW_DESC).
  Linear-to-sRGB conversion on write is preserved, so gamma behavior is
  unchanged.
- The resize path already calls ResizeBuffers with DXGI_FORMAT_UNKNOWN,
  which preserves the non-sRGB buffer format.

Compatibility: DXGI_SWAP_EFFECT_FLIP_DISCARD requires Windows 10. If Gio
still targets Windows 8.x with D3D11, DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL
(Windows 8+) is a drop-in alternative; happy to switch or add a fallback if
that floor matters.

Verified on Windows 10 (LTSC 2021, Intel and NVIDIA adapters) and on the
Intel Arc / Lunar Lake 140V machine that motivated the change: the
previously blank window renders correctly, resize/minimize/restore behave,
and colors are unchanged against the blt-model build on hardware where both
work.

### DIFF
--- a/internal/d3d11/d3d11_windows.go
+++ b/internal/d3d11/d3d11_windows.go
@@ -664,7 +664,10 @@ const (
 
 	MAP_READ = 1
 
-	DXGI_SWAP_EFFECT_DISCARD = 0
+	DXGI_SWAP_EFFECT_DISCARD      = 0
+	DXGI_SWAP_EFFECT_FLIP_DISCARD = 4
+
+	RTV_DIMENSION_TEXTURE2D = 4
 
 	FEATURE_LEVEL_9_1  = 0x9100
 	FEATURE_LEVEL_9_3  = 0x9300
@@ -1038,14 +1041,25 @@ func (d *Device) CreateTexture2D(desc *TEXTURE2D_DESC) (*Texture2D, error) {
 	return tex, nil
 }
 
-func (d *Device) CreateRenderTargetView(res *Resource) (*RenderTargetView, error) {
+// RENDER_TARGET_VIEW_DESC is D3D11_RENDER_TARGET_VIEW_DESC with its
+// trailing union flattened; the largest member is D3D11_TEX2D_ARRAY_RTV
+// at 12 bytes. For a RTV_DIMENSION_TEXTURE2D view only MipSlice is read.
+type RENDER_TARGET_VIEW_DESC struct {
+	Format        uint32
+	ViewDimension uint32
+	MipSlice      uint32
+	FirstArrSlice uint32
+	ArraySize     uint32
+}
+
+func (d *Device) CreateRenderTargetView(res *Resource, desc RENDER_TARGET_VIEW_DESC) (*RenderTargetView, error) {
 	var target *RenderTargetView
 	r, _, _ := syscall.Syscall6(
 		d.Vtbl.CreateRenderTargetView,
 		4,
 		uintptr(unsafe.Pointer(d)),
 		uintptr(unsafe.Pointer(res)),
-		0, // pDesc
+		uintptr(unsafe.Pointer(&desc)),
 		uintptr(unsafe.Pointer(&target)),
 		0, 0,
 	)
@@ -1645,18 +1659,24 @@ func CreateSwapChain(dev *Device, hwnd windows.Handle) (*IDXGISwapChain, error) 
 	}
 	swchain, err := (*IDXGIFactory)(unsafe.Pointer(dxgiFactory)).CreateSwapChain(
 		(*IUnknown)(unsafe.Pointer(dev)),
+		// The flip presentation model is required by modern drivers;
+		// several of them mishandle the legacy blt model and present
+		// nothing at all. Flip forbids an sRGB backbuffer format and
+		// needs at least two buffers, so gamma correctness moves to the
+		// sRGB render target view the caller creates over the
+		// backbuffer.
 		&DXGI_SWAP_CHAIN_DESC{
 			BufferDesc: DXGI_MODE_DESC{
-				Format: DXGI_FORMAT_R8G8B8A8_UNORM_SRGB,
+				Format: DXGI_FORMAT_R8G8B8A8_UNORM,
 			},
 			SampleDesc: DXGI_SAMPLE_DESC{
 				Count: 1,
 			},
 			BufferUsage:  DXGI_USAGE_RENDER_TARGET_OUTPUT,
-			BufferCount:  1,
+			BufferCount:  2,
 			OutputWindow: hwnd,
 			Windowed:     1,
-			SwapEffect:   DXGI_SWAP_EFFECT_DISCARD,
+			SwapEffect:   DXGI_SWAP_EFFECT_FLIP_DISCARD,
 		},
 	)
 	IUnknownRelease(unsafe.Pointer(dxgiFactory), dxgiFactory.Vtbl.Release)
--- a/app/d3d11_windows.go
+++ b/app/d3d11_windows.go
@@ -94,7 +94,13 @@ func (c *d3d11Context) Refresh() error {
 		return err
 	}
 	texture := (*d3d11.Resource)(unsafe.Pointer(backBuffer))
-	renderTarget, err := c.dev.CreateRenderTargetView(texture)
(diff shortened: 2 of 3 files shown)


internal/d3d11/d3d11_windows.go
```diff
@@ -1645,18 +1679,24 @@ func CreateSwapChain(dev *Device, hwnd windows.Handle) (*IDXGISwapChain, error)
...
 			BufferDesc: DXGI_MODE_DESC{
-				Format: DXGI_FORMAT_R8G8B8A8_UNORM_SRGB,
+				Format: DXGI_FORMAT_R8G8B8A8_UNORM,
 			},
 			SampleDesc: DXGI_SAMPLE_DESC{
 				Count: 1,
 			},
 			BufferUsage:  DXGI_USAGE_RENDER_TARGET_OUTPUT,
-			BufferCount:  1,
+			BufferCount:  2,
```
eliasnaur: Are you sure changing BufferCount to 2 won't fix DXGI_SWAP_EFFECT_DISCARD? The [documentation](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_effect) says

> DXGI_SWAP_EFFECT_DISCARD ... This flag is valid for a swap chain with **more than one** back buffer

If so, please submit this change as a separate bugfix commit, with the switch to DXGI_SWAP_EFFECT_FLIP_DISCARD on top.

internal/d3d11/d3d11_windows.go
```diff
@@ -1055,6 +1058,37 @@ func (d *Device) CreateRenderTargetView(res *Resource) (*RenderTargetView, error
...
+	ViewDimension uint32
+	MipSlice      uint32
+	FirstArrSlice uint32
+	ArraySize     uint32
+}
+
+// CreateRenderTargetViewDesc is CreateRenderTargetView with an explicit
+// view description, which is required to view a backbuffer through a
+// different (for example sRGB) format than it was created with.
+func (d *Device) CreateRenderTargetViewDesc(res *Resource, desc *RENDER_TARGET_VIEW_DESC) (*RenderTargetView, error) {
```
eliasnaur: Don't introduce a new function. Just change the existing `CreateRenderTargetView` to accept a descriptor. Pass the descriptor by value, not pointer.

internal/d3d11/d3d11_windows.go
```diff
@@ -1645,18 +1679,24 @@ func CreateSwapChain(dev *Device, hwnd windows.Handle) (*IDXGISwapChain, error)
...
 			SampleDesc: DXGI_SAMPLE_DESC{
 				Count: 1,
 			},
 			BufferUsage:  DXGI_USAGE_RENDER_TARGET_OUTPUT,
-			BufferCount:  1,
+			BufferCount:  2,
 			OutputWindow: hwnd,
 			Windowed:     1,
-			SwapEffect:   DXGI_SWAP_EFFECT_DISCARD,
+			SwapEffect:   DXGI_SWAP_EFFECT_FLIP_DISCARD,
```
eliasnaur, line 1679: I vaguely remember DXGI_SWAP_EFFECT_FLIP_DISCARD not working great with window resizes. Do you see any artifacts during resizing, or is the content perfectly synchronized (that is, not stretching, padding etc.)?

eliasnaur (changes requested):
Looks good, a few comments below.

Please note in the commit description that this change breaks compatibility with Windows versions < 10. I think that's ok given that recent Go versions also require Windows 10.

internal/d3d11/d3d11_windows.go
dkrisman: I tested this on the affected hardware (Arc 140V): DXGI_SWAP_EFFECT_DISCARD with BufferCount 2 still presents nothing - the window stays blank exactly as with BufferCount 1, while FLIP_DISCARD renders correctly in the same harness (minimal solid-fill window, automated pixel capture, repeated runs). Buffer count alone is not a fix on this hardware, so the change stays a single commit, with the negative result noted in the commit message.

internal/d3d11/d3d11_windows.go
dkrisman: CreateRenderTargetViewDesc is gone; CreateRenderTargetView takes RENDER_TARGET_VIEW_DESC by value. The texture render-target path passes its texture format explicitly, which resolves to the same view as the previous NULL descriptor (resource format, mip 0). gio's test suite passes, including gpu/headless and gpu/internal/rendertest on the D3D11 backend.

internal/d3d11/d3d11_windows.go
dkrisman: No stretching in automated probes on the Arc 140V (60 programmatic size changes plus an interactive corner drag through the modal resize loop, about 100 rect-verified captures - content covered the client area in every one). Resizing by hand, though, a transient white band (a frame or two) can appear along the shrinking edge - bottom when shrinking vertically, right when shrinking horizontally - but only within the first ~5 seconds of resizing after application start; after that it never reproduces. IDXGIDevice1::SetMaximumFrameLatency(1) makes no difference, so it does not look like present-queue depth. I cannot compare against DISCARD on this hardware because the blt path presents nothing at all there. I can dig into the startup-window artifact here or as a follow-up, whichever you prefer.

dkrisman:
The commit message now notes the Windows 10 requirement: FLIP_DISCARD is Windows 10+, so this drops Windows 8.x, in line with recent Go releases. The update also reworks the RTV binding and records the BufferCount test result, per the comments below.

internal/d3d11/d3d11_windows.go
eliasnaur: Alright. Perhaps this is how you tested it, but: does rendering work on your setup if you keep everything else (the sRGB changes etc.) and only keep DXGI_SWAP_EFFECT_DISCARD?

internal/d3d11/d3d11_windows.go
eliasnaur: Obviously, no rendering output is worse than flickering during resize. However, resize flicker is such a jarring artifact that I want to push back against switching to DXGI_SWAP_EFFECT_FLIP_DISCARD without adressing resizing.

I see two ways:

- Switching to DXGI_SWAP_EFFECT_FLIP_DISCARD and ensuring synchronization with window resize repainting. This is the better way for performance, future compatibility, but probably also the most intrusive. Gemini suggests some combination of `DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT`, SetMaximumFrameLatency(1) and a ID3D11Query (inserted after Present and then waited on).
- Fixing DXGI_SWAP_EFFECT_DISCARD on your setup. It surprises me that your graphics driver completely breaks DXGI_SWAP_EFFECT_DISCARD, because surely there must be a bunch of legacy programs that uses DXGI_SWAP_EFFECT_DISCARD that would also break. For example, does a minimal D3D11 example also break on your setup? If not, perhaps a fix can be bisected.

eliasnaur:
https://raphlinus.github.io/rust/gui/2019/06/21/smooth-resize-test.html talks about the problem more. It seems the problem is thorny on Windows (we had a similar resizing issue on macOS until someone figured out the magic recipe for synchronization).

PR #177, opened 2026-08-12
app: [Wayland] add IME support

bigwolfeman:
## Summary

- bind `text-input-v3` for each Wayland seat
- apply preedit, commit, and surrounding-text deletion as atomic `done` transactions
- synchronize protocol serials and convert between Gio rune offsets and UTF-8 byte offsets
- report surrounding text, cursor bounds, and content purpose to the input method

Related discussion:
https://lists.sr.ht/~eliasnaur/gio/%3C87tsyhhjkm.fsf@inet.name%3E

## Testing

- `go test ./app -run '^TestTextInput' -count=1`
- `go test ./app -count=1`
- `go build ./app`
- live Pinyin composition with Mango and fcitx5, including preedit, candidate commit, selection replacement, cancellation, rapid repeated composition, and focus restoration
- live `input-method-v2` deletion across the multibyte text `é🙂`, followed by a commit in the same transaction


app/os_wayland.go
```diff
@@ -104,13 +105,22 @@ type wlDisplay struct {
...
+	im         *C.struct_zwp_text_input_v3
+	imFocus    *window
+	imSerial   uint32
+	imActive   bool
+	imDirty    bool
+	imExternal bool
+	imHint     key.InputHint
+	pending    textInputUpdate
+	sent       textInputSnapshot
+	sentValid  bool
```
eliasnaur: This is quite a bit of state to keep in the driver. I haven't looked closely at the changes (and the PR is in draft), but my impression is that some of this should be subsumed by the general IM machinery in `ime.go` and `ime_test.go`. In short, drivers are meant to be minimal because they're plural and hard to test. Even if some machinery is required just for one platform, it's usually better to implement that in general purpose Go to make it hackable and testable everywhere.

bigwolfeman:
Changed based on your input. Created imState to maintain the state that is not necessarily Wayland unique. Driver is kept as minimal as possible.

whereswaldon:
Are you keeping this as a draft because it is unfinished? I'm not sure whether to review it in depth or not.

bigwolfeman:
> Are you keeping this as a draft because it is unfinished? I'm not sure whether to review it in depth or not.

Didn't mean to leave it as a draft. I have changed it.

app/os_wayland.go
```diff
@@ -111,6 +111,10 @@ type wlSeat struct {
 	touch    *C.struct_wl_touch
 	keyboard *C.struct_wl_keyboard
 	im       *C.struct_zwp_text_input_v3
+	imFocus  *window
+	imActive bool
+	imHint   key.InputHint
+	imState  textInputState
```
eliasnaur, line 117: Nit: group related fields with an anonymous struct:

```
   keyboard *C.struct_wl_keyboard
   im struct {
      handle *C.struct_zwp_text_input_v3
      focus  *window
	  active bool
	  hint   key.InputHint
	  state  textInputState
   }
```

eliasnaur (commented):
Did a superficial look and nothing stood out. I'll let Chris take a look and test it.

app/ime.go
```diff
@@ -164,3 +164,246 @@ func snippetSubstring(s key.Snippet, r key.Range) string {
...
+		text:   state.Snippet.Text[start:end],
+		start:  state.Snippet.Start + startRunes,
+		cursor: cursor - start,
+		anchor: anchor - start,
+	}, surroundingReady
+}
+
+// deleteRange converts a surrounding-text deletion, expressed in UTF-8
+// bytes relative to the selection, into a rune range.
+func (s textInputSnapshot) deleteRange(before, after uint32) (key.Range, bool) {
```
whereswaldon: Can we rename this to reflect the fact that it doesn't actually delete the range, it just converts it to a rune range within the editor? Perhaps call it `convertRange` or something?

app/ime.go
```diff
@@ -164,3 +164,246 @@ func snippetSubstring(s key.Snippet, r key.Range) string {
...
+		e.SetEditorSelection(selection)
+	}
+	return e.EditorState() != beforeState
+}
+
+// applyDone applies the pending update and reports whether the editor
+// state changed.
+func (s *textInputState) applyDone(e *callbacks, serial uint32) bool {
+	s.dirty = s.apply(e) || s.dirty
+	return serial == s.serial && s.dirty
```
whereswaldon: Note to self: check whether we respect the semantics of the `done` event:

> When the client receives a done event with a serial different than the number of past commit requests, it must proceed with evaluating and applying the changes as normal, except it should not change the current state of the zwp_text_input_v3 object. All pending state requests (set_surrounding_text, set_content_type and set_cursor_rectangle) on the zwp_text_input_v3 object should be sent and committed after receiving a zwp_text_input_v3.done event with a matching serial.

app/ime.go
```diff
@@ -164,3 +164,246 @@ func snippetSubstring(s key.Snippet, r key.Range) string {
...
+	cursor, ok := runeOffsetToByte(state.Snippet.Text, sel.Start-state.Snippet.Start)
+	if !ok {
+		return textInputSnapshot{}, surroundingUnavailable
+	}
+	anchor, ok := runeOffsetToByte(state.Snippet.Text, sel.End-state.Snippet.Start)
+	if !ok {
+		return textInputSnapshot{}, surroundingUnavailable
+	}
+	start, end := min(cursor, anchor), max(cursor, anchor)
+	if end-start > maxSurroundingTextBytes {
```
whereswaldon, line 339: I don't think I fully understand this case. Are we not in control of the snippet length? Under what circumstances would it exceed our maximum?

app/ime.go
```diff
@@ -164,3 +164,246 @@ func snippetSubstring(s key.Snippet, r key.Range) string {
...
+		Start: max(0, sel.Start-contextRunes),
+		End:   sel.End + contextRunes,
+	})
+}
+
+// prepareSurroundingText returns the surrounding text.
+func prepareSurroundingText(e *callbacks) (textInputSnapshot, surroundingStatus) {
+	snapshot, status := textInputSurrounding(e.EditorState())
+	if status == surroundingNeedsSnippet {
+		requestSurroundingText(e)
```
whereswaldon, line 417: So if we determine that we need a snippet, and we create one with `requestSurroundingText`, should we re-run `textInputSurrounding` now that we have a snippet? I'm surprised that we return `status = surroundingNeedsSnippet` after creating a snippet.

app/ime_test.go
```diff
@@ -264,3 +267,276 @@ func TestEditorCompositionBounds(t *testing.T) {
...
+			t.Fatalf("state after empty done = %+v, want %+v", got, state)
+		}
+	})
+
+	for _, tc := range []struct {
+		name    string
+		pending textInputUpdate
+	}{
+		{name: "empty commit", pending: textInputUpdate{commitSet: true}},
+		{name: "empty preedit", pending: textInputUpdate{preeditSet: true}},
```
whereswaldon: Let's just define the desired snippet text and compose range as additional fields on the table test struct, instead of adding conditional logic based on the name of the testcase in the test loop body.

app/ime_test.go
```diff
@@ -264,3 +267,276 @@ func TestEditorCompositionBounds(t *testing.T) {
...
+		if snapshot.cursor < 0 || snapshot.cursor > len(snapshot.text) ||
+			snapshot.cursor < len(snapshot.text) && !utf8.RuneStart(snapshot.text[snapshot.cursor]) {
+			t.Fatalf("cursor %d is not a valid UTF-8 boundary", snapshot.cursor)
+		}
+	})
+
+	t.Run("rejects selection larger than limit", func(t *testing.T) {
+		text := strings.Repeat("界", 1400)
+		state := textInputTestState(text, key.Range{Start: 0, End: 1400})
+		if _, status := textInputSurrounding(state); status != surroundingUnavailable {
```
whereswaldon, line 591: It is entirely likely that users will use ctrl+A or similar shortcuts to select the entire contents of large edit buffers. Is it really desirable for the IME to just not get surrounding text content in that case? I'm asking out of ignorance, I really don't know.

app/os_wayland.go
```diff
@@ -1561,30 +1589,161 @@ func gio_onKeyboardRepeatInfo(data unsafe.Pointer, keyboard *C.struct_wl_keyboar
...
+		s.imState.pending = textInputUpdate{}
+		return
+	}
+	if s.imState.applyDone(s.imFocus.w, uint32(serial)) && s.imActive {
+		if s.imHint != s.imFocus.inputHint {
+			s.activateTextInput(s.imFocus)
+			return
+		}
+		snapshot, status := prepareSurroundingText(s.imFocus.w)
+		if status != surroundingNeedsSnippet {
```
whereswaldon: Here again, I'm left wondering why we need `prepareSurroundingText` to ever return `surroundingNeedsSnippet`. If we could generate the snippet in the implementation of that method, we wouldn't need extra conditional logic at this level.

Also, what is the expected/intended flow from here if we do get `surroundingNeedsSnippet`? If I understand correctly, we'll skip committing our protocol state entirely until the next time the compositor sends us a done event. That doesn't seem desirable, as we will have already created the snippet we needed in the implementation of `prepareSurroundingText`.

app/os_wayland.go
```diff
@@ -1802,11 +1961,99 @@ func (w *window) surface() (*C.struct_wl_surface, int, int) {
...
+		s.imState.external = false
+		return
+	}
+	cancelled := shouldCancelComposition(old, new) && new.compose.Start != -1
+	if cancelled {
+		s.imState.clearPreedit(w.w)
+	}
+	surroundingChanged := cancelled || selectionChanged || old.Snippet != new.Snippet
+	caretChanged := old.Selection.Caret != new.Selection.Caret || old.Selection.Transform != new.Selection.Transform
+	committed := false
```
whereswaldon: This variable should be called `needsCommit` or similar, as it being true indicates that we've applied protocol messages that require us to perform a commit.

app/os_wayland.go
```diff
@@ -1802,11 +1961,99 @@ func (w *window) surface() (*C.struct_wl_surface, int, int) {
...
+		if status == surroundingReady {
+			s.sendSurroundingText(snapshot)
+		}
+		cause := uint32(C.ZWP_TEXT_INPUT_V3_CHANGE_CAUSE_INPUT_METHOD)
+		if s.imState.external {
+			cause = uint32(C.ZWP_TEXT_INPUT_V3_CHANGE_CAUSE_OTHER)
+		}
+		C.zwp_text_input_v3_set_text_change_cause(s.im, C.uint32_t(cause))
+		s.sendCursorRectangle(w)
+		s.commitTextInput()
```
whereswaldon: I don't really understand why this function can commit twice. Why not figure out all of the various protocol interactions we want to perform, and then commit them all at once? It would eliminate the need for the variable tracking whether to perform the second commit.

app/os_wayland.go
```diff
@@ -1802,11 +1961,99 @@ func (w *window) surface() (*C.struct_wl_surface, int, int) {
...
+	if s.imHint != w.inputHint {
+		s.activateTextInput(w)
+		return
+	}
+	selectionChanged := old.Selection.Range != new.Selection.Range
+	textChanged := selectionChanged || !areSnippetsConsistent(old.Snippet, new.Snippet)
+	if s.imState.dirty {
+		s.imState.external = s.imState.external || textChanged
+		snapshot, status := prepareSurroundingText(w.w)
+		if status == surroundingNeedsSnippet {
```
whereswaldon: Another place where we could simplify if `prepareSurroundingText` internally handled the snippet entirely.

app/ime.go
whereswaldon: I believe we do properly handle this, or mostly so. This function will never return true if the serial mismatches, and we only refresh our IME state at the protocol level if this function returns true.

I am left wondering about what happens if the compositor sends us a stale update and then simply doesn't send a non-stale one. According to spec, we should avoid making any protocol requests until we get a non-stale serial. We do not appear to have a lock-out mechanism that would prevent this in our implementation. Additionally, the only way to recover from such a compositor bug might be to restart the IME interaction entirely. What do you think @bigwolfeman?

whereswaldon (changes requested):
Firstly, let me say a huge "thank you" for working on this! It's very exciting to see IME support working on Wayland.

I've looked through the code, and I've tested it on Wayland (niri + IBus). The behaviors that I know how to test look good. However, I do not have enough literacy in IMEs to really exhaustively test it. I was able to insert emoji in a Gio editor with IBus, which I do understand how to do. It seemed to work well.

I've left some comments and questions throughout the implementation where I thought there were opportunities to improve things or where I didn't understand something. I think this is close to being ready for merge.

Thanks again!

app/ime_test.go
bigwolfeman: This is where there is a major complication. So Chinese runes are roughly 3 bytes per rune and emojis are roughly 4 on average.

Wayland specifies surrounding text cannot exceed 4000 bytes and must express cursor and anchor as byte offsets of the buffer.

So if someone ctrl + A's 2000 Chinese characters, visually it could look like they have it all. The selection exists in Gio, but it cannot all fit in the Wayland surrounding text buffer. If we are outside that buffer, the IME no longer has an accurate representation of the full selection, even though Gio does.


set_surrounding_text(text, cursor, anchor)

So if we set the cursor or anchor outside the bounds of the actual Wayland buffer, that is not good.

This is the solution GTK uses.

app/ime.go
bigwolfeman: Yes, this definitely is an issue. I will change it to have a lock out.

app/os_wayland.go
bigwolfeman: Yes. Doing this as a helper function can also apply to func gio_onTextInputDone(data unsafe.Pointer, im *C.struct_zwp_text_input_v3, serial C.uint32_t) dramatically cleans up both functions

app/ime.go
bigwolfeman: renamed to surroundingAwaitingSnippet to make this more clear. The issue is that the snippet is created async. 

requestSurroundingText calls e.SetEditorSnippet(rng) (window.go:466) emits an event and the snippet is untouched. EditorStateChanged is where the retry of prepareSurroundingText with a snippet that actually covers the selection happens.

app/os_wayland.go
bigwolfeman: commented on here: https://github.com/gioui/gio/pull/177#discussion_r3809975528

app/ime_test.go
bigwolfeman: I put a comment in ime.go about it. Wayland's behavior may change on this in the future.

whereswaldon:
I received quite a few email notifications about comments on this PR, but the comments appear to have been deleted. Some of them refer to implementation choices that are not present in the current revision of the PR. What happened there? Did you refactor this several times, deleting old comments between each revision? @bigwolfeman

bigwolfeman:
> I received quite a few email notifications about comments on this PR, but the comments appear to have been deleted. Some of them refer to implementation choices that are not present in the current revision of the PR. What happened there? Did you refactor this several times, deleting old comments between each revision? @bigwolfeman

I had cursor do the final push and it elected to comment on every single comment in this chain after I had already done so by hand. So I deleted all of them.

whereswaldon (approved):
Thank you, merged.

To git.sr.ht:~eliasnaur/gio
   c035a619..dee6153c  main -> main

PR #178, opened 2026-08-13
widget: allow selection to clear at edge of textbox with arrow key

samsrdc:
This pull request fixes editor behaviour to allow selection to be cleared when either:
a) The selection extends to the first character in the editor, and the user presses the left/up arrow keys.
b) The selection extends to the last character in the editor, and the user presses the right/down arrow keys.

Added ModCtrl to the filters as it should also take effect on any key modifier.

The previous behaviour without this patch is that when the selection extends to the start/end, the user cannot clear the selection when pressing an arrow key in the direction of the start/end.

### DIFF
--- a/widget/editor.go
+++ b/widget/editor.go
@@ -387,10 +387,10 @@ func (e *Editor) processKey(gtx layout.Context) (EditorEvent, bool) {
 		key.Filter{Focus: e, Name: key.NameEnd, Optional: key.ModShortcut | key.ModShift},
 		key.Filter{Focus: e, Name: key.NamePageDown, Optional: key.ModShift},
 		key.Filter{Focus: e, Name: key.NamePageUp, Optional: key.ModShift},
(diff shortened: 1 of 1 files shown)


samsrdc:
Just noting this came from an original patch here https://lists.sr.ht/~eliasnaur/gio-patches/patches/70445 but moving forward I think Github is easier for me.

inkeliz:
It seems your commit is missing the "Signed-off-by". [Use git commit -s](https://gioui.org/doc/contribute) to add it.

samsrdc:
Should be done now, although I used git commit --amend -S (edit: now with --signoff)

Let me know if there is an issue.

whereswaldon (approved):
Thank you very much, merged.

```
To git.sr.ht:~eliasnaur/gio
   0b2f18d2..f3f19973  main -> main
```

PR #180, opened 2026-08-16
app: [Windows] fix custom decoration after restore

qianniancn:
On Windows with extended displays, restoring a minimized undecorated
window can leave hover state and hit testing in the wrong state.

Clear the pointer state and refresh the non-client area when restoring.

Fixes: https://todo.sr.ht/~eliasnaur/gio/702
Fixes: https://todo.sr.ht/~eliasnaur/gio/703

### DIFF
--- a/app/internal/windows/windows.go
+++ b/app/internal/windows/windows.go
@@ -244,6 +244,7 @@ const (
 
 	MDT_EFFECTIVE_DPI = 0
 
+	MONITOR_DEFAULTTONULL    = 0
 	MONITOR_DEFAULTTOPRIMARY = 1
 
 	NI_COMPOSITIONSTR = 0x0015
@@ -451,7 +452,7 @@ var (
 	_LoadCursor                  = user32.NewProc("LoadCursorW")
 	_LoadImage                   = user32.NewProc("LoadImageW")
 	_MonitorFromPoint            = user32.NewProc("MonitorFromPoint")
-	_MonitorFromWindow           = user32.NewProc("MonitorFromWindow")
+	_MonitorFromRect             = user32.NewProc("MonitorFromRect")
 	_MoveWindow                  = user32.NewProc("MoveWindow")
 	_MsgWaitForMultipleObjectsEx = user32.NewProc("MsgWaitForMultipleObjectsEx")
 	_OpenClipboard               = user32.NewProc("OpenClipboard")
@@ -710,12 +711,15 @@ func GetWindowPlacement(hwnd syscall.Handle) WindowPlacement {
 	return wp
 }
 
-func GetMonitorInfo(hwnd syscall.Handle) MonitorInfo {
+func GetMonitorInfo(r Rect, flags uint32) (MonitorInfo, bool) {
+	monitor, _, _ := _MonitorFromRect.Call(uintptr(unsafe.Pointer(&r)), uintptr(flags))
+	if monitor == 0 {
+		return MonitorInfo{}, false
+	}
 	var mi MonitorInfo
 	mi.cbSize = uint32(unsafe.Sizeof(mi))
-	v, _, _ := _MonitorFromWindow.Call(uintptr(hwnd), MONITOR_DEFAULTTOPRIMARY)
-	_GetMonitorInfo.Call(v, uintptr(unsafe.Pointer(&mi)))
-	return mi
+	_GetMonitorInfo.Call(monitor, uintptr(unsafe.Pointer(&mi)))
+	return mi, true
 }
 
 func GetWindowLong(hwnd syscall.Handle, index uintptr) (val uintptr) {
--- a/app/os_windows.go
+++ b/app/os_windows.go
@@ -373,7 +373,10 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 		// state. See https://devblogs.microsoft.com/oldnewthing/20150304-00/?p=44543.
 		// Note that trying to do the adjustment in WM_GETMINMAXINFO is ignored by Windows.
 		szp := (*windows.NCCalcSizeParams)(unsafe.Pointer(lParam))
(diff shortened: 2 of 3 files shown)


app/os_windows.go
```diff
@@ -51,6 +51,7 @@ type window struct {
 
 	borderSize image.Point
 	config     Config
+	minimized  bool
```
eliasnaur: Tracking minimized state is unfortunate. Instead of this field, why not update the window state in the Windows message that restores the window?

app/os_windows.go
```diff
@@ -245,6 +255,11 @@ func (w *window) update() {
 	}
 	w.ProcessEvent(ConfigEvent{Config: w.config})
 	w.draw(true)
+	if restored && !w.config.Decorated {
+		// Refresh client-area hit testing after restoring.
+		windows.SetWindowPos(w.hwnd, 0, 0, 0, 0, 0,
+			windows.SWP_NOMOVE|windows.SWP_NOSIZE|windows.SWP_NOZORDER|windows.SWP_FRAMECHANGED)
```
eliasnaur: Can you say why this call is necessary? The documentation for `SWP_FRAMECHANGED` states that

> Applies new frame styles set using the SetWindowLong function. Sends a WM_NCCALCSIZE message to the window, even if the window's size is not being changed.

But we only call `SetWindowLong` in `Configure`. Perhaps our `WM_NCCALCSIZE` handler doesn't calculate the correct client size?

app/os_windows.go
```diff
@@ -236,6 +244,8 @@ func (w *window) update() {
 	)
 	style := windows.GetWindowLong(w.hwnd, windows.GWL_STYLE)
 	switch {
+	case p.IsMinimized():
+		w.config.Mode = Minimized
```
eliasnaur: Nit: this seems to belong in a separate change.

app/os_windows.go
```diff
@@ -217,6 +218,13 @@ func (w *window) init() error {
 // If anything has changed it emits a ConfigEvent to notify the application.
 func (w *window) update() {
 	p := windows.GetWindowPlacement(w.hwnd)
+	minimized := p.IsMinimized()
+	restored := w.minimized && !minimized
+	w.minimized = minimized
+	if restored {
+		// Clear stale pointer state after restoring.
+		w.ProcessEvent(pointer.Event{Kind: pointer.Cancel})
+	}
```
eliasnaur: Why cancel pointer event this late, and not when the window loses focus, or when minimized?

qianniancn:
@eliasnaur  Thanks , I updated the patch.
The problem was in WM_NCCALCSIZE. During restore, the window can still have its minimized position, so MonitorFromWindow may select the primary monitor. I now use the rectangle passed to WM_NCCALCSIZE with MonitorFromRect.
I also moved pointer.Cancel to WM_KILLFOCUS and removed the minimized state tracking, Minimized reporting, and SWP_FRAMECHANGED.
I tested the official customdeco example on an extended dual-monitor setup by maximizing, minimizing, and restoring it from the taskbar. The buttons and hover state work correctly now.

app/os_windows.go
```diff
@@ -328,6 +328,7 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 		w.ProcessEvent(ConfigEvent{Config: w.config})
 	case windows.WM_KILLFOCUS:
 		w.config.Focused = false
+		w.ProcessEvent(pointer.Event{Kind: pointer.Cancel})
```
eliasnaur: All platforms want to cancel any ongoing touch gesture when the window loses focus, no? If so, please perform this call in the abstraction layer (say, in window.go), in a separate change.

app/internal/windows/windows.go
```diff
@@ -711,10 +713,19 @@ func GetWindowPlacement(hwnd syscall.Handle) WindowPlacement {
 }
 
 func GetMonitorInfo(hwnd syscall.Handle) MonitorInfo {
+	v, _, _ := _MonitorFromWindow.Call(uintptr(hwnd), MONITOR_DEFAULTTOPRIMARY)
+	return getMonitorInfo(v)
+}
```
eliasnaur: While here, please make the flags explicit and drop the helper:

```
func GetMonitorInfo(hwnd syscall.Handle, flags uint32) (MonitorInfo, bool) {
    v, _, _ := _MonitorFromRect.Call(uintptr(unsafe.Pointer(&r)), flags)
    if v == 0 {
        return MonitorInfo{}, false
    }
    var mi MonitorInfo
    _GetMonitorInfo.Call(monitor, uintptr(unsafe.Pointer(&mi)))
	return mi, true
}
```

Also, I think `MONITOR_DEFAULTTONULL` is the more appropriate for the callers.

app/os_windows.go
```diff
@@ -373,7 +374,8 @@ func windowProc(hwnd syscall.Handle, msg uint32, wParam, lParam uintptr) uintptr
 		// state. See https://devblogs.microsoft.com/oldnewthing/20150304-00/?p=44543.
 		// Note that trying to do the adjustment in WM_GETMINMAXINFO is ignored by Windows.
 		szp := (*windows.NCCalcSizeParams)(unsafe.Pointer(lParam))
-		mi := windows.GetMonitorInfo(w.hwnd)
+		// The current window rectangle may still be the minimized position while restoring.
```
eliasnaur: Drop comment. It's clear enough that `GetMonitorInfo(hwnd)` is inappropriate.

qianniancn:
Updated and pushed.
Pointer cancellation is now handled in app/window.go when focus is lost. This is a separate commit.
GetMonitorInfo now takes a rectangle and flags and returns ok. Both callers use MONITOR_DEFAULTTONULL. The helper and comment were removed.

eliasnaur:
Thanks! Both changes are now merged.

PR #181, opened 2026-08-17
op/clip: add reusable path cache

qianniancn:
Path data is normally stored in the frame Ops and becomes invalid after
Ops.Reset. This means unchanged paths have to be rebuilt every frame.

Add PathCache to keep path data in a separate Ops list so static paths can
be reused across frames.

### DIFF
--- a/op/clip/clip.go
+++ b/op/clip/clip.go
@@ -116,6 +116,20 @@ type PathSpec struct {
 	hash        uint64
 }
 
+// PathCache stores path data for reuse across frames.
+// Call Begin again when the path geometry changes.
+type PathCache struct {
+	ops op.Ops
+}
+
+// Begin starts a new path and invalidates any previously returned PathSpec.
+func (c *PathCache) Begin() Path {
+	c.ops.Reset()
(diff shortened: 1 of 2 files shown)


qianniancn:
Benchmark results for a path with 64 cubic segments:

    Cached      35.4 ns/op    0 B/op    0 allocs/op
    Rebuild      2.27 us/op   0 B/op    0 allocs/op

Reusing the path is about 64x faster in this benchmark.

op/clip/clip.go
```diff
@@ -116,6 +116,20 @@ type PathSpec struct {
 	hash        uint64
 }
 
+// PathCache stores path data for reuse across frames.
+// Call Begin again when the path geometry changes.
```
eliasnaur, line 120: The point of immediate mode GUI is minimal worry about invalidation. "Call Begin again when the path geometry changes" adds to the worry, so I don't think this feature belongs here.

qianniancn: Okay, I'll close this PR first.

PR #183, opened 2026-09-02
app: map Windows key states and numpad keys

qianniancn:
Windows modifier state and numpad keys are now mapped consistently.

The key events keep the physical key name:
- Shift+1 reports 1, not !
- Shift+; reports ;, not :

The shifted characters are produced by text input, not by the key name. This keeps shortcut handling independent from the active keyboard layout while text fields still receive the expected characters.


eliasnaur:
Thanks, merged.

PR #184, opened 2026-09-08
app: [Windows] support per-monitor DPI

qianniancn:
Windows system DPI awareness does not update a window's DPI when it moves to a monitor with a different scale. Prefer per-monitor awareness and apply the rectangle suggested by WM_DPICHANGED.

Fixes: https://todo.sr.ht/~eliasnaur/gio/707


app/internal/windows/windows.go
```diff
@@ -489,6 +493,7 @@ var (
 	_SetForegroundWindow         = user32.NewProc("SetForegroundWindow")
 	_SetFocus                    = user32.NewProc("SetFocus")
 	_SetProcessDPIAware          = user32.NewProc("SetProcessDPIAware")
+	_SetProcessDPIContext        = user32.NewProc("SetProcessDpiAwarenessContext")
```
eliasnaur: Match the names of the variable and the function.

app/internal/windows/windows.go
```diff
@@ -499,8 +504,9 @@ var (
 	_UnregisterClass             = user32.NewProc("UnregisterClassW")
 	_UpdateWindow                = user32.NewProc("UpdateWindow")
 
-	shcore            = syscall.NewLazySystemDLL("shcore")
-	_GetDpiForMonitor = shcore.NewProc("GetDpiForMonitor")
+	shcore                  = syscall.NewLazySystemDLL("shcore")
+	_GetDpiForMonitor       = shcore.NewProc("GetDpiForMonitor")
+	_SetProcessDPIAwareness = shcore.NewProc("SetProcessDpiAwareness")
```
eliasnaur: Match the name ("Dpi" vs "DPI").

app/internal/windows/windows.go
```diff
@@ -953,6 +959,16 @@ func SetFocus(hwnd syscall.Handle) {
 }
 
 func SetProcessDPIAware() {
+	if _SetProcessDPIContext.Find() == nil {
```
eliasnaur: Drop the check. `SetProcessDpiAwarenessContext` is available from Windows 8.1, and the minimum Windows version of recent Go releases is 10.

app/internal/windows/windows.go
```diff
@@ -953,6 +959,16 @@ func SetFocus(hwnd syscall.Handle) {
...
+		if r, _, _ := _SetProcessDPIContext.Call(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2); r != 0 {
+			return
+		}
+	}
+	if _SetProcessDPIAwareness.Find() == nil {
+		if r, _, _ := _SetProcessDPIAwareness.Call(PROCESS_PER_MONITOR_DPI_AWARE); int32(r) >= 0 {
+			return
+		}
+	}
 	_SetProcessDPIAware.Call()
```
eliasnaur: Drop the check and the fallback to `SetProcessDPIAware`. `SetProcessDpiAwareness` is available from Windows 8.1, and the minimum Windows version of recent Go releases is 10.

eliasnaur:
Thank you. I think if we're going to do this we might as well go all the way and only use `SetThreadDpiAwarenessContext` after the call the `LockOSThread`, removing all fallbacks. `SetThreadDpiAwarenessContext` is available from Windows 10 1607 which was released 10 years, and I think it's fine to require that.

The advantange of the per-thread version is that we won't interfere with or be affected by an executable manifest, or another part of the program setting and expecting a different global DPI setting.

The advantage of dropping old methods is that DPI behaviour will be the same across every Gio program, reducing the risk of hard to reproduce bugs. DPI switching between multiple monitors is already a corner case.

qianniancn:
Thanks,  I changed the code to useSetThreadDpiAwarenessContext after LockOSThread and removed the
process-wide DPI calls.

I also changed Configure to use GetWindowDPI so sizes are calculated from the monitor the window is currently on.

eliasnaur:
Thank you. Merged.

PR #185, opened 2026-09-14
preserve editor selection highlight for context menu

ddkwork:
This PR contains one changes:

1. **widget/editor: Preserve text selection highlight when context menu opens**
Remove the focus guard check in `paintSelection` to retain selection highlight after the editor loses focus. This matches Windows `EN_NOHIDESEL` behavior and fixes the visual issue where selected text disappears while the right-click context menu is shown, improving copy/cut UX.

Both changes are backward compatible.

Signed-off-by: ddkwork


io/input/pointer.go
```diff
@@ -582,6 +582,15 @@ func (q *pointerQueue) invTransform(areaIdx int, p f32.Point) f32.Point {
 	return q.areas[areaIdx].trans.Invert().Transform(p)
 }
 
+// getTransform returns the accumulated transform for the specified area.
+// The areas[areaIdx].trans is already the accumulated global transform.
+func (q *pointerQueue) getTransform(areaIdx int) f32.Affine2D {
```
whereswaldon, line 587: Now, this transform is 1 frame behind when the user actually sees it, right? We provide the transform that was present last frame when the event was generated. If the UI is highly dynamic or in the midst of a resize, the transform will be wrong for the current frame. That is one reason we've been reluctant to implement a feature like this.

Another reason is simply that providing absolute coordinates tends to make widgets that consume them not be portable and composable.

I guess I'd be okay with surfacing a feature like this, with appropriate warnings about the possibly stale transform. It facilitates some use-cases that are currently impossible or extremely janky in a fairly elegant way. @eliasnaur, what do you think?

eliasnaur:
A global transform, one frame lagged doesn't seem like a good idea to me. What's this change trying to solve?

ddkwork:
> A global transform, one frame lagged doesn't seem like a good idea to me. What's this change trying to solve?

I am developing two components: one is a dock layout control similar to Qt's dock widgets, and the other is a visual designer that generates layout code via drag-and-drop controls. Having access to absolute coordinates would be convenient for these projects. That said, while the coordinates obtained this way are correct in about 98% of cases, roughly 2% of scenarios can produce errors due to various environmental factors. It is important to note that this change is not intended for layout management. As mentioned in the comments above, we should rely on layout containers to dynamically adjust control positions.

eliasnaur:
I'd rather find a way to do this right. If you'd like to proceed, please describe your issue clearly, ideally with example programs. We can then discuss possible solutions.

ddkwork:
They are used in scenarios such as anchor positioning for pop-up right-click menus, overlay overflow prevention, and dock widgets.

eliasnaur:
Note that you'll have to carry the weight of effort here - I'm dumping my thoughts but don't have time to evaluate the pros and cons.

Concretely, let's consider right-click context menus. Why do you need the transform? Or more limited, why do you need the absolute window coordinates? Why not always anchor the popup to the element that triggered it?

ddkwork:
You're right. We're abandoning the absolute coordinate approach. Now let's look at the editor control. Do you think this fix makes sense?

ddkwork:
You're right. We're abandoning the absolute coordinate approach. Now let's look at the editor control. Do you think this fix makes sense?

eliasnaur:
What do you mean by "editor control"? Please be specific, list particular use-cases so it's easier for me to evaluate needs.

ddkwork:
When right-clicking selected text in the editor to open the context menu, the menu steals focus, causing the editor to clear the selection highlight before the copy command runs. We need to retain the selection highlight until the menu action finishes or a mouse click dismisses the menu; otherwise users see no highlighted text and get confused about what will be copied.

qianniancn:
问题解决了吗？

ddkwork:
> 问题解决了吗？

没有，我重新封装了编辑框状态期取出选中位置自绘选中色解决了这个问题。

你的仓库写得真好，方便留个微信或者qq深入交流吗？我还测试了gio的3d渲染等等方面感觉不错，准备抽空制作视频播放器

qianniancn:
@ddkwork  你加我qq吧  40210590

ddkwork:
> @ddkwork  你加我qq吧  40210590
> 

已发出请求

ddkwork:
> @ddkwork  你加我qq吧  40210590
> 

你有没有发现滚动条bug，不能拖哪到哪，是dpi缩放的锅，应该取屏幕坐标才能根治，官方滚动条控件也需要修复


另外一个是拖放外部文件支持，您英文好您来pr好一些

PR #186, opened 2026-09-15
app: [Windows] hide maximize button when MaxSize is set

qianniancn:
Maximizing a window with MaxSize moves the bounded window to the top-left corner instead of filling the work area. Hide the maximize button for such windows.

Keep frame resizing enabled unless MinSize and MaxSize are equal.


eliasnaur:
I don't understand why you're removing the maximize button just because a maximum window size is set. Don't you want to be able to maximize the the maximum window size? Or is that unusual behaviour in Windows?

qianniancn:
@eliasnaur In Windows, if `MaxSize` is set and the maximize button is not disabled, clicking the maximize button will move the window to the top-left corner. Since a maximum size has been set, I don't think the window should be maximized.

qianniancn:
Notice that when you click the maximize button, the window moves to the top-left corner, and its decorative elements—such as border shadows and borders—are removed.

https://github.com/user-attachments/assets/56fc2c7f-40e7-4e9d-81a1-17dadf847eca

eliasnaur:
I believe you. But that sounds like a bug, and hiding the maximize button is papering over it. It seems to me the correct behaviour is that clicking the maximize button should make the window go as large as possible.

According to https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-getminmaxinfo

> Sent to a window when the size or position of the window is about to change. An application can use this message to override the window's default maximized size and position, or its default minimum or maximum tracking size.

it sounds like our handling of WM_GETMINMAXINFO is not incorrect.

qianniancn:
@eliasnaur I tested this with MaxSize(800, 600).

WM_GETMINMAXINFO sets the expected size, but Windows still treats the window as maximized. Because the window does not fill the work area, Windows removes the normal border and shadow, so the result looks wrong.

I think disabling the maximize button when MaxSize is set is the right behavior. On Windows, maximizing a window normally means filling the available work area, which conflicts with a smaller maximum size.

eliasnaur:
I notice we only set `ptTrackMaxSize` in our [WM_GETMINMAXINFO handler](https://git.sr.ht/~eliasnaur/gio/tree/main/item/app/os_windows.go#L425). What about `ptMaxSize`?

qianniancn:
@eliasnaur Setting ptMaxSize works, but this is how Windows handles maximized windows: it removes the normal border and shadow even when the maximized size is smaller than the work area.

qianniancn:
I found the previous merge history and discussion: https://git.sr.ht/~eliasnaur/gio/commit/95354d80

eliasnaur:
Merged. Thanks for walking me through the consequences of this change.

PR #187, opened 2026-09-18
widget: add Editor and Selectable PaintSelectionWhenUnfocused

qianniancn:
The selection highlight is hidden when the widget loses focus, even though the selection itself is retained. While a context menu opened from a selection has taken focus, the selection is thus invisible while still being operated on.

Add an opt-in PaintSelectionWhenUnfocused field to both widgets. It can be updated between frames, e.g. set only while a menu is open.

Before Repair:
PixPin_2026-09-18_15-07-39

After Repair:
PixPin_2026-09-18_15-21-14


ddkwork:
nice work

whereswaldon (approved):
Thank you. This approach is noninvasive, and elegantly resolves the problem.

whereswaldon:
Merged!

```
To git.sr.ht:~eliasnaur/gio
   dc7e6295..08c90034  main -> main
```

